PR #29, opened 2015-11-11
pass through noProject flag from bin/ts-node

snapwich:
````ts-node --noProject test.ts```` doesn't seem to work without this boolean being passed through to the register function


blakeembrey:
Thanks :+1:

blakeembrey:
Released with [0.5.1](https://github.com/TypeStrong/ts-node/releases/tag/v0.5.1)

PR #36, opened 2015-11-20
allow tsconfig.json to have a different module than required by ts-node

jerryorta-dev:
I'm running into an issue where I need the tsconfig.json to have compilerOptions.module defined as "system", instead of "commonjs" due to karma testing. Yet I need "module" defined as "commonjs" to run gulp tasks. It may have been a feature to be able to override the "module" node via the tsconfig.json, but I'm not seeing the use case for node modules just yet. 

If tsconfig.json is supposed to overwrite the modules definition, then I could implement a feature such as 
haveing a compilerOptions.tsNodeModule node, that can then be defined as needed. And for node commands ts-node would use this instead.

See https://github.com/UIUXEngineering/angular2-jspm-seed

### DIFF
--- a/src/ts-node.ts
+++ b/src/ts-node.ts
@@ -79,15 +79,15 @@ function readConfig (options: Options, cwd: string, ts: TSish) {
     compilerOptions: {}
   }
 
-  config.compilerOptions = extend({
-    target: 'es5',
-    module: 'commonjs'
-  }, config.compilerOptions, {
+  config.compilerOptions = extend(config.compilerOptions, {
     rootDir: cwd,
     sourceMap: true,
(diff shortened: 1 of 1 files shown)


blakeembrey:
This can't be merged right now. You've disabled the ability to output non-ES5 (E.g. ES6) and although the module change is definitely unlikely to happen, it's still a breaking change to release it. I'll keep this in mind for a 1.0, but there's no good solution I can think of right now. The best would be to not load `tsconfig.json` unless specified, but I need to figure out how to have it usable with task runners like Gulp at the same time.

jerryorta-dev:
Thank you for the feedback.

PR #49, opened 2015-12-22
Improve test script for all TypeScript versions

blakeembrey:


blakeembrey:
Ugh, the thing that is killing this entirely is `tslint` with `EPEERINVALID` errors. I think I'll just have to force install everything.

blakeembrey:
Fixed by removing `tslint` after I run linting and before I install a different TypeScript version.

Issue #56, opened 2016-02-03
Allow `disableWarnings` to be set by environment variable


joeskeen:
I've been looking at #44 and I noticed that one of the things you want to do is make it fast and production-ready.  I'm really excited about this, but I realize that a lot of time is involved in a rewrite and that I can't realistically expect this to be done in the near future (although I'd be happy to help).

I'm working on a tool called `node-script ` (https://github.com/joeskeen/node-script), which allows you to create self-contained Node scripts (not requiring a package.json file or running `npm install` first).  One of the big things I'm in the process of tackling before v1.0 is transpiler support.  I'm currently using the `interpret` module to determine which packages are required to register the script file.  Currently `ts-node` is first in line for the `ts` file extension, so that is the one that `node-script` calls upon to run TypeScript `node-script`s.  The issue comes when I copy the script file to another location to run it in isolation.  Removed from its home during development, the script's `/// <reference />` comments and other typing resolution no longer resolves.  Since `interpret` points me directly to `require('ts-node/register')`, I can't call `require('ts-node').register({options})` without making a special case.

If `ts-node`'s `register` function was able to pick up whether to `disableWarnings` from `process.env` or something similar, I could make sure that that is set before `require('ts-node/register')` is called.

I think that this could be a good workaround until the new production-ready version of `ts-node` is ready.

blakeembrey:
Sure, seems reasonable. I'm certainly accepting PRs for work like this, I'm not sure when I'll get to it myself since I'm trying to keep up with Typings. Personally, I'd make it part of the default options: https://github.com/TypeStrong/ts-node/blob/master/src/ts-node.ts#L118.

blakeembrey:
Released with https://github.com/TypeStrong/ts-node/releases/tag/v0.6.0

PR #59, opened 2016-02-15
Allow the `disableWarnings` option to be set by an environment variable

joeskeen:
Implementation of #56.  

With these changes you can set the environment variable `TSNODE_DISABLEWARNINGS` to a truthy value in order to have ts-node ignore the warnings.

Also, while I was in there I exposed `compiler`, `noProject`, and `isEval` in a similar way.

### DIFF
--- a/src/ts-node.ts
+++ b/src/ts-node.ts
@@ -115,7 +115,17 @@ interface Project {
  */
 export function register (opts?: Options) {
   const cwd = process.cwd()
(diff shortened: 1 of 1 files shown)


blakeembrey:
Merged with https://github.com/TypeStrong/ts-node/commit/0afb09153c900080ce779e32f01e6f79863c46b8, sorry it took so long to get around to it.

PR #85, opened 2016-03-21
Update mocha example

### LINKED ISSUE #84: How to run mocha with ts-node?
Or a ts-mocha is needed? Or need to use tsify instead?

unional:
The `test.ts` and `src/**/*.spec.ts` are just file or file-globs example. The position doesn't matter.

So I suggest to keep the example a bit simplier.

By the way, how do I specify --project in this case?

Currently I have to: `cd source-test && mocha --require ts-node/register test.ts && cd ..`

Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ echo "console.log('Hello, world!')" | ts-node
 ### Mocha
 
 ```sh
-mocha test.ts --require ts-node/register src/**/*.spec.ts
+mocha --require ts-node/register <files or file-globs>...
 ```
 
 ### Tape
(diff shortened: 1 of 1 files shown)


blakeembrey:
The other examples should probably be consistent with this too then.

> By the way, how do I specify --project in this case?

There's environment variable, or you can run things from `ts-node` usually. E.g. `ts-node node_modules/mocha/bin/_mocha ...`.

unional:
is `<files or file-globs>...` clear enough?

blakeembrey:
Eh, I'm indifferent :smile: Just `[...args]` is good with me.

PR #87, opened 2016-03-23
Fixing absolute paths resolution

outring:
Hi!

I've found out, that it's not possible to run `ts-node` with absolute paths like that:
```
ts-node /home/user/test.ts
```
Here's the fix for that.

Cheers

### DIFF
--- a/src/bin/ts-node.ts
+++ b/src/bin/ts-node.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { join } from 'path'
+import { join, resolve } from 'path'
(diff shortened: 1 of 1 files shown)


outring:
Also I haven't found any tests, but found test scripts in `package.json` — is that ok?

blakeembrey:
Thanks for the catch, the tests are https://github.com/TypeStrong/ts-node/blob/master/src/ts-node.spec.ts.

outring:
@blakeembrey, ooops, wait for a PR with a tests then :)

PR #112, opened 2016-05-11
Allow environment variable to override default config.

cthrax:
While attempting to override the TS_NODE_PROJECT environment variable for use with gulp, which uses 'ts-node/register', I found that the environment variable did not actually override the default config. It does now. :)

### DIFF
--- a/src/ts-node.ts
+++ b/src/ts-node.ts
@@ -120,7 +120,7 @@ const DEFAULT_OPTIONS: Options = {
   getVersion,
   disableWarnings: process.env.TS_NODE_DISABLE_WARNINGS,
   compiler: process.env.TS_NODE_COMPILER,
-  project: process.env.TS_NODE_PROJECT,
+  project: process.env.TS_NODE_PROJECT || process.cwd(),
   noProject: process.env.TS_NODE_NO_PROJECT,
   ignoreWarnings: process.env.TS_NODE_IGNORE_WARNINGS
 }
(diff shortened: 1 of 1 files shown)


src/ts-node.ts
```diff
@@ -130,7 +130,7 @@ const DEFAULT_OPTIONS: Options = {
  */
 export function register (opts?: Options) {
   const cwd = process.cwd()
-  const options = extend(DEFAULT_OPTIONS, { project: cwd }, opts)
+  const options = extend(DEFAULT_OPTIONS, opts)
```
blakeembrey: Please just re-order the parameters here. We don't want cwd coming from two places.

blakeembrey:
Thanks for the catch, I only have the one comment.

src/ts-node.ts
cthrax: @blakeembrey I don't think re-ordering will work. The testing that I did showed "extend" would overlay an "undefined" if provided. So if DEFAULT_OPTIONS was overlayed on top of {project:cwd} and no environment variable was set, then the options would end up with undefined in project. I could move it down into this method and say "if (!options.project) { options.project = cwd; }" but that seemed less straightforward than putting it in the initialization of DEFAULT_OPTIONS. Thoughts?

blakeembrey: You're absolutely right! Sorry I missed that yesterday. In that case, I think it's good 👍

PR #122, opened 2016-05-28
Make it possible to supply a configFileName

voortwis:
to make it possible to override the default tsconfig.json file that is looked up
Created this change for angular2-webpack-starter the e2e fails and needs a different tsconfig.json than the normal build-run cycle.

new configuration option: configFileName

### DIFF
--- a/src/ts-node.ts
+++ b/src/ts-node.ts
@@ -59,6 +59,7 @@ export const EXTENSIONS = ['.ts', '.tsx']
 export interface Options {
   compiler?: string
   noProject?: boolean
+  configFileName?: string
   project?: string
   ignoreWarnings?: Array<number | string>
   disableWarnings?: boolean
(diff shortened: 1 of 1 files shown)


blakeembrey:
Is it possible to mirror `tsc` compiler behaviour here? I don't know if they've exposed the function yet, I was hoping it'd be an extension of the same function, but they accept paths as `--project` now IIRC.

voortwis:
Have been thinking about that too.. When looking at the tsconfig I was trying to create this file in a child-folder and make the paths to the sources relative. Sadly this didn't make the angular-webpack-starter (npm run e2e) work. Maybe any other suggestions i should try.

blakeembrey:
I've switched back to `tsconfig` for resolve, as `typescript@latest` still has too many `tsconfig.json` resolving issues present. In `tsconfig`, I've enabled support for a filename.

Issue #126, opened 2016-06-10
Option to allow overriding tsconfig.json compiler options


niieani:
It would be really useful to be able to override parameters passed to the compiler in the hook.
For example, I like to use "allowJs" set to true, so that VSCode gives typing Intellisence, but I don't actually compile JS files, because I'm using a Webpack loader with a separate `tsconfig.json`.

niieani:
Implemented here: https://github.com/TypeStrong/ts-node/pull/127

PR #127, opened 2016-06-10
allow overriding tsconfig.json compiler options

niieani:
https://github.com/TypeStrong/ts-node/issues/126

### DIFF
--- a/src/ts-node.ts
+++ b/src/ts-node.ts
@@ -65,6 +65,7 @@ export interface Options {
   disableWarnings?: boolean
   getFile?: (fileName: string) => string
   getVersion?: (fileName: string) => string
+  compilerOptions?: Object
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/ts-node.ts
```diff
@@ -97,7 +98,8 @@ function readConfig (options: Options, cwd: string, ts: TSCommon) {
       inlineSources: true,
       declaration: false,
       noEmit: false
-    }
+    },
+    options.compilerOptions
```
blakeembrey: You can't be allowed to override these options, you should put it between our options and the config options.

blakeembrey:
Do you think we can expose CLI flags for this too? Either as JSON, I guess, or using https://www.npmjs.com/package/subarg.

niieani:
@blakeembrey JSON CLI addition would be good indeed. I take your point regarding not allowing overriding of:
```
    {
      sourceMap: false,
      inlineSourceMap: true,
      inlineSources: true,
      declaration: false,
      noEmit: false
    },
```

I'll have to come back to this at a later time.

blakeembrey:
Closed with https://github.com/TypeStrong/ts-node/pull/128.

niieani:
@blakeembrey awesome, thanks!

Issue #44, opened 2015-12-03
TypeScript Node 1.0


blakeembrey:
From the issues I've seen logged thus far, I'm looking at rewriting `ts-node` into something that can suit more people. Here's some of the points that need covering for 1.0. Comments are appreciated.

* The goal is to be fast and production-ready out of the box (breaking change)
  * This means no type information, because that requires loading the entire project into memory
  * All files will transpile directly, in-memory using TypeScript `transpileModule`
  * Can enable a cache directory for a simple speed up
* One level up, semi-production ready, is the `--typed` flag
  * This transpiles the entire project and caches it **once** on startup
  * This has type information, but only for things in the initial project/`require` - subsequent requires of TypeScript files would fall back to the first point
* The next level, non production ready, is the `--fully-typed` flag
  * This is basically the current version of `ts-node`, suitable for a test suite
  * Everything needs to be in-memory because `require` can be "out-of-band". E.g. `.ts` -> `.js` -> `.ts`, not possible to handle by just the TypeScript compiler
  * We can think about way to fix this "out of band" require issue, but I believe changed files should always be re-loaded correctly (E.g. watching a test suite should get new diagnostics without re-compiling everything) /cc @basarat Any thoughts on this point? Maybe a way to serialize the compiler state in some form?
* The project flag is disabled by default (still considering this)
  * Enable ability to specify directory or file
* Compiler option flag - specify TypeScript compiler options via JSON
* All flags should have an environment variable for environments that can't specific CLI flags - E.g. Gulp, Grunt, etc.

basarat:
> The goal is to be fast and production-ready out of the box (breaking change)

:100:  for this.  People can have their (IDE | `tsc -w`) provide errors :rose:

blakeembrey:
@basarat I wish I had my IDE (cough, cough, `atom-typescript`, cough) displaying errors, but they aren't part of my `tsconfig.json` file :wink: They don't need compiling, they don't need publishing and they break when moved away from the fixtures, etc. 

But in general, yes. I've had a few people ask about production, when production is really just a 5 liner and simpler than the current `ts-node` implementation which supports everything else. Type checking is fantastic for the CLI and demos though, it's pretty nifty (will be better once node exposes multiline REPL support).

tamascsaba:
:+1:

blakeembrey:
[...]

PR #130, opened 2016-06-11
Enable `--fast` CLI flag for transpile mode only

blakeembrey:
Related to https://github.com/TypeStrong/ts-node/issues/44.

Without `--fast` tests on `typings-core`:

```
npm run test-spec  7.98s user 0.52s system 72% cpu 11.691 total
```

With `--fast` tests on `typings-core`:

```
npm run test-spec  3.52s user 0.47s system 56% cpu 7.098 total
```


mjohnsonengr:
1.5x is an understatement.  Upgrading from 0.8 -> 0.9.1 got my test suite from 3.5 minutes to 22 seconds; the --fast flag then took me down to 2 seconds!

Awesome work and cheers @blakeembrey!! :grin:

blakeembrey:
Wow, that is ridiculous! Sorry it took so long for that improvement to make it's way in 😄

mjohnsonengr:
Hah! No worries! I just now started working on getting our test suites to run in node last week and only hit the performance issues this morning before I found this upgrade! (previously it was a mess of hacks making them run with compiled JS coverage in the browser)

PR #139, opened 2016-06-28
:memo: mocha watch

basarat:
Merge if happy :rose:

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ echo "console.log('Hello, world!')" | ts-node
 ```sh
 mocha --require ts-node/register [...args]
 ```
+With watch mode: 
+
+```sh
+mocha --require ts-node/register [...args] -- --watch --watch-extensions tsx,ts
+```
 
 ### Tape
 


blakeembrey:
I'd be happy to merge the section together, if that makes sense. Also, does Mocha have a transpiler option that works for both? IIRC, it's something like `--extension ts:ts-node`. Been a long while since I used Mocha myself.

basarat:
> IIRC, it's something like --extension ts:ts-node. Been a long while since I used Mocha myself.

I searched long and hard to find `watchExtensions`. Seems to be the only way in town.

blakeembrey:
@basarat Oh, it's `--compilers` - did you try it? https://github.com/mochajs/mocha/blob/master/bin/_mocha#L81

blakeembrey:
It seems like it does both automatically: https://github.com/mochajs/mocha/blob/master/bin/_mocha#L297-L306. I think the reason I avoided it in the beginning was the duplication of having to do `--compilers ts:ts-node --compilers tsx:ts-node`.

basarat:
I just did a blind run (i.e. didn't investigate if what was proposed made sense). Didn't work :rose: 

screen shot 2016-06-28 at 3 18 40 pm

blakeembrey:
Thanks! I think I missed something though - I'm guessing it should be `ts-node/register` in both those places.

basarat:
> I think I missed something though - I'm guessing it should be ts-node/register in both those places.

Indeed it worked! :rose: `mocha --compilers ts:ts-node/register --compilers tsx:ts-node/register` 

We can change the default `--require` to `--compilers` and then `--watch` is intutive. How does that sound?

blakeembrey:
Great! 👍

basarat:
closing in favor of : https://github.com/TypeStrong/ts-node/pull/140

PR #141, opened 2016-06-30
Allow require of d.ts files.

hansl:
This will only trigger if someone runs require(somefile.d.ts).

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -169,6 +169,19 @@ describe('ts-node', function () {
       expect(m.example('foo')).to.equal('FOO')
     })
 
+    it('should be able to require typescript definitions', function () {
+      expect(function () { require('../tests/module3') }).to.throw()
+
+      require('../tests/module2.d.ts')
+      const m = require('../tests/module3')
+
+      expect(m.example()).to.equal(3)
+    })
+
+    it('should not be able to require typescript definitions without extension', function () {
+      expect(function () { require('../tests/module2') }).to.throw()
+    })
+
     it('should compile through js and ts', function () {
       const m = require('../tests/complex')
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -280,7 +280,14 @@ export function register (opts?: Options) {
         throw new TSError(diagnosticList)
       }
 
-      return output.outputFiles[0].text
+      if (output.outputFiles && output.outputFiles.length > 0) {
+        return output.outputFiles[0].text
+      } else {
+        // There was no errors and no output file; this might be a definition.
+        // At this point we have the definitions in our program, so just return
+        // an empty file for Node to execute.
+        return ''
+      }
     }
 
     compile = function (fileName: string) {
--- a/tests/module2.d.ts
+++ b/tests/module2.d.ts
@@ -1,0 +1,3 @@
+
+// This type is ambient.
+declare type MyType = number
--- a/tests/module3.ts
+++ b/tests/module3.ts
@@ -1,0 +1,5 @@
+
+// Use an ambient type. This should fail if loaded directly.
+export function example(): MyType {
+  return 3
+}


hansl:
@blakeembrey I found that using `d.ts` files in some cases is really useful. For example, my normal code does not import or `/// <reference ... />` Jasmine typings, but I want to add it to my spec. With this I can `require(...)` the `d.ts` files directly.

Since I did not add the `d.ts` as an extension to Node, I have to use the full file name (including extension) directly. In my example above that takes the form of `require('@types/jasmine')`.

blakeembrey:
I don't think I'll merge this. You can look at a previous PR for this as to why. Basically, this code is not valid and I won't be implementing hacks in this module to make it work. You should not be requiring a definition file like that, you're probably using typescript incorrectly.

blakeembrey:
I'll probably use this as a basis to throw an error so people can learn how to use it properly though - can you give more information on what you're trying to do?

hansl:
My use case is relatively simple but common: I have a tsconfig for my project, and I want to run tests. The tsconfig does not include typings for Jasmine as I want to make sure these aren't used. But my spec files need them.

Using the TS program directly (or the command line), I would simply add the `d.ts` for jasmine (+ addition dev-only typings) and register the `ts` extension as you do here.

Note that using the command line I can actually do that by passing additional files. This case is that I have a special runner that uses the `register()` call to skip the compilation (I don't want to compile spec files, I have a separate publish process). Also note that Jasmine `d.ts` isn't a module, so I cannot `import` it directly from TypeScript.

I have three alternatives for this, if I want to keep proper typings separated in my project; either use `/// <reference ...` at the top of every spec files (not ideal, unnecessary boilerplate), or have a special `tsconfig.json` for spec files (not ideal, I want to reuse the same one). The last one is what I'm doing currently; having a facade that `/// <reference ...` and export wrapper functions that call the ambient definitions. This makes awkward and would be duplicated for a lot of modules.

This PR was trying to allow adding additional definition files to the `ts-node` program. I understand using `require()` seems awkward, would you rather see access to the program instance, or maybe an exported function to add files separately (e.g. `require('ts-node').addDefinition(...)`) or something alike?

hansl:
@blakeembrey: See above.

blakeembrey:
An API method is definitely better. However, it sounds like you're using TypeScript references. It sounds like you're using TypeScript 2.0 - for which, I believe there's a `tsconfig.json` option that enables "libraries" you can load. You'll need to find it though, I haven't used it personally. Even in 1.8 though, you can use `tsconfig.json` to reference definitions.

hansl:
So the solution you’re proposing is number #2; use different tsconfig for
specs. Which means I have to maintain 2 tsconfig for what is basically a
one-line change.

I cannot have Jasmine definitions in my browser / node. I just want it for
my spec files.

As for TypeScript 2.0, I’m compiling using 1.8 but using the @types
node_modules packages because it’s simpler. That is beside the problem
though, I would have the same issue if I had handmade d.ts files (or using
typings).

-- 
Hans Larsen

On June 30, 2016 at 1:57:21 PM, Blake Embrey (notifications@github.com)
wrote:

An API method is definitely better. However, it sounds like you're using
TypeScript references. It sounds like you're using TypeScript 2.0 - for
which, I believe there's a tsconfig.json option that enables "libraries"
you can load. You'll need to find it though, I haven't used it personally.
Even in 1.8 though, you can use tsconfig.json to reference definitions.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/TypeStrong/ts-node/pull/141#issuecomment-229785632>,
or mute
the thread
<https://github.com/notifications/unsubscribe/AApn8aaVdulllrDTg1cHweqaXr0Tot3tks5qRCzRgaJpZM4JCcgv>
.

blakeembrey:
No, it's not beside the point. If you were using Typings or copying them, they would be used automatically and be in the context.

blakeembrey:
Right now I have to make a lot of assumptions on what you're doing. You can share your `tsconfig.json` if you want support, but I will not be merging this hack.

PR #145, opened 2016-07-18
Change outDir so that its on the same level as the typescript project

rundef:
Please check out the following issue https://github.com/istanbuljs/nyc/issues/314 to understand why I opened this pull request

Thank you !

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1,6 +1,6 @@
 import { expect } from 'chai'
 import { exec } from 'child_process'
-import { join, normalize } from 'path'
+import { join } from 'path'
 import semver = require('semver')
 import ts = require('typescript')
 import proxyquire = require('proxyquire')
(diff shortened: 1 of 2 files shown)


blakeembrey:
There doesn't seem to be any guarantee this won't be broken again in the future. I think it'd be better to move back to manual source map parsing and rewriting to make these absolute paths as neither of those `../..` mean anything and I have no idea why they are there. E.g. https://github.com/TypeStrong/ts-node/commit/2d20d65819ad8a474ce37d5310da233d4b51426b.

rundef:
Thanks for the quick reply

I've changed the sourcemap mapping back to absolute URLs.

Let me give you a better picture of what's going on with the nyc+ts-node problem, there may be a better way to fix that problem ?

1. nyc requires **ts-node/register** 
2. As an example, let's say my project only contain one file: **app/helpers/Logger.ts** - it gets compiled to **tmpXYZ/app/helpers/Logger.js** (because _ts-node_ sets the _outDir_ to **tmpXYZ**)
3. The base64 sourcemap comments at the end of the file contains a sources array with the value of **../../../app/helpers/Logger.ts** which is absolutely the correct relative path
4. nyc doesn't know that ts-node compiled the file into a sub-directory, it simply does a *join(ts_file, sourcemap.sources)*, which ends up being the incorrect path

blakeembrey:
Thanks! It seems that method no longer exists in `typescript@next` so the tests are failing. It might be easier to keep the `inlineSources: true` flag and just update the filenames instead. That way we can avoid that API call altogether. What do you think?

blakeembrey:
Merged it https://github.com/TypeStrong/ts-node/commit/85745f574f80e3a2a53337bf60c33d00f25cf469.

rundef:
:+1: thanks ! hopefully this gets pushed on npm soon

PR #160, opened 2016-07-28
Add `typings` property to `package.json`

janslow:
Allows import of the definitions in packages which want to programatically use `ts-node`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "preferGlobal": true,
   "description": "TypeScript execution environment and REPL for node",
   "main": "dist/index.js",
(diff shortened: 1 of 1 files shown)


blakeembrey:
No thanks. This module has dependencies Typings (https://github.com/TypeStrong/ts-node/blob/master/typings.json). If any of those types leak through, it will break. See https://github.com/typings/typings/blob/master/docs/faq.md#should-i-use-the-typings-field-in-packagejson. You can always install it with Typings.

janslow:
Ok, that makes sense.

PR #162, opened 2016-07-29
Fix compilerOptions environment settings

cdschmitz:
The documentation mentions `process.env.TS_NODE_COMPILER_OPTIONS` as a way to set compiler options through the environment, however this variable was never set in `DEFAULT_OPTIONS`.


blakeembrey:
Thanks!

PR #214, opened 2016-10-08
fix jsx sourcemaps with "jsx": "preserve"

Mosho1:
When `"jsx": "preserve"` TS compiler option is set, the output filename is `filename.jsx`, but currently the implementation assumes `.js`. This causes the sourcemap url comment to point to an invalid path, as well as caching the output as `.js` rather than `.jsx`.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -4,7 +4,7 @@ import { join } from 'path'
 import semver = require('semver')
 import ts = require('typescript')
 import proxyquire = require('proxyquire')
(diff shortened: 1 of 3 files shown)


Mosho1:
note: coverage decreased at first because I skipped a few tests that were failing (on windows 10) and accidentally pushed.

Mosho1:
note: decreases coverage at first due to skipping a few tests which were failing (on windows 10) and accidentally pushing.

PR #217, opened 2016-10-11
use final ts configuration

Mosho1:
Right now it doesn't use the configuration merged with `tsconfig.json`


blakeembrey:
Thanks!

PR #226, opened 2016-10-26
Fix https://github.com/Microsoft/TypeScript/issues/11870: Do not failon empty file list diagnostic messages

mhegazy:
Originally reported in https://github.com/Microsoft/TypeScript/issues/11870.

Change https://github.com/Microsoft/TypeScript/pull/11743 added a new error for TS 2.1 if the `files` list is empty, or `include`+`exclude` result in an empty set. 
This breaks usage of ts-node. 

@blakeembrey I could not find a better solution here, so appreciate your guidance.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -139,7 +139,10 @@ export interface Register {
  */
 export function register (options: Options = {}): () => Register {
   const compiler = options.compiler || 'typescript'
-  const ignoreWarnings = arrify(options.ignoreWarnings || DEFAULTS.ignoreWarnings || []).map(Number)
+  const emptyFileListWarnings = [18002, 18003]
+  const ignoreWarnings = arrify(
+    options.ignoreWarnings || DEFAULTS.ignoreWarnings || []
+  ).concat(emptyFileListWarnings).map(Number)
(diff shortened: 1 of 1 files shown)


blakeembrey:
Thanks. I think this looks good, I'll merge this and make a new release ASAP, sounds like TypeScript relies on it. I didn't expect that feature to be used like that 😄 

FWIW, would it make more sense for these sorts of checks to be done at a higher level like the CLI/server instead? I think this will be an edge-case that'll come up for every compiler tool such as `ts-loader`, `tsify`, etc. I've run into a few of these API idiosyncrasies now and it makes a tricky exercise for someone new to use the APIs effectively.

Also, the original issue (https://github.com/Microsoft/TypeScript/issues/11621) isn't really solved by this as one might have a single file in `src/` and still run into the issue. Perhaps tooling/VS Code could make it more obvious which configuration file is in use (I imagine this feature would be similar to selecting which `tsconfig.json` to use as a manual override - where it defaults to saying "none selected" or "not in project" (which is what Atom does) - since `tsc` can accept custom paths to a config file)?

PR #228, opened 2016-10-29
Support --nolazy

Llorx:
--nolazy is usually used along with --debug-brk


blakeembrey:
Thanks! I will probably move to the `v8flags` module soon.

PR #229, opened 2016-10-29
´opts´ array mutation inside loop fix

Llorx:
´ops´ length was decreased while ´i++´ was increased, so not all array elements were parsed.


Llorx:
Compile errors

PR #230, opened 2016-10-29
´opts´ array mutation inside loop fix

Llorx:
`ops` length was decreased while `i++` was increased, so not all array elements were parsed.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -6,18 +6,22 @@ import { join } from 'path'
 const args = [join(__dirname, '_bin.js')]
 const opts = process.argv.slice(2)
 
-let i = 0
-for (i = 0; i < opts.length; i++) {
+optsloop:
+for (let i = 0; i < opts.length; i++) {
   const arg = opts[i]
   const flag = arg.split('=', 1)[0]
 
   switch (flag) {
     case '-d':
       args.unshift('--debug')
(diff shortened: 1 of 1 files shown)


Llorx:
What is happening?
```
            throw new TSError(formatDiagnostics(configDiagnostics, cwd, ts, 0)
                  ^
TSError: ⨯ Unable to compile TypeScript
Unknown compiler option 'strictNullChecks'. (5023)
```

blakeembrey:
Good catch! It seems this change has broken the opts parsing and the `--project` flag is being ignored now. At least, I can see a pretty big issue here in non-node options littered between node options aren't parsed properly (you slice them all now). E.g. `ts-node --project src --debug script` would fail now. I think the correct solution is just to keep two arrays or to decrement `i` each splice.

Llorx:
I wanted to avoid the ´i--´ thingy, but, well, finally added it.

blakeembrey:
Thanks. I think I'm on the super hated side 😄  At this point, maybe it should be a function and `return` the result instead. I'll merge and refactor soon, thank you.

Llorx:
Yeah, I also don't like it. Is the first time I use a label on javaScript. Remembered me of Quick Basic... good times.

Holy shit, I feel so old right now xDD

PR #235, opened 2016-11-08
Include explicit direction for mocha ts-node configuration.

jamelt:


blakeembrey:
This isn't really a Mocha thing though, this is just anywhere you can't pass options (E.g. anything using `ts-node/register`). That means `node -r ts-node/register`, `gulp`, etc. If you can make the PR more generic, I'd be happy to merge.

Issue #146, opened 2016-07-18
Don't use `cwd` at `register` time


weswigham:
`gulp` respects a `cwd` command - however this argument only sets the process's `cwd` _after_ the register hook has been run (but before the `gulpfile` is required). This causes dissonance where `ts-node` can only correctly transpile files when the cwd does not change between the register invocation and a require call. See [this issue](https://github.com/gulpjs/gulp-cli/issues/89) for discussion.

blakeembrey:
What do you suggest as an alternative? It's used to load `tsconfig.json`.

weswigham:
Can we delay loading of the `tsconfig.json` until at least the first call to `require`?

blakeembrey:
That might be possible - so it'd only initialize TypeScript after we require our first TypeScript file?

weswigham:
Yep. That seems like it should work.

blakeembrey:
@weswigham Actually, I just realised this doesn't exactly work unless we add a new flag. The current behaviour of `ts-node` also supports `allowJs`, but that can't be known unless we load `tsconfig.json`. I suppose it could be changed to a `ts-node` flag instead, but what are your thoughts?

blakeembrey:
It could also be a major change to _always_ requiring the `--project` option to be set and make it be "no project" by default. This would avoid the `cwd` from being used at all until TypeScript is created. Should the TypeScript option (assuming this is for `getCurrentDirectory`) always use `process.cwd()` too?

weswigham:
So the exact issue we were having is that, without our Gulpfile.ts, if used from a subdirectory like `cd src && gulp local --cwd ../`, `ts-node` compiles the `Gulpfile` with the `cwd` still being within the `src` directory (resulting in many type-not-found errors, ofc). We're not even in a scenario where passing `--project` is an option (so the only method to configure the hook is via the config file) - the require hook is non-configurably loaded by `gulp` via `liftoff` via `rechoir`. On the other hand, `rechoir` via `interpret` only loads `ts-node` for either `.ts` or `.tsx` files, so the js-mode will never come up via gulp. _On the other hand_ we definitely still want the ability to include js files in our Gulpfile.ts compilations (usually when a lib has no types and it's effort to write some), so keeping the ability to have `allowJs` on would be good. We just don't _need_ a `js` entrypoint.

blakeembrey:
[...]

PR #238, opened 2016-11-18
Fix bug where js extension gets not registered when lazy is set to true and allowJS is used via the configuration

mlegenhausen:
This bugfix solves a problem when using `ts-node/register` with a not ts file as entry point.
The lazy option prevents the evaluation of the tsconfig.json file until the first ts file is loaded.
In scenarios where only js files are used and ts-node is used as "transpiler" this will never happen.
The result is that ts-node is never used for javascript files even when `allowJS` is defined in the `tsconfig.json`.

This pull request solves the issue by reading the configuration as soon as possible to determine if `allowJS` is set to true or not. If so the js extension get registered.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -152,6 +152,10 @@ export function register (options: Options = {}): () => Register {
   const cacheDirectory = options.cacheDirectory || DEFAULTS.cacheDirectory || join(tmpdir(), 'ts-node')
   const compilerOptions = extend(DEFAULTS.compilerOptions, options.compilerOptions)
   const originalJsHandler = require.extensions['.js']
+  const cwd = process.cwd()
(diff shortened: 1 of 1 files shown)


blakeembrey:
Pretty sure I can't merge this as it's doing the opposite of what it's designed to do now. It was purposely written to lazily load the configuration file. See https://github.com/TypeStrong/ts-node/issues/146. I'll need to think on how to solve this for everyone.

blakeembrey:
Decided to move ahead for this (see https://github.com/TypeStrong/ts-node/commit/9caa1c57d80de2f6b70def80de6e7f1c99f89154). It's a bit tidier and I removed the, now useless, option for `lazy`.

PR #253, opened 2016-12-27
Pass through --preserve-symlinks to Node.js

zkochan:
We'd really need this at [pnpm](https://github.com/rstacruz/pnpm)


kirillgroshkov:
Where did this flag (`--preserve-symlinks`) now? I don't see it supported in current codebase...

Issue #219, opened 2016-10-12
Not using paths from tsconfig


Jontem:
Typescript 2.0 has added support for having modules in other places than node_modules with the paths property. More info can be found in the [documentation](https://www.typescriptlang.org/docs/handbook/module-resolution.html). 

When running with ts-node this does not work. It just gives me the error that the module cannot be found.

test1.ts
```js
import * as foo from "foo";

console.log(foo.hello);
```

lib/foo/index.ts:
```js
export const hello = "hello";
```

Then running 
```
ts-node --compilerOptions '{"baseUrl": "./", "paths": {"foo": ["lib/foo"]}}' test1.ts
```

Same error with this tsconfig.json
```json
{
    "compilerOptions": {
        "baseUrl": "./",
        "paths": {
            "foo": ["lib/foo"]
        }
    }
}
```

Running `tsc` works

blakeembrey:
This would be a runtime error. TypeScript does not rewrite the paths for you. Node does not know where to find your paths based on your configuration. Please search existing issues.

Jontem:
#138

MrCrimp:
For any search-engine travellers reading this,
This can to some extent be worked around by setting the NODE_PATH to your package.json dir, e.g

----
tsconfig.json
```

baseUrl: ".",

"paths": {
      "*": [
        "*"
      ]
    }
```

----
package.json
>"test": "export NODE_PATH=./ && mocha"

Then imports like these works for `ts-node/register` runs as well as tools providing resolve paths (babel, webpack)

`import * as log from 'src/utils/log'`

mnasyrov:
Guys, this is a very useful workaround with NODE_PATH should be somewhere in README file. I stuck in two hours with baseUrl/paths problem while not found it in closed issues :)

jonaskello:
For those interested I did a PR to enable execution of projects with tsconfig paths in #254.

jonaskello:
@MrCrimp @mnasyrov I also used the NODE_PATH work-around, but this can now be solved easier with the [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) package. Just add it as a dev dependency and use the `-r tsconfig-paths/register` option of ts-node or mocha and all paths will work automatically without NODE_PATH :-).

bennyn:
[...]

Issue #138, opened 2016-06-19
tsconfig.json/paths not working with ts-node


unional:
One neat trick to use `tsconfig.json/paths` is to pointing the module name to your source file:

```ts
// tsconfig.json of blue-tape-fixture
{
  ...
  "paths": {
    "blue-tape-fixture": [ "src/" ]
  }
}
```
This way, your test files can reference the source as if it is a proper module:
![image](https://cloud.githubusercontent.com/assets/3254987/16176202/fcb426da-35bc-11e6-98c6-c7c61b8a0bd7.png)
You can see it is working in vscode. But it does not when I run test with ts-node:
```js
// package.json
{
  "scripts": {
    "test": "ts-node -P tsconfig.build.json node_modules/blue-tape/bin/blue-tape \"test/**/*.ts\" | tap-spec"
  }
}
```

The error is:
```sh
Error: Cannot find module 'blue-tape-fixture'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
```

You can see it here: https://github.com/unional/blue-tape-fixture/tree/ts-node
```sh
# clone
npm install
npm test
```

blakeembrey:
@unional That's the node.js error - TypeScript is working fine, but node will never be able to resolve an alias like that unless TypeScript emits code that rewrites paths. What do you think is the solution here?

unional:
I see. In order for it to work, a module-path-rewriter need to be written and loaded by `ts-node` for it to work.
Alternatively is to load the transpiled code `dist/*`, which kind of defeat the purpose of `ts-node` (I said kind of because the test file, in this use case, can remain as ts files).

blakeembrey:
This module uses the emitted code though - it's TypeScript itself that doesn't do any rewriting right?

unional:
Just check. Nope it does not do any rewriting. Seems like the `baseUrl` and `paths` is only used during compilation. By design or bug on TS side?

unional:
Update https://github.com/unional/blue-tape-fixture/tree/ts-node to demo the issue.
run `npm run build-test` and see under `dist/test/index.ts` it is will requiring `blue-tape-fixture`

unional:
Confirm above statement: https://github.com/Microsoft/TypeScript/issues/5039#issuecomment-206451221

blakeembrey:
[...]

PR #254, opened 2016-12-28
Add paths support

jonaskello:
Add support for executing typescript projects that uses `paths` in tsconfig. Discussed in #138 and #219.

It is achieved by patching node's `Module._load` function which is undocumented and but I found this technique used elsewhere such as in [mock-require](https://github.com/boblauer/mock-require). The loader patch is only applied if there are paths specified in tsconfig so it should not alter the behaviour if there are no paths.

There is also a small example [here](https://github.com/jonaskello/ts-node-paths-example) to test it. I would like to add some tests for this PR but not sure how to do it since my functions are not exported.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 import { relative, basename, extname, resolve, dirname, join } from 'path'
-import { readdirSync, writeFileSync, readFileSync, statSync } from 'fs'
+import { readdirSync, writeFileSync, readFileSync, statSync, existsSync } from 'fs'
 import { EOL, tmpdir } from 'os'
 import sourceMapSupport = require('source-map-support')
 import extend = require('xtend')
@@ -185,6 +185,11 @@ export function register (options: Options = {}): () => Register {
     const config = readConfig(compilerOptions, project, cwd, ts)
     const configDiagnostics = filterDiagnostics(config.errors, ignoreWarnings, disableWarnings)
     const extensions = ['.ts', '.tsx']
+
+    // Augment loading of node modules according to paths in tsconfig
+    if (config.options.baseUrl && config.options.paths) {
+      installModuleLoadForPaths(config.options.baseUrl, config.options.paths)
+    }
 
     const cachedir = join(
       resolve(cwd, cacheDirectory),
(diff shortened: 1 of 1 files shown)


blakeembrey:
I mentioned it in https://github.com/TypeStrong/ts-node/issues/138#issuecomment-250992304, but I don't have any desire to enable this in `ts-node` directly as it'd be promoting poor node.js practices and portability. I think it'd be better done as a separate module people can just require (e.g. `ts-node -r tsconfig-paths-runtime index.ts`). That way you can also work on enhancements and changes separately to the codebase here, as I probably won't be able to maintain that section of code myself.

sanex3339:
+1 for plugin

jonaskello:
Yes, I did read that comment but I wasn't aware of the `-r` option, so I thought "plugin" was refering to that it should be opt-in or a typescript plug-in. Sorry for the misunderstanding, guess I should have RTFM before trying anything :-). I considered hiding it behind a CLI flag but since it is only activated if you actually have paths in your tsconfig it is already opt-in. And if you have paths in your tsconfig it will not work today anyway so I thought using an undocumented feature in node to make it work was better than not working at all. But I do understand the reluctance to maintain undocumented features.

Now that I know the `-r` option I will certainly explore doing it as a plug-in. However I would say that at some point it would be nice if paths would "just work" with ts-node. The main reasons for using paths AFAIK:

* It is the only supported way to maintain module style custom typings in your project, as discussed [here](https://github.com/Microsoft/TypeScript/issues/11329) and [here](https://github.com/Microsoft/TypeScript/issues/11137).
* It is the only way to inter-op with jspm (as discussed in the original paths issue in the TS repo).
* It is the only way to structure a TS project without having a lot of "../.." references.

For these reasons I think there is a growing number of projects that use paths. I consider ts-node a great tool for the typescript community and I think it would be unfortunate if the part of the community that uses paths in their tsconfig could not use it, or had to make special steps to make it work. 

Just to clarify, the only reason it cannot be included directly in ts-node is that it is using the undocumented patch of Mode._load or are there other reasons?

One question about doing it as a `-r` option plug-in, how do I know which tsconfig file ts-node has selected to use?

jonaskello:
After further research I found that the `-r` option is actually something supported by node rather than ts-node and it is also available when executing tests in mocha. Knowing that it makes more sense to make it a separate module since it could have merit outside of ts-node.

I still think that eventually having the "it just works" experience would be the nicest way in ts-node, but until we find an agreeable way to do that I'll close this and report back on my progress as a separate module loadable with `-r`.

blakeembrey:
Mostly don't want to include it in core as it's not something I can maintain myself. My time is pretty limited for OSS lately and new features like this can sometimes come with a lot of bugs. Bugs mean issues I need to fix. A better way to support a feature like this is to build the module independently (I always intended to write an abstract version, but as you know - not a priority for me) and allow it's usage programmatically. Once it's stable, I can just require it and include it in `ts-node` easily, but until then people can use something like a `tsconfig-paths/register` hook or a programmatic API to register the paths to use.

jonaskello:
I hear you about the lack of time, I did not get my standing christmas wish for 36h days this year either :-).

My current efforts are now in the [node-tsconfig-paths](https://github.com/jonaskello/node-tsconfig-paths) package. I can get it to work with `node` but not `ts-node`. Specifically, I have a stand-alone example [here](https://github.com/jonaskello/ts-node-paths-example) which installs the `node-tsconfig-paths`package and has two scripts:

```
$ yarn start
$ ts-node -r node-tsconfig-paths main.ts 
Error: Cannot find module 'node-tsconfig-paths'
```

```
$ yarn start:js
$ node -r node-tsconfig-paths js_out/main.js
Hello1
Hello2
Hello3
```

I'll investigate whats going on.

jonaskello:
Btw, awesome-typescript-loader already has this [paths functionality](https://github.com/s-panferov/awesome-typescript-loader/blob/master/src/paths-plugin.ts) for webpack so it might make sense to make a separate resolver package and share it as you mentioned.

jonaskello:
I think I found what is going on, I started #257 rather than discussing it here.

ericmorand:
Just stumbling into this issue and I want to add my voice to this:

> I don't have any desire to enable this in ts-node directly as it'd be promoting poor node.js practices and portability.

paths is bad. It defeats portability and is a hell to maintain.

cyraid:
> Just stumbling into this issue and I want to add my voice to this:
> 
> > I don't have any desire to enable this in ts-node directly as it'd be promoting poor node.js practices and portability.
> 
> paths is bad. It defeats portability and is a hell to maintain.

How does it defeat portability? Used with baseUrl it can separate 'modules'. I use relative paths in my 'paths', combined with extends in my tsconfig.json.

blakeembrey:
> How does it defeat portability?

I assume you don’t share your code on NPM, or changing bundling options, or with other people that don’t also have to enable `paths` support? Portable is subjective here, maybe portable to you is just having someone else also use `ts-node`, but the original goal of the project was that the same code could run as if you did `tsc && node`.

cyraid:
> > How does it defeat portability?
> 
> I assume you don’t share your code on NPM, or use different bundlers, or work with other people that don’t also have `paths` support enabled? Portable is subjective here, maybe portable to you is just having someone else also use `ts-node`, but the original goal of the project was that the same code could run as if you did `tsc && node`, which would mean the same code works everywhere else (Deno, ESBuild, Webpack, etc). For everything else there’s plugins which is how this work got done which is another thing to consider. I probably would have never had the time to invest in `paths` support myself, no one pays me for this.

Don't worry Blake, it's all good. Thank you for your source code contributions! :)

blakeembrey:
To follow up on this, I think @jonaskello would always love help maintaining the paths support 😄 It looks like it's a project as big as `ts-node` on its own, especially for adding ESM support. Also happy to help expose additional hooks that people need to build cool plugins like this.

fatihaziz:
to solve this issue, i have workaround on this issue answer : https://github.com/TypeStrong/ts-node/issues/138#issuecomment-926322588

PR #256, opened 2016-12-28
create script using new Script 

xaclincoln:
createScript is mark deprecated on node.d.ts, and no documentation in (https://nodejs.org/dist/latest-v6.x/docs/api/vm.html). 
All tests passed on Ubuntu 16.

### DIFF
--- a/src/_bin.ts
+++ b/src/_bin.ts
@@ -6,7 +6,7 @@ import Module = require('module')
 import minimist = require('minimist')
 import chalk = require('chalk')
 import { diffLines } from 'diff'
-import { createScript } from 'vm'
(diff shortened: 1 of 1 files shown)


blakeembrey:
Sure, it's marked as deprecated since v0.12 but the old syntax was there to support v0.10.

blakeembrey:
I'm not going to make this change for now as releasing this would mean releasing 2.0, which isn't really needed right now - the deprecated code, as far as I know, is not going away. When it does, we can revisit this.

blakeembrey:
Looks like I'll merge and release this earlier than expected after https://github.com/TypeStrong/ts-node/pull/258 😄

Issue #257, opened 2016-12-28
--require flag does not load from node_modules


jonaskello:
As added in #181 ts-node accepts a --require flag. When trying to use this with the [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) package it fails to locate the package. Example:

```
$ ts-node -r tsconfig-paths main.ts 
Error: Cannot find module 'tsconfig-paths'
```

The reason seems to be [this line](https://github.com/TypeStrong/ts-node/blob/master/src/_bin.ts#L165) which does `Module._load(id)` in order to require the modules specified by `--require`. If it is changed to `require(id)`it will locate the package. It seems like `Module._load(id)` skips loading from `node_modules`. 

The `mocha` package also has a `-r` option and looking [here](https://github.com/mochajs/mocha/blob/master/bin/_mocha#L344) it seems they use `require` to load the modules rather than `Module._load`. So this example works:

```
$ mocha --compilers ts:ts-node/register -r node-tsconfig-paths mocha-example.ts
```

blakeembrey:
We shouldn't technically be doing `require` as that could result in loading our own dependencies instead of the users, which would be a confusing situation. I believe this code was based on the `node` CLI itself here: https://github.com/nodejs/node/pull/881. I'm not sure about the current implementation in node anymore, I'll need to find out how to do a better implementation.

blakeembrey:
Found it. Let's use `_preload_modules` found here: https://github.com/nodejs/node/blob/0fb21df6e692bef9f55b9bfa876f3c59dc590332/lib/internal/bootstrap_node.js#L391. We may need to fallback to `Module._load` on older node.js versions (I can't recall if this ever worked now, I'll check though).

jonaskello:
Nice! I'm collaborating with @jontem on the `node-tsconfig-paths` package and he'll make a PR for this.

blakeembrey:
Looks like this should have been resolved with https://github.com/TypeStrong/ts-node/pull/258 IIRC.

PR #258, opened 2016-12-29
Fixed preloading of modules

Jontem:
Hi!

This is a PR for #257. I've added a unit test which verifies the solution. Unfortunately the typings for node doesn't contain the definition for `_preloadModules` so i had to cast to any. Maybe the typings could be added to? I saw that you authored the [node typings](https://github.com/types/env-node).

### DIFF
--- a/src/_bin.ts
+++ b/src/_bin.ts
@@ -161,9 +161,7 @@ const service = register({
 })
 
 // Require specified modules before start-up.
-for (const id of arrify(argv.require)) {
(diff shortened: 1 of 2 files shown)


jonaskello:
Seems like there is a problem with node v4. Is there some other method that will work there?

Jontem:
@blakeembrey i changed the test case to preload `typescript` instead. Looks good

Issue #225, opened 2016-10-25
Sharing `ts-node` Temp Directory


destroyerofbuilds:
The current behavior of `ts-node` on a Linux system is to store transpiled files under a `/tmp/ts-node/` directory.

When the `ts-node` directory is first created the ownership of that directory is assigned to the user creating it. Furthermore, the permissions are 755 (or `rwxr-xr-x`). 

Meaning that on a multi-user system, only the original creator of the `ts-node` directory can use `ts-node` in its default configuration.

Others will get an error like `Error: EACCES: permission denied, mkdir '/tmp/ts-node/f84fcc3f56f0ac645f6723f0333120790d00c88f`.

I received that error while running `gulp` in a directory containing a `gulpfile.ts` file.

Instead, when a temp directory is created by `ts-node`, can that path be `ts-node-HASH` instead of a hashed directory under `ts-node`?

blakeembrey:
Yes, absolutely. Sounds like a good idea.

blakeembrey:
Actually, I'm an idiot. I can't create a random hash as that's the whole point of the shared temp directory, to speed up the startup when nothing has changed. Do you have any ideas? Perhaps we can have better directory default permissions?

destroyerofbuilds:
@blakeembrey I see what you mean by needing to use the same temp directory across runs to gain the benefit of caching.

I'm not sure how we could adjust the directory permissions. Giving everyone read/write access to `ts-node` would be a security risk (I could inject code into the cache of other projects).

Also, the hash is based on the compiler options, so I believe it's possible for individual user's builds to get the same hash. 

Perhaps getting the username and including that in the path would be sufficient (so `/tmp/ts-node-USERNAME/`)?

blakeembrey:
Sounds good. I'm happy to make any change so that it's still consistent for re-use. It may also be possible to use a hash of the current working directory if consistently getting the username across OS turns out to be tricky (haven't investigated). Feel free to make a PR with whichever option you feel is most suitable.

egandro:
Hi,

I also noticed this issue. So here my 5 cents.

1) On Windows this is no problem.

%TMP% and %TEMP% is always set per user and even in Terminal Server enviorments where a user can have multiple logins this unique. Example:

TEMP=C:\Users\foo\AppData\Local\Temp
TMP=C:\Users\foo\AppData\Local\Temp

2) On Unix / OSX this is an isse.

[...]

PR #267, opened 2017-02-09
Use os.homedir when calculating tmpdir

shortenda:
See: [issue 225](https://github.com/TypeStrong/ts-node/issues/225)

Note that instead of basing off username as I'd said there this is based off the homedir from node os.

I also wasn't sure how best to go about testing this change, if there's any recommendations on that I would happily add tests to accompany.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -150,7 +150,7 @@ export function register (options: Options = {}): () => Register {
   const fast = !!(options.fast == null ? DEFAULTS.fast : options.fast)
   const project = options.project || DEFAULTS.project
 
-  function getTmpDir(): string{
+  function getTmpDir(): string {
     const hash: string = crypto.createHash('sha1')
       .update(homedir(), 'utf8')
       .digest('hex')
--- a/src/index.ts
+++ b/src/index.ts
@@ -134,6 +134,13 @@ export interface Register {
   getTypeInfo (fileName: string, position: number): TypeInfo
 }
 
+function getTmpDir (): string {
+  const hash: string = crypto.createHash('sha1')
+    .update(homedir(), 'utf8')
+    .digest('hex')
+  return join(tmpdir(), 'ts-node-' + hash)
+}
+
 /**
  * Register TypeScript compiler.
  */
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -149,7 +149,15 @@ export function register (options: Options = {}): () => Register {
   const shouldCache = !!(options.cache == null ? DEFAULTS.cache : options.cache)
   const fast = !!(options.fast == null ? DEFAULTS.fast : options.fast)
   const project = options.project || DEFAULTS.project
-  const cacheDirectory = options.cacheDirectory || DEFAULTS.cacheDirectory || join(tmpdir(), 'ts-node')
+
+  function getTmpDir(): string {
```
blakeembrey: Can you move this function outside of the main run loop and add a space after the function name to the parenthesis to match the rest of the project? Also, I'll double check to make sure I didn't miss it, but why base it on the home directory instead of username or cwd?

shortenda: The reason for not using the username was mostly because there is a method to get the homedir provided by node, but not one for username. So I thought this would be the easiest way to get good behavior across platforms. Alternatively to find the username we could look through env variables, or somehow use [this libuv code](http://docs.libuv.org/en/v1.x/misc.html#c.uv_os_get_passwd) (not sure how easy the ffi would be for this...); but neither of these are things I feel like I could do well. I could attempt this if it is desirable though.

Additionally, I think that homedir should be a reasonable option. I don't think that shared homedir's will be a common environment, and even if there are some setups like that this will not be a regression.

I opted not to use the cwd because I thought it would be desirable to share the cache between projects run by the same user. Additionally, two users could run code from the same directory and have conflicting caches. The only case I can think of where this would be superior would be if the user's homedir is empty string, in which case using the cwd would probably be better. Could fall-back to using the cwd if the users homedir is empty string? At the cost of a little more complexity this should provide better behavior.

I will update the code per your comments.

Thanks,
David

blakeembrey: No worries, it seems reasonable. I just wanted to find out why, thanks! 👍  There is a `userInfo` method, but it appears to be new in v6 so won't be very useful to us here.

shortenda:
I've addressed your comments in the most recent commit.

Would you mind if I open a pr in tslint-config-standard to add the rule about space after function names?

blakeembrey:
It looks good, thanks 👍 

> Would you mind if I open a pr in tslint-config-standard to add the rule about space after function names?

Absolutely. The rules all come from the `eslint` collection (https://www.npmjs.com/package/tslint-eslint-rules) and I don't believe that particular one has been implemented yet. Might be an easier one to implement 😄

Issue #283, opened 2017-03-06
Typeroots not inferred with TS2.1+ on Windows


filipesilva:
I have a simple repo at https://github.com/filipesilva/tsnode-typeroots that shows this issue. 

To reproduce, do:
```
git clone https://github.com/filipesilva/tsnode-typeroots
cd tsnode-typeroots
npm install
npm run tsc
npm run tsnode
```

On a OSX machine, both npm scripts will complete successfully. On a Windows machine, `npm run tsnode` will fail compilation with the following error:

```
TSError: ⨯ Unable to compile TypeScript
Cannot find type definition file for 'node'. (2688)
index.ts (1,13): Cannot find name 'module'. (2304)
```

This can be fixed either by:
- installing `typescript@2.0`
- adding `"typeRoots": [ "node_modules/@types" ]` to `compilerOptions` in `tsconfig.json`

I believe this is a bug because the `tsc` compilation itself completes successfully, whereas the `tsnode` one does not.

blakeembrey:
Is this the same issue as the other Windows issues? See https://github.com/TypeStrong/ts-node/issues/216.

filipesilva:
@blakeembrey I'm not sure if it's the same. I agree that it seems similar, but in my case I also see this:
```
Cannot find type definition file for 'node'. (2688)
```

Which seems to indicate that the `node` types are being looked for explicitly and not found.

I also tried adding jasmine+types into the example and updating it, and it seems to not be specific to `node` typings as the original poster in #216 hypothesized.

```
kamik@T460p MINGW64 /d/sandbox/tsnode-typeroots (master)
$ npm run tsc

> tsnode-typeroots@1.0.0 tsc D:\sandbox\tsnode-typeroots
> tsc -p ./tsconfig.json


kamik@T460p MINGW64 /d/sandbox/tsnode-typeroots (master)
$ npm run tsnode

> tsnode-typeroots@1.0.0 tsnode D:\sandbox\tsnode-typeroots
> ts-node -P ./tsconfig.json index.ts


D:\sandbox\tsnode-typeroots\node_modules\ts-node\src\index.ts:319
          throw new TSError(formatDiagnostics(diagnosticList, cwd, ts, lineOffset))
                ^
TSError: ⨯ Unable to compile TypeScript
Cannot find type definition file for 'jasmine'. (2688)
Cannot find type definition file for 'node'. (2688)
index.ts (1,13): Cannot find name 'module'. (2304)
[...]

Issue #216, opened 2016-10-11
"Cannot find module 'http'. (2307)" with "@types/node" under typescript@2.0.3


zixia:
**TypeScript Version:**  

2.0.3

**Code**

``` ts
// A *self-contained* demonstration of the problem follows...
// t.ts
import * as http from 'http'
console.log(typeof http)
```

``` shell
$ ./node_modules/.bin/ts-node -v
ts-node v1.4.1
node v6.0.0

$ ./node_modules/.bin/tsc -v     
Version 2.0.3

$ npm install @types/node
wechaty@0.4.0 /Users/zixia/git/wechaty
└── @types/node@6.0.45

$ ./node_modules/.bin/ts-node --no-cache t.ts

D:\cygwin64\home\zixia\git\wechaty\node_modules\ts-node\dist\index.js:166
                    throw new TSError(diagnosticList);
                          ^
TSError: ⨯ Unable to compile TypeScript
t.ts (1,23): Cannot find module 'http'. (2307)
```

**Expected behavior:**

should import http as well

**Actual behavior:** 

TS2307 error

**Another STRANGE behavior**

if I add another @types into t.ts, it will work without any problem.

``` shell
$ cat t.ts
import * as http from 'http'
console.log(typeof http)

import * as express from 'express'
console.log(typeof express)

zixia@zixia-desktop ~/git/wechaty
$ ./node_modules/.bin/ts-node --no-cache t.ts
object
function
```

very interesting, what's wrong with me?

ljani:
I have the same problem. It seems I can workaround it using the following directive at the top of my file (which is located in `./src/`):
```
///<reference path="../node_modules/@types/node/index.d.ts"/>
```

Running `tsc` works fine. `ts-node` seems to be missing the global references `tsc` has.

paulsouche:
Hi,

I had the same issue with mocha on #246

The problem is that module http is ambient declared by node declaration file. What ts-node do is
 - find http directory in @types folder
 - no directory - Error

My first solution was to import module mocha

```typescript
import * as mocha from 'mocha';
```

To have all mocha declarations available but I did not know that we could declare an ambient module via

```typescript
/// <reference types="mocha" />
```

Declaring mocha as ambient in tsconfig.json did not work for me

```json
{
  "compilerOptions": {
    "types": ["mocha"]
  }
}
```

I don't know why exactly because it should be equivalent...

Again I faced a new issue not declaring these references outside of my tests in a separated file with only one spec file.

[...]

PR #285, opened 2017-03-07
Normalize slashes for windows paths

Jontem:
This PR solves the problem discussed in #283 and should also solve #216. I've only tested on a minimal repro on windows. There already was a function named `slash` that was exactly implemented as `normalizeSlashes` that typecript exports. So i just used that one.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -445,14 +445,15 @@ function readConfig (compilerOptions: any, project: string | boolean | undefined
   delete result.config.compilerOptions.outFile
   delete result.config.compilerOptions.declarationDir
 
-  const basePath = result.path ? normalizeSlashes(dirname(result.path)) : cwd
+  const normalizedConfigPath = result.path && normalizeSlashes(result.path)
+  const basePath = normalizedConfigPath ? dirname(normalizedConfigPath) : normalizeSlashes(cwd)
 
   if (typeof ts.parseConfigFile === 'function') {
     return ts.parseConfigFile(result.config, ts.sys, basePath)
   }
 
   if (typeof ts.parseJsonConfigFileContent === 'function') {
-    return ts.parseJsonConfigFileContent(result.config, ts.sys, basePath, null, normalizeSlashes(result.path as string))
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -445,14 +445,14 @@ function readConfig (compilerOptions: any, project: string | boolean | undefined
   delete result.config.compilerOptions.outFile
   delete result.config.compilerOptions.declarationDir
 
-  const basePath = result.path ? dirname(result.path) : cwd
+  const basePath = result.path ? normalizeSlashes(dirname(result.path)) : cwd
```
blakeembrey: It looks like you should be normalizing the slashes on both sides, not just the left side of the ternary.

src/index.ts
```diff
@@ -574,6 +574,7 @@ export function fileExists (fileName: string): boolean {
  * Get directories within a directory.
  */
 export function getDirectories (path: string): string[] {
+  console.log(`getDirectories: ${path}`)
```
Jontem: Oops. Fixed :)

src/index.ts
```diff
@@ -445,14 +445,15 @@ function readConfig (compilerOptions: any, project: string | boolean | undefined
   delete result.config.compilerOptions.outFile
   delete result.config.compilerOptions.declarationDir
 
-  const basePath = result.path ? dirname(result.path) : cwd
+  const normalizedConfigPath = result.path && normalizeSlashes(result.path)
```
Jontem: I realized we need to handle when no `tsconfig.json` is found. Can't do replace on undefined. So fixed that to.

src/index.ts
Jontem: Fixed. Interesting note is that without a `tsconfig.json` this problem did not exists for me. But as soon as i added a `tsconfig.json` i got the error. It seems that typescript handles the `basePath` differently as cwd also uses windows path separator. Either way the best would be to treat them the same.

massimocode:
@Jontem thanks for picking this up! 👍

PR #321, opened 2017-05-03
Delete config options after config has been parsed by typescript

### LINKED ISSUE #320: Error "declarationDir" cannot be specified without "declaration" - when "declarationDir" is set within extended config
I have a project that has the following configurations:

tsconfig.json:
```
{ extends: "./node_modules/@our-company/build-toolchain/tsconfig.json" }
```

@our-company/build-toolchain/tsconfig.json:
```
{
  "compileOnSave": true,
  "compilerOptions": {
    "strict": true,
    "forceConsistentCasingInFileNames": true,
    "declaration": true,
    "lib": [ "es6" ],
    "moduleResolution": "node",
    "module": "commonjs",
    "noEmitOnError": true,
    "noFallthroughCasesInSwitch": true,
    "noImplicitReturns": true,
    "preserveConstEnums": true,
    "target": "es5",
    "pretty": true,
    "sourceMap": true,
    "rootDir": "../../../lib",
    "outDir": "../../../dist",
    "declarationDir": "../../../dist"
  },
  "include": [
    "../../../lib/**/*"
  ]
}
```

These configurations result in the following error:
```
Option 'declarationDir' cannot be specified without specifying option 'declaration'. (5052)
    at getOutput (/.../node_modules/ts-node/src/index.ts:312:15)
    at /.../node_modules/ts-node/src/index.ts:343:16
    at Object.compile (/.../node_modules/ts-node/src/index.ts:487:17)
    at Module.m._compile (/.../node_modules/ts-node/src/index.ts:395:43)
    at Module._extensions..js (module.js:579:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/.../ts-node/src/index.ts:398:12)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Function.Module.runMain (module.js:604:10)
```

After digging around in the code, it appears that the `declaration` property is getting overridden to false, but the declarationDir property is just getting deleted from the config (which ends up not impacting the extended configuration. [Relevant Lines: index.ts:419](https://github.com/TypeStrong/ts-node/blob/54128f26bd6d99770a6c2d0fb9f0848d7bf8ae2b/src/index.ts#L419-L421)

danthegoodman:
This ensures that configurations that came from inherited/extended tsconfig files do not override our needs.

Fixes #320

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -415,23 +415,30 @@ function readConfig (compilerOptions: any, project: string | boolean | undefined
     outDir: '$$ts-node$$'
   })
 
-  // Delete options that *should not* be passed through.
-  delete result.config.compilerOptions.out
-  delete result.config.compilerOptions.outFile
-  delete result.config.compilerOptions.declarationDir
-
   const configPath = result.path && normalizeSlashes(result.path)
   const basePath = configPath ? dirname(configPath) : normalizeSlashes(cwd)
 
   if (typeof ts.parseConfigFile === 'function') {
+    deleteConfigOptionsThatShouldNotBePassedThrough(result.config.compilerOptions)
     return ts.parseConfigFile(result.config, ts.sys, basePath)
   }
 
   if (typeof ts.parseJsonConfigFileContent === 'function') {
-    return ts.parseJsonConfigFileContent(result.config, ts.sys, basePath, undefined, configPath as string)
+    let finalResult = ts.parseJsonConfigFileContent(result.config, ts.sys, basePath, undefined, configPath as string)
+    deleteConfigOptionsThatShouldNotBePassedThrough(finalResult.options)
+    return finalResult
   }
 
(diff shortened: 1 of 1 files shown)


danthegoodman:
@blakeembrey I'm not sure if the fix I just submitted it what you had in mind, but it turned out all that was needed was to delete the properties after typescript performs it's config parsing.

blakeembrey:
Similar. We'd also need to move the `Object.assign` override. I'd prefer `const` usage, but we should probably just write this in a way we can do `tidyConfig(ts.parseConfig(...))`. I can tidy it up though 😄

danthegoodman:
@blakeembrey Sweet! I just pulled down your changes and tried them against my project and it all worked! 

Thank you for cleaning up my commit too. `tidyConfig` is a nicer name than that verbose monstrosity I settled with. 😄

Issue #322, opened 2017-05-04
Cache defaults to disabled when run via ts-node/register


chrisleck:
When run via the command line, `ts-node` defaults the `cache` property to `true`.  https://github.com/TypeStrong/ts-node/blob/master/src/_bin.ts#L104

When run inside mocha via `ts-node/register`, cache defaults to `false`.
https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L143

I believe caching should be enabled by default in both cases.  This is causing our team's mocha tests to take a really long time.  Forcing caching on speeds up testing by an order of magnitude or two.

blakeembrey:
Probably a good idea to make it consistent. I'd just correct the default option here: https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L83. Feel free to submit a PR, something like `process.env['TS_NODE_CACHE'] ? yn(...) : true` might be reasonable.

blakeembrey:
Fixed with https://github.com/TypeStrong/ts-node/pull/323.

PR #323, opened 2017-05-04
Enable caching by default when running ts-node via `ts-node/register`

chrisleck:
* Fixes issue #322 

@blakeembrey PTAL

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -80,7 +80,7 @@ export interface TypeInfo {
 const DEFAULTS = {
   getFile,
   fileExists,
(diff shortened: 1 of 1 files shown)


blakeembrey:
LGTM 👍 I made a tiny request for https://github.com/sindresorhus/yn/issues/10 to make this simpler and handle when `yn()` fails to parse without adding unnecessary complexity to that line.

chrisleck:
Thanks, Blake.  We appreciate the ultra fast response and turnaround on this issue.  One of the most pleasurable interactions I've had in the OSS world!

PR #324, opened 2017-05-06
Adds example repo of programmatic usage

realappie:
This would help a lot to anyone setting up new projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ of the script. From here you can use `require` to bring in modules from TypeScri
 var someModule = require('path_to_a_typescript_file');
 ```
 
+Checkout this [example](https://github.com/realappie/nodejs-typescript) repository
+
 ## License
 
 MIT


blakeembrey:
Can you detail how this would help or what was lacking that this improves? I'd prefer to improve the documentation in this repo or add an examples folder here over linking to a third party I don't control.

realappie:
I personally had a hard time figuring out how to use ts-node programmatically, luckily one of my colleagues who took the time to figure it out before explained me how to use it.

The example repo would serve as a guide to anyone trying to setup a new project, and although it seems very straight forward once you have setup a project before. An example would help anyone new to ts-node.

> I'd prefer to improve the documentation in this repo or add an examples folder here over linking to a third party I don't control.

I understand why you would prefer that. If you want, I would submit another PR of my example repo under an examples folder.

blakeembrey:
> took the time to figure it out before explained me how to use it.

Can we improve on the documentation with whatever he said? I just don't see much `ts-node` related other than https://github.com/realappie/nodejs-typescript/blob/ce46e5ed40aafbb292a5698f5b4964f0a3e2af65/server.js#L1, and I'd recommend just using `ts-node` or `node -r ts-node/require` instead. I'm not sure why `ts-node` is being used programmatically there.

realappie:
What he said was the line you references to, the example serves as a boilerplate library anybody can  immediately start with.

Also, why do you recommend using ts-node through the command line? Beside performance advantages.

blakeembrey:
But that's already in the documentation: https://github.com/TypeStrong/ts-node#programmatic-usage.

I only recommend it so you don't have to create a `.js` file, you can immediately point to a `.ts` file as the entry point instead because it's already registered. That's my personal preference anyway.

realappie:
@blakeembrey I actually agree with you, the example repo isn't related enough to be part of this repo. Its more of a light boilerplate project for anyone wanting building for nodeJS in typescript.

PR #352, opened 2017-06-12
updated package.json devDependency

EsrefDurna:
updated package proxyquire from 1.7.2 to 1.8.0
 Npm link :  https://www.npmjs.com/package/proxyquire
 Repository link : https://github.com/thlorenz/proxyquire/releases

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "istanbul": "^0.4.0",
     "mocha": "^3.0.0",
     "ntypescript": "^1.201507091536.1",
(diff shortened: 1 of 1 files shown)


blakeembrey:
Hey! We don't need to update every time there's a new release. The `^` means it'll install the latest version of `1.x` anyway, so it would install `1.8.0` anyway if that's the latest release. All these repos also use Greenkeeper (see https://github.com/TypeStrong/ts-node#typescript-node) which updates the versions automatically when required.

PR #356, opened 2017-06-14
Fix issue 355

chrisleck:
Verify a cache file has non-zero size before considering it valid.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -544,7 +544,8 @@ export function fileExists (fileName: string): boolean {
   try {
     const stats = statSync(fileName)
 
-    return stats.isFile() || stats.isFIFO()
+    // Ensure the file size is non-zero to avoid racing against cache users in other processes.
+    return (stats.isFile() && stats.size > 0) || stats.isFIFO()
   } catch (err) {
     return false
   }


chrisleck:
Could we get this reviewed, submitted, and pushed out?  My team continues to struggle with this issue.

blakeembrey:
Also, as it is, I don't think I can accept this change. This function is used in many places - not just checking the cache. Accepting this would mean a real empty file would create an error in the TypeScript compiler, I believe - feel free to test it though.

chrisleck:
I see two uses of this function - one for the cache and the other for implementing the interface into TS.  I could fork the method such that we only do the zero check for the cache lookup...would that be acceptable to you?  If a cache file is legitimately zero-length, then the cache would never report an entry for that file which seems ok?

blakeembrey:
Did you want to try the latest release which merged the two cache entries into one? You can fork it and wrap it if you'd like, but it may not back compatible with the existing interface either since the implementations can be passed in. I might need to think on it, but you can pass it in your own register for now as a workaround.

chrisleck:
I tried the latest release and it seems to have gotten worse - more complaints about null constructors than usual :)
In any case, I'll go ahead and close this one given the changes in the latest release - will read through the new code and see if I can come up with something.

PR #363, opened 2017-06-19
Check map file exists as well as compile JS target files

### LINKED ISSUE #362: Uncaught errors when target JS file exists and MAP file doesn't exist
> Error: ENOENT: no such file or directory, open '/var/folders/44/40437cm97md6b8zn0rs8fy5mhhh0hc/T/ts-node-17b38112533313593deb1e0606cb377b3ed838e2087adea7b5cc097b01b8f44a/639d71863f6c7f45f1df4ada0591cabc1ce103b9a28c16ddb08fbb10f4c3646f/8421e53249316d93f9faaf989bf056d61457a056cb0531fa3bd39aa8c9198bed.js.map'

This is a problem I faced today and spent couple of hours working on. it looks like source map file generation failed for some reason for me. After that my ts-node process was working fine until error occured. It was trying to build the stack trace from source maps: the cache.sourceMaps contained information about the file but the map file didn't exist in the file system: https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L172

The problem is that fileExists only checks the target file: https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L485

1999:
close #362 

The problem here is that second fs.writeFileSync can fail for some reason but the target compiled JS file would be built. In this case it will be used which leads to #362

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -482,7 +482,7 @@ function readThrough (
     cache.sourceMaps[fileName] = sourceMapPath
 
     // Use the cache when available.
-    if (fileExists(outputPath)) {
+    if (fileExists(outputPath) && fileExists(sourceMapPath)) {
       return getFile(outputPath)
     }
 


1999:
It looks like Travis CI error is not related to this change.

blakeembrey:
Thanks! The CI issue was fixed by someone else today (thanks!) and I did https://github.com/TypeStrong/ts-node/pull/368 in favour of this change. I'll do a new minor release today with both these new features!

PR #366, opened 2017-06-20
Fix service host interface to support ts2.5

cevek:
In typescript 2.5 language service host needs to implement fileExists, readFile and readDirectory to be able to correctly resolve the modules (otherwise only ambient modules would be resolved)
Described here https://github.com/Microsoft/TypeScript/blame/master/src/services/types.ts#L166


src/index.ts
```diff
@@ -278,6 +278,9 @@ export function register (options: Options = {}): Register {
       },
       getDirectories: getDirectories,
       directoryExists: directoryExists,
+      fileExists: ts.sys.fileExists,
+      readFile: ts.sys.readFile,
+      readDirectory: ts.sys.readDirectory,
```
blakeembrey: We have existing implementations of two of these functions, should we just use those?

cevek: @blakeembrey Fixed

johnnyreilly:
@blakeembrey - do you know if we need to make a similar change to ts-loader here: https://github.com/TypeStrong/ts-loader/blob/master/src/servicesHost.ts#L34 ?

blakeembrey:
@johnnyreilly I would think so, yes. You can just use the `ts.sys` functions though.

johnnyreilly:
Thanks @blakeembrey - I've tested and it looks good.

PR #392, opened 2017-07-14
docs(project-option): make description less specific and say explicitly that you can pass directory in this option

sharikovvladislav:


blakeembrey:
What about something shorter (and keeping the "also" section toward the end?):

> Path to load TypeScript configuration from (either JSON file or directory containing `tsconfig.json`, or `false` to disable) (also `process.env.TS_NODE_PROJECT`)

sharikovvladislav:
@blakeembrey I fixed it.

PR #394, opened 2017-07-20
Check cached files for valid sourcemaps

chrisleck:
@blakeembrey How does something like this look for the sourcemap check?  Not sure what you'd like to do regarding default behaviors - should we make this check optional?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -58,7 +58,6 @@ export interface Options {
   fast?: boolean | null
   cache?: boolean | null
   cacheDirectory?: string
-  cacheVerification?: boolean | null
   compiler?: string
   project?: boolean | string
   ignore?: boolean | string | string[]
@@ -94,7 +93,6 @@ const DEFAULTS = {
   fileExists,
   cache: yn(process.env['TS_NODE_CACHE'], { default: true }),
   cacheDirectory: process.env['TS_NODE_CACHE_DIRECTORY'],
-  cacheVerification: yn(process.env['TS_NODE_CACHE_VERIFICATION'], { default: false }),
   disableWarnings: yn(process.env['TS_NODE_DISABLE_WARNINGS']),
   compiler: process.env['TS_NODE_COMPILER'],
   compilerOptions: parse(process.env['TS_NODE_COMPILER_OPTIONS']),
@@ -156,7 +154,6 @@ export function register (options: Options = {}): Register {
   const getFile = options.getFile || DEFAULTS.getFile
   const fileExists = options.fileExists || DEFAULTS.fileExists
   const shouldCache = !!(options.cache == null ? DEFAULTS.cache : options.cache)
-  const shouldVerifyCache = !!(options.cacheVerification == null ? DEFAULTS.cacheVerification : options.cacheVerification)
   const fast = !!(options.fast == null ? DEFAULTS.fast : options.fast)
   const project = options.project || DEFAULTS.project
   const cacheDirectory = options.cacheDirectory || DEFAULTS.cacheDirectory || getTmpDir()
@@ -250,7 +247,6 @@ export function register (options: Options = {}): Register {
   let compile = readThrough(
     cachedir,
     shouldCache,
-    shouldVerifyCache,
     getFile,
     cache,
     getOutput,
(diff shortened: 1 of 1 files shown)


blakeembrey:
Looks like a good start. Couple of comments:

1. Let's make it always default on
2. What do you think of checking just the final character instead of `indexOf`? It should always end in the same thing since it's always a base 64 `}` (correct me if I'm wrong). Something like `/(?:9|0=|Q==)$/`?

src/index.ts
```diff
@@ -275,7 +279,12 @@ export function register (options: Options = {}): Register {
             return undefined
           }
 
-          cache.contents[fileName] = getFile(fileName)
+          const content = getFile(fileName)
+          if (shouldVerifyCache && !isValidCacheContent(content)) {
```
blakeembrey: This isn't coming from the cache, it's coming from your original source file (e.g. this is your TypeScript, so should never pass this check).

chrisleck: Ah- didn't realize we're caching both input (TS) and output (transpiled JS) files.

chrisleck:
Base64 only pads the end with `=` if necessary.  There's no guarantees on the last character AFAIK.

chrisleck:
I can simply strip out the flag entirely if you think we should always do this check?

chrisleck:
Updated.  Please take another look when you get a chance.

src/index.ts
blakeembrey: We aren't caching input - this code isn't reading from the cache.

blakeembrey:
> Base64 only pads the end with = if necessary. There's no guarantees on the last character AFAIK.

I know that, see the original comment and regex. It'll always be a consistent final character excluding the `=` sign so there's only three possible combinations.

> I can simply strip out the flag entirely if you think we should always do this check?

Yes, absolutely.

src/index.ts
chrisleck: I meant 'in-memory' cache here...different cache.

blakeembrey: Right, I see. Yeah, the input files are cached because they are read multiple times by TypeScript. I would love to improve this, but it's possible to end up inconsistent since the node.js loader can generate a different file _technically_ than the original source file. But the initial source file is required for TypeScript to work (importing types). Maybe it's worth experimenting whether we can go through `require` here instead, but I have a feeling that will fail with anything that needs to go _after_ `ts-node` to work such as code coverage.

chrisleck:
Regarding the regex, this SO thread seems to indicate it's much more permissive: https://stackoverflow.com/questions/8571501/how-to-check-whether-the-string-is-base64-encoded-or-not

`([A-Za-z0-9+/]{4}|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)$ means the string ends in one of three forms: [A-Za-z0-9+/]{4}, [A-Za-z0-9+/]{3}= or [A-Za-z0-9+/]{2}==`

I imagine perfectly good partial source without a sourceMap could match that regex as well.

blakeembrey:
Sorry, I'm not trying to validate that it's valid base 64, only that the final character looks like it's probably `}` (technically there's a chance it's another character by only checking the final chars, but it's unlikely to be corrupt in any of those cases).

chrisleck:
This does seem to work properly on my big repo.  No more loading errors with this change so far.

chrisleck:
Updates: Removed the flag and now sniffing for `}` character via your regex.

src/index.ts
```diff
@@ -536,6 +538,14 @@ function getCacheName (sourceCode: string, fileName: string) {
 }
 
 /**
+ * Ensure the given cached content is valid by sniffing for a base64 encoded '}'
+ * at the end of the content, which should exist if there is a valid sourceMap present.
+ */
+function isValidCacheContent (content: string) {
+  return content.match(/(?:9|0=|Q==$)/)
```
blakeembrey: Please use `RegExp#test` instead of `String#match` (since we want `true` or `false`, not results). Also, you've got a small typo - need to put the `$` outside of the matching group - right now it's matching slightly incorrectly (any `9`, `0=` or `Q==` at the end). Since people can be iffy with regexps like this (we've already got one mistake just by a typo) we could change it to a string match like `content.slice(-1) === ...` and so on. Either way is good with me.

chrisleck: Good catch.  How's this look?

PR #413, opened 2017-08-25
Migrate from typings to @types packages

devoto13:
Allows to import this package into another TypeScript file and get proper type definitions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,13 @@
   "version": "3.3.0",
   "description": "TypeScript execution environment and REPL for node",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "bin": {
     "ts-node": "dist/bin.js",
     "_ts-node": "dist/_bin.js"
   },
   "files": [
     "dist/",
-    "typings.js",
     "register.js",
     "LICENSE"
   ],
@@ -21,7 +21,7 @@
     "test-spec": "mocha dist/**/*.spec.js -R spec --bail",
     "test-cov": "istanbul cover node_modules/mocha/bin/_mocha -- \"dist/**/*.spec.js\" -R spec --bail",
     "test": "npm run build && npm run lint && npm run test-cov",
-    "prepublish": "typings install && npm run build"
+    "prepublish": "npm run build"
   },
   "engines": {
     "node": ">=4.2.0"
@@ -49,7 +49,11 @@
   },
   "homepage": "https://github.com/TypeStrong/ts-node",
   "devDependencies": {
+    "@types/chai": "^4.0.4",
+    "@types/mocha": "^2.2.42",
+    "@types/proxyquire": "^1.3.28",
     "@types/react": "^16.0.2",
+    "@types/semver": "^5.3.34",
     "chai": "^4.0.1",
     "istanbul": "^0.4.0",
     "mocha": "^3.0.0",
@@ -60,10 +64,18 @@
     "semver": "^5.1.0",
     "tslint": "^5.0.0",
     "tslint-config-standard": "^6.0.1",
-    "typescript": "^2.4.1",
-    "typings": "^2.0.0"
+    "typescript": "^2.4.1"
   },
   "dependencies": {
+    "@types/arrify": "^1.0.1",
+    "@types/chalk": "^0.4.31",
+    "@types/diff": "^3.2.1",
+    "@types/minimist": "^1.2.0",
+    "@types/mkdirp": "^0.5.0",
+    "@types/node": "^8.0.27",
+    "@types/source-map-support": "^0.4.0",
+    "@types/v8flags": "types/npm-v8flags#de224ae1cd5fd7dbb4e7158a6cc7a29e5315930d",
+    "@types/yn": "types/npm-yn#ca75f6c82940fae6a06fb41d2d37a6aa9b4ea8e9",
     "arrify": "^1.0.0",
     "chalk": "^2.0.0",
     "diff": "^3.1.0",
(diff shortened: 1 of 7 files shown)


blakeembrey:
We can't enable this until it's migrated away from Typings: https://github.com/TypeStrong/ts-node/blob/master/typings.json.

blakeembrey:
I'm happy to have the package replace Typings, I just haven't had the time, so I'll re-open the PR in case you'd like to tackle it.

devoto13:
I've updated most of the packages. We need what to do with packages, which don't have available `@types` packages and don't bundle type definitions.

These are missing types:
- `tsconfig` (I see you just have migrated it)
- `v8flags`
- `proxyquire`
- `yn`

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "lib": ["es2015"],
+    "lib": ["es2015", "dom"],
```
devoto13: `@types/react` depends on DOM APIs. We should either specify all types packages [explicitly](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types) in this tsconfig to prevent `react` from being included or add DOM APIs as it is done now.

blakeembrey: I'd rather be specific than include DOM APIs since this project doesn't use React or the DOM directly.

blakeembrey:
Sounds good. An alternative we can do is just to install from GitHub if we add `package.json` files to each of these projects.

tsconfig.json
devoto13: Okey, will do it than.

devoto13:
Good idea with installing from GitHub! Adding `package.json` was actually not necessary. 

Remaining type errors will require updating type definition for Node. Will work on it tomorrow.

tsconfig.json
```diff
@@ -13,10 +13,24 @@
     "moduleResolution": "node",
     "sourceMap": true,
     "inlineSources": true,
-    "types": []
+    "types": [
```
blakeembrey: You should only need to put global types here, not all of them, so it should work using just `node`.

package.json
```diff
@@ -60,10 +64,18 @@
...
   },
   "dependencies": {
+    "@types/arrify": "^1.0.1",
+    "@types/chalk": "^0.4.31",
+    "@types/diff": "^3.2.1",
+    "@types/minimist": "^1.2.0",
+    "@types/mkdirp": "^0.5.0",
+    "@types/node": "^6.0.88",
+    "@types/source-map-support": "^0.4.0",
+    "@types/yn": "types/npm-yn",
```
blakeembrey: Can we put the commit hash for the GitHub installed dependencies? Otherwise we might get a major version that's incompatible later and we won't want that to break everything 😄

src/types.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'tsconfig'
```
blakeembrey: Released a major version of `tsconfig` now and we can update to it 👍

package.json
devoto13: Sure!

tsconfig.json
devoto13: Also `mocha` is needed.

blakeembrey: Right, for global test functions 👍  Good catch.

devoto13:
I have prepared update for Node type definitions. Take a look [here](https://github.com/devoto13/DefinitelyTyped/commit/f2627839e9175a40665d65d161b6b8f5cca6e194). I think it worths bumping `@types/node` to 8.* to not mess up with type definitions for older versions and CI is running Node 8 as well. 

I decided to add extra Module class to avoid breaking people, who depend on `NodeModule` being an interface. And also unsure about including private variables/methods. Maybe we should access them by casting to `any` instead. What do you think?

src/index.ts
```diff
@@ -12,7 +12,7 @@ import * as TS from 'typescript'
 import { loadSync } from 'tsconfig'
 
 const pkg = require('../package.json')
-const shouldDebug = yn(process.env.TS_NODE_DEBUG)
+const shouldDebug = yn(process.env.TS_NODE_DEBUG!)
```
devoto13: Just to get it working for now. Type definitions fro `yn` should be updated instead as environment variables can be missing.

src/index.spec.ts
```diff
@@ -79,6 +79,10 @@ describe('ts-node', function () {
 
     it('should throw errors', function (done) {
       exec(`${BIN_EXEC} -e "import * as m from './tests/module';console.log(m.example(123))"`, function (err) {
+        if (err === null) {
```
devoto13: Since in newer Node type definitions `err` can be `null`. It was required to add check to satisfy `strictNullChecks`.

devoto13:
Waiting for https://github.com/types/npm-yn/pull/17 to be merged and it should be ready for review. 

Let me know what you think about casting to `any` to access private methods. It's not recommended to add private APIs on DefinitelyTyped, so I added only public APIs.

blakeembrey:
> It's not recommended to add private APIs on DefinitelyTyped, so I added only public APIs.

That's a bummer, it is the node.js API and hasn't changed for years (at least since 0.8 for most of those, a couple since 4.0). How is one meant to type check between node versions for things like this?

I'm ok casting it to `any`, but would prefer pulling the required changes in locally.

devoto13:
> That's a bummer, it is the node.js API and hasn't changed for years (at least since 0.8 for most of those, a couple since 4.0). How is one meant to type check between node versions for things like this?

I can see their reasons. Using private APIs is always "at your own risk". And even if these APIs don't change often, others do. So deciding which should be allowed and which should not gonna be a mess.

I have rebased and squashed commits. I can't pull in missing types locally because of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/19612. The problem is that it's not possible to augment symbol exported with `export =` syntax.

We can either merge "as is" and add missing types in a separate PR or wait for linked PR to be merged.

PR #419, opened 2017-09-07
forward signals correctly (was: signals: ignore them)

stelcheck:
signals are passed to the spawned child process; this means
that the main process should not react to them, and instead
wait for the spawned process to exit as a result of receiving
the forwarded signal.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -61,34 +61,24 @@ v8flags(function (err, v8flags) {
       //
       // See: https://nodejs.org/api/child_process.html#child_process_options_detached
       detached: true,
-
-      // Pipe all input and output to this process
       stdio: 'inherit'
     }
   )
 
-  // Ignore signals, and instead forward them to the
-  // child process
-  const forward = (signal: NodeJS.Signals) => process.on(signal, () => proc.kill(signal))
+  // Ignore signals, and instead forward them to the child process.
+  ;['SIGINT', 'SIGTERM', 'SIGWINCH'].forEach(signal => {
+    process.on(signal, () => proc.kill(signal))
+  })
 
-  // Interrupt (CTRL-C)
-  forward('SIGINT')
-
-  // Termination (`kill` default signal)
-  forward('SIGTERM')
-
-  // Terminal size change must be forwarded to the subprocess
-  forward('SIGWINCH')
-
-  // On exit, exit this process with the same exit code
+  // On spawned close, exit this process with the same code.
   proc.on('close', (code: number, signal: string) => {
     if (signal) {
       process.kill(process.pid, signal)
-    } else if (code) {
+    } else {
       process.exit(code)
     }
   })
 
-  // If this process is exited, kill the child first
-  process.on('exit', (_code: number) => proc.kill())
+  // If this process exits, kill the child first.
+  process.on('exit', () => proc.kill())
 })
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -5,6 +5,7 @@ import { join } from 'path'
 import v8flags = require('v8flags')
 
 const argv = process.argv.slice(2)
+const signals: NodeJS.Signals[] = ['SIGINT', 'SIGTERM', 'SIGWINCH']
 
 v8flags(function (err, v8flags) {
   if (err) {
@@ -66,9 +67,7 @@ v8flags(function (err, v8flags) {
   )
 
   // Ignore signals, and instead forward them to the child process.
-  ;['SIGINT', 'SIGTERM', 'SIGWINCH'].forEach(signal => {
-    process.on(signal, () => proc.kill(signal))
-  })
+  signals.forEach(signal => process.on(signal, () => proc.kill(signal)))
 
   // On spawned close, exit this process with the same code.
   proc.on('close', (code: number, signal: string) => {
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -53,16 +53,42 @@ v8flags(function (err, v8flags) {
   const proc = spawn(
     process.execPath,
     nodeArgs.concat(join(__dirname, '_bin.js'), scriptArgs),
-    { stdio: 'inherit' }
+    {
+      // We need to run in detached mode so to avoid
+      // automatic propagation of signals to the child process.
+      // This is necessary because by default, keyboard interrupts
+      // are propagated to the process tree, but `kill` is not.
+      //
+      // See: https://nodejs.org/api/child_process.html#child_process_options_detached
+      detached: true,
+
+      // Pipe all input and output to this process
+      stdio: 'inherit'
+    }
   )
 
-  proc.on('exit', function (code: number, signal: string) {
-    process.on('exit', function () {
-      if (signal) {
-        process.kill(process.pid, signal)
-      } else {
-        process.exit(code)
-      }
-    })
+  // Ignore signals, and instead forward them to the
+  // child process
+  const forward = (signal: NodeJS.Signals) => process.on(signal, () => proc.kill(signal))
+
+  // Interrupt (CTRL-C)
+  forward('SIGINT')
+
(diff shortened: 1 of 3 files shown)


blakeembrey:
Can you share the situation that this solves? I've never come across this being an issue myself, nor have I seen it handled in other libraries that spawn child processes (https://github.com/babel/babel/blob/777a9ae6e42db1c9d57bbbdca8b233c2dc9b89ac/packages/babel-node/src/babel-node.js).

stelcheck:
Given the following code:

¨¨¨typescript
process.on('SIGINT', () => {
  console.log('received sigint')
  setTimeout(() => {
    console.log('exit')
    process.exit(0)
  }, 2000)
})

console.log('running')
require('http').createServer(() => true).listen(8080)
¨¨¨

Running directly in `node`:

¨¨shell
$ node index.ts   
running
^Creceived sigint
exit
$
¨¨¨¨

While in `ts-node`
¨¨¨shell
ts-node index.ts
running
^Creceived sigint

$ exit

¨¨

As you can see, with `ts-node`, the `exit` log is shown after the process has been terminated. This is because the `ts-node` process need only to forward the signals, not react to them.

`npm` also suffered from the same problem until 4.5.0, for what it is worth: https://github.com/npm/npm/commit/878aceb25e6d6052dac15da74639ce274c8e62c5

blakeembrey:
Awesome, thanks. Do you think we get add a test case so this doesn't occur in the future?

stelcheck:
I wrote a quick sample test, but for some reason, the child process does not get killed *at all*. Will look into it a bit further.

src/index.spec.ts
```diff
@@ -20,6 +20,17 @@ describe('ts-node', function () {
...
   describe('cli', function () {
+    it('should not react to signals, but forward them to the child process', function (done) {
+      const proc = exec(`${BIN_EXEC} tests/signals`, function (err, stdout) {
+        expect(err).to.equal(null)
+        expect(stdout).to.equal('exit')
+
+        return done()
+      })
+
+      setTimeout(() => proc.kill('SIGTERM'), 100)
```
stelcheck: For some reasons, the kill signal is not properly sent.

blakeembrey: Would it work if you forwarded the signals to the child? It's possible that the current code change is just ignoring the signal.

stelcheck:
The point of this test is to ensure that the signal is properly caught by the binary, not by the child. 

I tried to use 'SIGKILL', and that didn't work either, so maybe Mocha is getting in the way...

stelcheck:
Found the issue. In a nutshell, signals triggered by keyboard interrupts are sent to the process and all subprocesses at once, except if the suprocess was spawned in `detached` mode. 

The solution is to detach the subprocess and manually forward relevant signals. I have found three so far:

  - SIGINT
  - SIGTERM
  - SIGWINCH (for cases where the terminal size is changed)

Will update the PR shortly, but have already updated its title. I am trying to add a test to verify that keyboard interrupts have the same behavior as sending a signal to the top process first to make sure behavior always remain consistent.

stelcheck:
I gave up on testing specifically for keyboard interrupts. If you know of a way to test for that and would like me to do so, please let me know how I could set up the test.

stelcheck:
This passes just fine on my desktop (Linux), so I am not sure as to why it would fail on CI. Apologies for the spam.

stelcheck:
https://github.com/travis-ci/travis-ci/issues/704#issuecomment-328278149

stelcheck:
All tests are now passing. Please let me know if further changes are warranted before this can be merged.

src/bin.ts
```diff
@@ -54,16 +54,43 @@ v8flags(function (err, v8flags) {
...
+  forward('SIGTERM')
+
+  // Terminal size change must be forwarded to the subprocess
+  forward('SIGWINCH')
+
+  // On exit, exit this process with the same exit code
+  proc.on('close', (code: number, signal: string) => {
+    if (signal) {
+      process.kill(process.pid, signal)
+    } else if (code) {
```
blakeembrey: Do we need to check the `else if` here? If it closes, shouldn't it have one or the other? In a hypothetical case it has neither, wouldn't we still want to exit?

src/bin.ts
```diff
@@ -54,16 +54,43 @@ v8flags(function (err, v8flags) {
...
-    process.on('exit', function () {
-      if (signal) {
-        process.kill(process.pid, signal)
-      } else {
-        process.exit(code)
-      }
-    })
+  // Ignore signals, and instead forward them to the
+  // child process
+  const forward = (signal: string) => process.on(signal, () => proc.kill(signal))
```
blakeembrey: Probably easier just to make this an array loop. The comments are super verbose and mostly unnecessary too. E.g.

```js
['SIGINT', 'SIGTERM', 'SIGWINCH'].forEach(signal => process.on(signal, () => proc.kill(signal))
```

blakeembrey:
Are you absolutely sure we need to be running in detached mode? Even NPM doesn't run in detached mode and based on https://github.com/npm/lifecycle/blob/latest/index.js they just forward the two signals specifically.

src/bin.ts
stelcheck: TBH I never remember what `SIGWINCH` does, hence why I wrote it in a slightly more verbose way. If you consider this a blocker, I will change it, of course.

src/bin.ts
stelcheck: I *think* we want the process to exit naturally (no process.exit) upon a signal, but to exit with a code when there is one.

stelcheck:
https://github.com/TypeStrong/ts-node/pull/419#issuecomment-328273482 Gives the explanation. We forward terminal size change as well since programs may be long-running and meant to be executed within a terminal (e.g a REPL interface, or something like `htop`)

narthur157:
Would like to see this PR go in, currently this kills my workflow because I have a zombie proc every time I Ctrl+C out of ts-node, leaving the port binded

stelcheck:
@narthur157 I take it you tested this PR against your use case? If not, could you give it a spin to confirm it would solve your problem properly?

endel:
I've just sent a simpler alternative to the same problem here: https://github.com/TypeStrong/ts-node/pull/458

src/bin.ts
endel: On the latest version of NodeJS type definitions, `signal` should be of type `NodeJS.Signals` instead of `string`. (it's not compiling)

src/bin.ts
```diff
@@ -54,16 +54,43 @@ v8flags(function (err, v8flags) {
...
+    if (signal) {
+      process.kill(process.pid, signal)
+    } else if (code) {
+      process.exit(code)
+    }
   })
+
+  // If this process is exited with any other signals,
+  // kill the subprocess with the same signal
+  process.on('exit', (_code: number, signal: string) => proc.kill(signal))
```
endel: On the latest version of NodeJS type definitions, the [`"exit"` callback only accepts one argument](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L420). (it's not compiling)

stelcheck:
@endel fixed the two issues you reported.

What do we need to do to get this in? I'd like to do everything I can ASAP to get this merged (and hopefully released in a timely fashion), so if there is anything else I would need to do or any other concerns I would need to research or address, please let me know.

stelcheck:
Apologies for insisting, but is there anything that needs to be done to get this merged?

stelcheck:
Thank you! Let me know if I can ever be of any further help.

endel:
@stelcheck @blakeembrey It seems the problem still persists when using child processes (cluster module). Here's an issue and example project demonstrating the problem: https://github.com/TypeStrong/ts-node/issues/519

PR #423, opened 2017-09-11
Include .jsx files when allow-js is true

rharriso:
Load `.jsx` files when `allow-js: true`

Previously these files were skipped. I had to modify the registration.
```js
require('ts-node/register');
require.extensions['.jsx'] = require.extensions['.tsx'];
```

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -63,6 +63,21 @@ describe('ts-node', function () {
           }
         )
       })
+
+      it('should include jsx when `allow-js` true', function (done) {
+        exec(
+          [
+            BIN_EXEC,
+            '-O "{\\\"allowJs\\\":true}"',
+            '-p "import { Foo2 } from \'./tests/allow-js/with-jsx\'; Foo2.sayHi()"'
+          ].join(' '),
+          function (err, stdout) {
+            expect(err).to.equal(null)
+            expect(stdout).to.equal('hello world\n')
+            return done()
+          }
+        )
+      })
     }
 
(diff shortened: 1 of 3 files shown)


rharriso:
It looks like the tests are broken in Node 6.

blakeembrey:
@rharriso My fault, I made some edits in the GitHub GUI but it can be difficult to use. Had to go file-by-file to make the tweaks, and left trailing whitespace when I hit enter.

rharriso:
@blakeembrey I noticed this locally as well.

Screenshot from node 8 results (travis is using 8):
screen shot 2017-09-11 at 3 30 37 pm

Screenshot from node 4 (defined in package.json):
screen shot 2017-09-11 at 3 31 25 pm

It looks like there is a difference with how `exec` handles uncaught exceptions. I'm not sure if this is something you're worried about.

rharriso:
@blakeembrey thanks for the merge

blakeembrey:
@rharriso Yeah, the error source map output with node 8 and eval mode is janky, I believe the test is currently being skipped in `master`. If you have an idea on how to fix it, let me know - I'm confounded as to how it failed in that specific situation.

blakeembrey:
By the way, it may be a week or so until I get to a release for this. Trying to find some free time to merge a few more changes into the next major release since it's a pretty big change all together.

rharriso:
@blakeembrey thanks for the estimate. I'll be good for the near term, I just have to add that extra line to the top of my `main.js` file. So no big deal

PR #427, opened 2017-09-15
feat(bin): pass through harmony flag to node

MeirionHughes:
* allow using harmony features when running ts-node. 
* support native async-iteration. 

I'm doing work with `for await` and `[Symbol.asyncIterator]` both are supported in node>=7 with `--harmony-async-iteration`. However, currently ts-node doesn't allow enabling this via args.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -32,7 +32,9 @@ v8flags(function (err, v8flags) {
     '--perf-basic-prof',
     '--preserve-symlinks',
     '--expose-gc',
-    '--expose-http2'
+    '--expose-http2',
+    '--harmony',
+    '--harmony-async-iteration'
   ])
 
   for (let i = 0; i < argv.length; i++) {


blakeembrey:
@MeirionHughes These flags are already provided by `v8flags`, since they come from V8. You should already be able to use them if your node.js version supports it.

blakeembrey:
image

MeirionHughes:
okay, I see the issue; `--harmony-async-iteration` doesn't work via `ts-node`, but does with `node`. However, `--harmony_async_iteration` does work.

blakeembrey:
That's odd, thanks for the heads up. Anything `ts-node` can do to avoid the processing overhead of adding new flags each time is great. I'm wondering if there's a better way to handle this instead of trying to detect what node.js supports - likely a blacklist of `ts-node` flags instead of a whitelist of `node` flags.

PR #429, opened 2017-09-18
Add support for custom transformers, fixes #296

### LINKED ISSUE #296: Add support ts transformers
But only after this PR will be commited 
https://github.com/Microsoft/TypeScript/pull/14779

kriszyp:
This adds support for using custom transformers, basically just a pass-through option. Should resolve #296 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,7 @@ ts-node --compiler ntypescript --project src --ignoreWarnings 2304 hello-world.t
 * **--no-cache** Skip hitting the compiled JavaScript cache (also `process.env.TS_NODE_CACHE`)
 * **--cache-directory** Configure the TypeScript cache directory (also `process.env.TS_NODE_CACHE_DIRECTORY`)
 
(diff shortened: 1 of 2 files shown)


blakeembrey:
Do you think we should allow paths to modules to require from the CLI? This is a great start 👍

kriszyp:
We could. However, the custom transformers is a little complicated. Custom transformer can be configured as `before` or `after` transforms, the module can return higher-order functions that take additional config options before returning the transformer, and the module might return the function as the module's `default` or `transform`. My inclination would be to wait and see if definitive patterns emerge in how custom transformers are written and used, and then consider how we might want a (hopefully succinct) CLI option (and with this level of configuration involved, I'd guess this would primarily be used programmatically anyway).

That being said, if you do want a CLI option now, I'd be glad to add that to this PR, and just take a shot at best guess on these (`before`, no options to the transformer, etc.). Let me know if you would like me to add that.

PR #446, opened 2017-10-19
Update mocha docs in readme.md

LoicPoullain:
Using `mocha --compilers ts:ts-node/register,tsx:ts-node/register` is now deprecated (see notes at https://github.com/mochajs/mocha/wiki/compilers-deprecation)

The new way to start tests with mocha using ts-node would be:
`mocha --require ts-node/register "test/**/*.{ts,tsx}"`

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,11 @@ You can require `ts-node` and register the loader for future requires by using `
 mocha --require ts-node/register "test/**/*.{ts,tsx}" [...args]
 ```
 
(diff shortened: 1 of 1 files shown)


blakeembrey:
Looks good. Did you want to make a note on `--watch-extensions`? That's pretty much the only reason I used that syntax over the `--require`, so it might make sense to ensure it's covered correctly.

LoicPoullain:
Sure, I just added the note. Please let me know if something else is needed.

PR #451, opened 2017-10-25
fix chalk imports

rpd10:
This should fix issue #444


blakeembrey:
Thanks. It's unfortunate the library switched the definition export styles - the import is still correct for TypeScript, but their definition is incomplete.

Issue #457, opened 2017-11-04
ts-node doesn't handle graceful shutdown properly


endel:
Hey there,

I've created a repository that reproduces this issue here: https://github.com/endel/ts-node-graceful-shutdown-issue

All versions of `ts-node` doesn't have the expected behaviour when gracefully shutting down a Node.js application.

### `ts-node` behaviour:

Process is dettached from terminal as soon as `SIGINT` is received.

```
./node_modules/.bin/ts-node index.ts
^CGraceful shutdown started...

[21:22:41] endel:~/Projects/ts-node-graceful-shutdown-issue [master #]
$ Done, let's finish!
```

### `node` behaviour:

Process is dettached from terminal only after `process.exit()` is called.

```
$ node index.js
^CGraceful shutdown started...
Done, let's finish!
[21:23:34] endel:~/Projects/ts-node-graceful-shutdown-issue [master #]
```

I believe `ts-node` ideally should behave exactly how `node` does, shouldn't it?

Thanks

blakeembrey:
Closing with https://github.com/TypeStrong/ts-node/pull/419. Let me know if it doesn't solve your issue!

PR #458, opened 2017-11-05
graceful shutdown

endel:
Fixes https://github.com/TypeStrong/ts-node/issues/457

I noticed when hitting Ctrl-C, this block of code was never called, due to the main process being terminated: https://github.com/TypeStrong/ts-node/blob/1d630f90f31ac7cc7f82b10145819e9405d65fd3/src/bin.ts#L59-L67

By creating a trap to `SIGINT`, the main process will be closed only after the child one does.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -56,6 +56,10 @@ v8flags(function (err, v8flags) {
     { stdio: 'inherit' }
   )
 
+  // trap SIGINT. prevents main process from finishing
+  // before the child one does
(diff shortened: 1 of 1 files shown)


stelcheck:
The issue is that while in terminal this will work, IIRC it won't properly propagate signals when daemonized.

blakeembrey:
Should be closed with https://github.com/TypeStrong/ts-node/pull/419, but please let me know if it doesn't!

Issue #475, opened 2017-12-08
Move @types deps to devDependencies


whitecolor:
There is a bunch of `@types` packages in `dependencies`, seems they should not be needed by the end user, just for the sake of keeping things clean.

stelcheck:
I believe this is incorrect; for instance, you may want to use this repository programmatically to create your own customized TypeScript-enabled REPL. In such case, I believe the types need to be present for your dependent code to compile, but let me know if this is incorrect.

whitecolor:
It depends on whether exported API uses those typings or not. 
```
    "@types/arrify": "^1.0.1",
    "@types/diff": "^3.2.1",
    "@types/minimist": "^1.2.0",
    "@types/mkdirp": "^0.5.0",
    "@types/node": "^8.0.27",
    "@types/source-map-support": "^0.4.0",
    "@types/v8flags": "types/npm-v8flags#de224ae1cd5fd7dbb4e7158a6cc7a29e5315930d",
    "@types/yn": "types/npm-yn#ca75f6c82940fae6a06fb41d2d37a6aa9b4ea8e9",
```
I see that https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L29 exported interfaces don't use them.

blakeembrey:
It would be nice to do, for the sake of keeping things clean, but we can't really guarantee or easily test that these interfaces won't be accidentally exported at some time in the future. For that reason, I'd prefer to keep the `@types` in the same location as the plain dependency.

MaxCCC:
@blakeembrey we are facing proxy issues since we are using 4.0.1 because of 
```
    "@types/v8flags": "types/npm-v8flags#de224ae1cd5fd7dbb4e7158a6cc7a29e5315930d",
    "@types/yn": "types/npm-yn#ca75f6c82940fae6a06fb41d2d37a6aa9b4ea8e9",
```

We did'n have this before.

whitecolor:
> For that reason, I'd prefer to keep the @types in the same location as the plain dependency.

Well, my point is  this need to be checked and the author should understand what they exposing or what not, ts-node API surface is not so huge to be really difficult to reason about for sure.

stelcheck:
@whitecolor A human error remains quite possible. Do you know of a TypeScript tool that allows verifying that the exposed types need to be shared or not?

whitecolor:
> Do you know of a TypeScript tool that allows verifying that the exposed types need to be shared or not?

[...]

PR #484, opened 2017-12-14
do not depend on @types, which are not runtime libraries

gfx:
`@types/*` is not required for ts-node users. It just adds installation time.


blakeembrey:
If you can add tests to ensure this is true, I'd be happy to merge it.

blakeembrey:
See https://github.com/TypeStrong/ts-node/issues/475 also.

blakeembrey:
Will merge and make a quick release now, but it'd be good if you could submit tests so this doesn't break anyone.

blakeembrey:
Also, although they are not "runtime" libraries, that doesn't mean they aren't used by people depending on this library. Any types can be transitively exposed by TypeScript, so they would usually be required as a dependency and not a dev dependency. The only reason not to do this, that I can think of, is dependency size. If these deps were on NPM and not GitHub (a separate issue, I can't contribute these definitions easily so never have) I'm not sure it'd be a problem. And if it were, this is really an issue with TypeScript - it shouldn't make dependency management harder than using node.js.

gfx:
Added tests in 3f66c44

gfx:
Wow! PR is merged while I was writing tests 😆

Will make a PR for tests

PR #489, opened 2017-12-15
Disable detaching sub-process on Windows

### LINKED ISSUE #480: Does not work after upgrading to 4.0.0
`ts-node` just suddenly stop working after upgrading to 4.0.0 (The same with 4.0.1):

nodemon.json
```json
{
	"watch": ["src"],
	"ext": "ts",
	"ignore": ["src/**/*.spec.ts"],
	"exec": "ts-node -P ./tsconfig.json --no-cache ./src/index.ts"
}
```

tsconfig.json
```json
{
	"compilerOptions": {
		"target": "esnext",
		"module": "commonjs",
		"sourceMap": true,
		"lib": ["esnext"],
		"outDir": "build",
		"rootDir": "src",
		"importHelpers": true,
		"noEmitOnError": true,
		"strict": true,
		"noUnusedLocals": true,
		"noUnusedParameters": true,
		"noImplicitReturns": true,
		"noFallthroughCasesInSwitch": true,
		"moduleResolution": "node",
		"experimentalDecorators": true,
		"emitDecoratorMetadata": true
	}
}
```

nodemon: 1.12.5 (dev dependency)
typescirpt: 2.6.2 (dev dependency)
ts-node: 4.0.1 (dev dependency)

Reverted ts-node back to 3.30 solved the problem.

stelcheck:
Fixes #480

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -61,7 +61,12 @@ v8flags(function (err, v8flags) {
       // are propagated to the process tree, but `kill` is not.
       //
       // See: https://nodejs.org/api/child_process.html#child_process_options_detached
-      detached: true,
+      //
+      // This fix is not being required on Windows; besides, detached mode
+      // runs differently on Windows than on other platforms, and it would break
+      // the behavior of this application. See https://github.com/TypeStrong/ts-node/issues/480
+      // for more details.
+      detached: process.platform !== 'win32',
       stdio: 'inherit'
     }
   )


stelcheck:
@blakeembrey this should solve the reported issue.

dickon:
I found ts-node with no arguments was exiting without giving me a repl on Windows and traced it down to the detached spawn not working as expected with stdio inherited. The source branch of this PR had the fix which came to mind (i.e. backs out for Windows to the behaviour before #419 came along) so looks good to me.

ocboogie:
This fixed the issue for me 👍

PR #490, opened 2017-12-15
[test] Not to include types in dependencies

gfx:
Tests for https://github.com/TypeStrong/ts-node/pull/484

Thanks for quick response!


gfx:
Oops. It includes #44 itself. Will retry it.

PR #491, opened 2017-12-15
[test] package.json's dependencies not to include @types

gfx:
Tests for https://github.com/TypeStrong/ts-node/pull/484

Thanks for quick response.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -15,8 +15,16 @@ const SOURCE_MAP_REGEXP = /\/\/# sourceMappingURL=data:application\/json;charset
 describe('ts-node', function () {
   this.timeout(10000)
 
-  it('should export the correct version', function () {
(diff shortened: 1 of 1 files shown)


blakeembrey:
Sorry, I mustn't have communicated it well enough. I wanted the tests to ensure that any `.d.ts` file does not export an `@types` interface accidentally. This PR isn't necessary, what's really needed is something that'll go over all the dependencies and `@types` to find the types referenced in generated `.d.ts` files to ensure that any imports aren't used accidentally. Does that make more sense? For example:

```
import * as Foo from 'foo'

export const x: Foo.Bar = { ... }
```

In the above example `Foo.Bar` is a type, maybe coming from `@types/foo`, and it has been exported as part of the package. For a TypeScript user to use this file, `@types/foo` **must** be a dependency.

gfx:
I see.

PR #502, opened 2017-12-25
Enable IPC

### LINKED ISSUE #501: Missing IPC chanel
Hi, I've noticed yet another problem with a `spawn` of child process. 
For now *ts-node* has the following option while creating a child node: [`stdio: 'inherit'`](https://github.com/TypeStrong/ts-node/blob/89d62aa53fc969e0fbf8c4e4e5b63d2339856b7c/src/bin.ts#L71). 
And according [node.js docs](https://nodejs.org/api/child_process.html#child_process_options_stdio) it means "no **IPC**" (Inter-process communication)

In reality this breaks a communication with yet another parent:

parent.js:
```
const cp = require('child_process');
cp.fork('child.ts', [], {
        env: process.env,
        execPath: 'node_modules/.bin/ts-node'
    });
```

child.ts:
```
if (!process.send)
    console.log('IPC is not enabled');
else
    console.log('IPC is enabled')
```

will produce **IPC is not enabled**

See more details in the [sample repository](https://github.com/arthot/ts-node-ipc-issue) with proper environment.

So I propose to replace the `stdio` option value with `['inherit', 'inherit', 'inherit', 'ipc']` in order to enable IPC. I think it's quite safe change because according the stdio doc

> the child is launched with the IPC channel unreferenced until the child registers an event handler for the process

After such modification the sample will produce **IPC is enabled**

If someone is wondering about real-life use case sample: 

1.  npm start 
2.  webpack build client source 
3.  run server api + webpack dev server
4.  wait until the server will start 
5.  open browser

In this chain *wait until the server will start* is usually implemented in IPC way like `process.send('online')` and ts-node breaks this for now

PR #502

arthot:
This will enable IPC channel for parent/child processes while using `spawn'/`fork`

Fixes #501

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1295 @@
+{
+  "name": "ts-node",
+  "version": "4.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/arrify": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/@types/arrify/-/arrify-1.0.2.tgz",
+      "integrity": "sha512-gB1Oqypj/WbMjnWyCcvQdynRyfusKtqzUpt+EN/OtgFcjikC7ZV4qzS3SLbO1Ai2B0iVSgWvwR9A49lZGfivYg==",
+      "dev": true
+    },
+    "@types/chai": {
+      "version": "4.0.10",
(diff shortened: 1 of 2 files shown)


blakeembrey:
Closes with https://github.com/TypeStrong/ts-node/pull/536.

Issue #513, opened 2018-01-17
SIGWINCH causes process to terminate on Windows


coreh:
Whenever I resize my terminal on Windows, `ts-node` terminates with `ENOSYS`.

Here's a video of it happening:

https://streamable.com/xsoep

That video was recorded while I had v4.0.0 installed, but I've updated to v4.1.0 and even with the detached terminal change, the error still takes place.

The problem seems to be caused by this line: 
https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L76

I've confirmed it happens on both the integrated VSCode terminal and the native Windows terminal.

thasmo:
Happens for me too, although it's not a `ENOSYS` but a `EINVAL`.
```sh
0 info it worked if it ends with ok
1 verbose cli [ 'C:\\Program Files\\nodejs\\node.exe',
1 verbose cli   'C:\\Users\\user\\AppData\\Roaming\\npm\\node_modules\\npm\\bin\\npm-cli.js',
1 verbose cli   'start' ]
2 info using npm@5.6.0
3 info using node@v9.5.0
4 verbose run-script [ 'prestart', 'start', 'poststart' ]
5 info lifecycle @~prestart: @
6 info lifecycle @~start: @
7 verbose lifecycle @~start: unsafe-perm in lifecycle true
[...]

PR #514, opened 2018-01-17
Filter unsupported signals on Windows (#513)

coreh:
This fixes issue #513.

From https://nodejs.org/api/process.html#process_signal_events:

> *Note:* Windows does not support sending signals, but Node.js offers some emulation with `process.kill()`, and `subprocess.kill()`. Sending signal 0 can be used to test for the existence of a process. Sending `SIGINT`, `SIGTERM`, and `SIGKILL` cause the unconditional termination of the target process.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -73,7 +73,11 @@ v8flags(function (err, v8flags) {
   )
 
   // Ignore signals, and instead forward them to the child process.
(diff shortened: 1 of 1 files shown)


blakeembrey:
Killed subprocesses altogether to save my sanity. Closed with https://github.com/TypeStrong/ts-node/pull/536.

PR #526, opened 2018-01-24
Fix bug of TS_NODE_TYPE_CHECK and TS_NODE_CACHE not taking effect

hitochan777:
This PR fixes the bug of the environment variables `TS_NODE_TYPE_CHECK` and `TS_NODE_CACHE` not taking effect.
This is because `cache` and `typeCheck` are set to `null` by default, so `options.cache === undefined` and `options.typeCheck === undefined` are never evaluated to true. 
I thought of setting the default values to `undefined`, but `minimist` sets undefined options to the default values of their types, which is `false`.
So, I ended up negating `option.cache` and `options.typeCheck`, though I don't know this is the best choice.


blakeembrey:
Should be fixed with https://github.com/TypeStrong/ts-node/pull/536.

PR #533, opened 2018-02-14
Add `--stack-trace-limit` to known flags

Codesleuth:
I use `ts-node` to run TypeScript tests and often get truncated stack traces in the output. This PR simply adds this as a recognised flag to be passed to node.

See:
* https://nodejs.org/docs/latest-v6.x/api/cli.html#cli_v8_options (LTS)

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -34,7 +34,8 @@ v8flags(function (err, v8flags) {
     '--preserve-symlinks',
     '--expose-gc',
     '--expose-http2',
-    '--trace-warnings'
+    '--trace-warnings',
+    '--stack-trace-limit'
   ])
 
   for (let i = 0; i < argv.length; i++) {


blakeembrey:
Closing with https://github.com/TypeStrong/ts-node/pull/536. I recommend using `node -r ts-node/register` instead for node.js flags.

Codesleuth:
OK, that's fine. However, with specifying `ts-node/register` as a pre-loaded module, how can I pass the options I would have passed previously to the ts-node entry point, such as `-P myproject`?

(note that I am looking specifically to avoid using environment variables)

blakeembrey:
I'm not sure there is a way unfortunately. It's a trade off and I decided to make the trade off for project maintenance reasons. Dealing with the many flag additions, changes and bugs related to spawning child processes took too much time for myself to triage. If you come up with a solution let me know! I think the solution would live in node.js somehow supporting sub-args like `node -r [ts-node --project .]`.

PR #586, opened 2018-05-07
chore: add Node.js 10

DanielRuf:


blakeembrey:
It shouldn't be necessary to run all tests on every platform, only the current stable release. I'd recommend changing this to use `NODE_VERSION=stable` every where, then adding the additional node versions with only the current release of TypeScript.

DanielRuf:
> It shouldn't be necessary to run all tests on every platform, only the current stable release. I'd recommend changing this to use NODE_VERSION=stable every where, then adding the additional node versions with only the current release of TypeScript.

Makes sense. So should I remove all the other Node.js versions and just add `stable`?

So far the test do not fail which is good.

blakeembrey:
That sounds good to me. We can keep the minimum release the same and just bump the 8.x to stable.

DanielRuf:
Next time you can also push to my branch or try the coauthor feature ;-)
Nevertheless, thanks for the collaboration and applying the proposed change.

DanielRuf:
Coauthors can be used with git commit and amend. Pushing would have been to my origin user/repo/branch endpoint (shown here on GitHub).

https://help.github.com/articles/creating-a-commit-with-multiple-authors/

You are right, I was probably busy with other things and the answer was not quite clear to me (if this was just basic feedback or a suggested change which can be done through the review system - so I could filter my open PRs by "changes requested").

You can selfrequest a review at the top right and then give feedback on PRs (ok, request changes, comment). Comments rapidly go under in my timelines and with about 900+ notifications at the moment (different to my open PR queue which I can filter much better).

blakeembrey:
Awesome, thanks. It’s tricky to follow everyone’s preferred style. I’ll admit I’m a bit lazy when it comes to doing things the long way, though I have in the past (usually by pulling down the PR, editing locally and merging without GitHub - this was pre-branch edit days). I’ll try to do it again, I forget about the new branch pushing feature except when I’m in the UI editing. 

I don’t think the coauthors is necessary if I followed the branch flow right? I’ll try it sometime, but IIRC and I worked on your branch then you’d be the author and I’d be the commiter like usual.

How do you find using the GitHub UI to track issues/PRs - I stopped a while back when it got out of control and the I never figured out how to properly find the various things I had opened around the place or needed attention.

DanielRuf:
> I don’t think the coauthors is necessary if I followed the branch flow right? I’ll try it sometime, but IIRC and I worked on your branch then you’d be the author and I’d be the commiter like usual.

At least, but specifically you would be a coauthor (squash & merge or rebase with fixup into 1 commit).

DanielRuf:
> How do you find using the GitHub UI to track issues/PRs

It works only in a small scale. I have some browser addons like GitHub Plus and Refined GitHub but often have specific bookmarks for (tag / reaction / updated / attribute) filters and tried some tools for issue / PR tracking but I have not yet found the ideal solution for me.

Issue #565, opened 2018-04-05
Cannot find module 'ts-node/register' in VS Code Debug


Jancat:
The VS Code Node debugger **launch.json**:

```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Launch Program",
      "runtimeArgs": ["-r", "ts-node/register"],
      "args": ["${workspaceFolder}/index.ts"]
    }
  ]
}
```

I had globally installed the `typescript` and `ts-node`
```shell
ts-node -v

ts-node v5.0.1
node v9.10.1
typescript v2.8.1
```

When I execute the debugger, it threw the error:
```shell
node -r ts-node/register /home/jancat/projects/Node/huaban-crawl/index.ts 
module.js:545
    throw err;
    ^
Error: Cannot find module 'ts-node/register'
    at Function.Module._resolveFilename (module.js:543:15)
    at Function.Module._load (module.js:470:25)
    at Module.require (module.js:593:17)
    at Function.Module._preloadModules (module.js:750:12)
    at preloadModules (bootstrap_node.js:532:38)
    at startup (bootstrap_node.js:191:9)
    at bootstrap_node.js:666:3
```

blakeembrey:
You need to `npm install ts-node` locally, not globally. This may need a PR to documentation if that's unclear.

samuela:
I have ts-node installed locally. Here's my package.json:
```json
{
  "name": "api",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "start": "ts-node src/index.ts",
    "test": "tslint --project tsconfig.json"
  },
  "devDependencies": {
    "ts-node": "^7.0.1",
    "tslint": "^5.11.0",
    "typescript": "^3.1.3"
  },
  "dependencies": {
    "graphql-yoga": "^1.16.7",
    "prisma-client-lib": "^1.19.0"
  }
}
```
but I'm still seeing
```
/usr/local/bin/node -r ts-node/register --inspect-brk=40133 api/src/index.ts 
Debugger listening on ws://127.0.0.1:40133/ceca7723-bfa1-4461-b97a-a17c1a1ec2fc
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
internal/modules/cjs/loader.js:583
    throw err;
    ^
Error: Cannot find module 'ts-node/register'
[...]

PR #590, opened 2018-05-17
[docs] Usage with VS Code Debugger.

rakannimer:
Hey ! 

I bumped into the problem referenced here : https://github.com/TypeStrong/ts-node/issues/565

I added this for future me and others :) 

Cheers !

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,16 @@
 ## Installation
 
 ```sh
+# Global
+
 npm install -g ts-node
 npm install -g typescript
-```
 
-Note that to use ts-node with VSCode Debugger, you will need to install ts-node as a local dev dependency in the project you're working on : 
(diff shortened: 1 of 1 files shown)


blakeembrey:
This is true for pretty much any usage of `ts-node`, it must be installed alongside the desired TypeScript version (which is usually local and not global). Maybe there’s a way to phrase this so it’s clearer and doesn’t need to describe every usage. I would imagine it is installed locally 99% of the time because it’s a dependency of the project. Also, you can do `-D` for saving to dev.

rakannimer:
@blakeembrey Yeah, that's a good point ! 

Going to do the change to npm -D but please feel free to dismiss the PR. It might be more confusing than helpful.

PR #594, opened 2018-05-22
Add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # TypeScript Node
 
 [![NPM version][npm-image]][npm-url]
+[![NPM Install Size][install-size-image]][install-size-url]
 [![NPM downloads][downloads-image]][downloads-url]
 [![Build status][travis-image]][travis-url]
 [![Test coverage][coveralls-image]][coveralls-url]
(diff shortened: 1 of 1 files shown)


blakeembrey:
This doesn't appear relevant and the website is down.

styfle:
Hi @blakeembrey 

Sorry for the inconvenience, Package Phobia is getting a lot of traffic now and went down.

The value is to show that `ts-node` is not a huge implementation of node.js but rather a 2 MB middleman.

These are metrics that are not easily attainable so the badge is nice to show new users who are interested in the project as well as devs can try to achieve a smaller size.

Package Phobia actually caught a bug in TypeScript 5 days ago where the publish size doubled https://github.com/Microsoft/TypeScript/issues/23339#issuecomment-390365339

blakeembrey:
Cool, thanks. That seems reasonable and I'd love to accept it, do you think it could be added to https://shields.io/?

styfle:
Thanks! 🎉 
I'll need to figure out how to add it to shields.io
I guess I should just make a PR so that shields.io/packagephobia is proxy to my server?

blakeembrey:
I noticed `bundlephobia` is already on there, it seems like you could use that for reference? I've never really looked into it myself.

Issue #601, opened 2018-06-01
Failure when `declarationMap` is set in tscofnig


nomcopter:
Same as #176 but for this new option introduced in TS 2.9.

blakeembrey:
@nomcopter Do you know if this breaks compilation in `ts-node` and how? I would have thought it's a backward compatible change to the compiler output, right?

nomcopter:
In projects that set declarationMap: true it fails with the same error as the linked issue. Typescript sees an invalid configuration combination.

caseyWebb:
The same issue applies for the newly introduced `emitDeclarationOnly`

zixia:
I ran into this issue today, thanks for the quick fix in the latest version.

PR #605, opened 2018-06-11
Resolves issue when `emitDeclarationOnly` is set in tsconfig.json

caseyWebb:
ts-node was failing when `emitDeclarationOnly` is set to `true` in tsconfig.json

Related:
#601
#602


blakeembrey:
Thank you 👍

Issue #470, opened 2017-12-02
Official logo for ts-node?


kumarashwin:
We use `ts-node` heavily at my workplace and we're on the cusp of launching a major production application. As a little something for myself and my coworkers, to show support and pride for this project, I was planning to make a `ts-node` sticker as a gift for us all.

**Since there isn't a ts-node logo at the moment, I would like to understand if there is perhaps interest in having one?**

### Idea 1

![ts-node-logo](https://user-images.githubusercontent.com/12243763/33518868-6e2595c4-d76a-11e7-8260-31b4e8110c93.png)

### Idea 2

![new-logo-ts-node](https://user-images.githubusercontent.com/12243763/40760176-7f92ceb8-6463-11e8-9c4b-f65907c613ae.png)

ORESoftware:
nice work :)

stelcheck:
Outside of the visual reference, might be nice to include the word "node" somewhere, so not to confuse newcomers into thinking this is a TypeScript logo.

blakeembrey:
Love the logo and would be happy to add it to the README 😄 Agree with @stelcheck that adding `node` somewhere on there is probably a good idea - it definitely looks like it could be the TS logo as is.

kumarashwin:
@stelcheck @blakeembrey Thanks for the support. Will take your comments into account and work on perhaps another similar logo that could include what you guys are suggesting.

blakeembrey:
@kumarashwin I don't suppose you have the logo in SVG or the original somewhere we can use?

kumarashwin:
@blakeembrey I completely forgot about this! I see that I'd promised to come up with another design based on you guys' suggestion.

I'll look for the original and also see if I can do up the idea I had in mind for including `node` so you guys can see if you'd like to use it.

blakeembrey:
Thank you 😄 If you don't have the time, feel free just to share the original and I can take a look.

kumarashwin:
Sorry for the delay! Finally had some time today to look into this again. I didn't have access to the computer on which the .svgs for the image above are stored but I should be able to look at it tomorrow and retrieve them.

In the mean-time, here's another idea I had based on your suggestions:

![new-logo-ts-node](https://user-images.githubusercontent.com/12243763/40760176-7f92ceb8-6463-11e8-9c4b-f65907c613ae.png)

blakeembrey:
[...]

PR #607, opened 2018-06-13
Added logo.svg

kumarashwin:
**Relevant Issue:** https://github.com/TypeStrong/ts-node/issues/470

**UPDATE:** Seems like it's not working as I expected across multiple browsers. I'm going to go back home today and have a look at making sure the text is displayed properly. Unless someone else feels like they can fix it with the `.svg` provided. :+1:

### DIFF
--- a/logo.svg
+++ b/logo.svg
@@ -1,6 +1,4 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!-- Created with Inkscape (http://www.inkscape.org/) -->
-
 <svg
    xmlns:dc="http://purl.org/dc/elements/1.1/"
    xmlns:cc="http://creativecommons.org/ns#"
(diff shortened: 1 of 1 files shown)


blakeembrey:
@kumarashwin Would the font issue be caused by using `'Neutra Text'`? I could have a look, maybe we can turn the text into objects instead or I'll just re-export it as `png` if that doesn't work.

kumarashwin:
@blakeembrey Your clue about turning the text to path had me do some research and I think I've managed to fix it in the latest update. Let me know if it shows up fine on your side.

blakeembrey:
Looks great on my end now 👍

blakeembrey:
Would you like me to go through and delete some of the Inkscape specific comments for you? Seems odd it would export with things such as an attribute of where you exported it to on your desktop (maybe there's a "web export mode")?

kumarashwin:
@blakeembrey I believe that should take care of it. : )

kumarashwin:
@blakeembrey I tried that first but it didn't look the best (IMO) since it caused dissonance with the main hexagon having the gradients and then the other two smaller ones in `node` having a solid color.

Issue #656, opened 2018-08-06
`ts-node` fails on composite projects


sgoll:
Running `ts-node` fails when the `tsconfig.json` file contains the new attribute `"composite": true` (Microsoft/TypeScript#25600).

The error message is as follows:
```
TSError: ⨯ Unable to compile TypeScript:
error TS6304: Composite projects may not disable declaration emit.
```

The reason for this is that `ts-node` explicitly removes `declaration` from the TypeScript configuration but leaves `composite` intact, hence TypeScript complaining about this invalid combination.

The fix would be to adjust the `fixConfig` function to also always remove `composite` from the TypeScript configuration.

PR #657, opened 2018-08-06
Fix `ts-node` running on composite projects

sgoll:
See #656 for details.


dotbear:
Any chance of this getting released anytime soon? :)

sgoll:
I'd also like a quick patch release for this since it's somewhat disruptive to my workflow and patching `ts-node` inline after every `npm/yarn install` is only fun for so long before it becomes tedious. :)

Issue #655, opened 2018-08-05
[idea] tsconfig files based on file location.


trusktr:
Before adding the `ignore` option, I saw that the concern was that it could fragment the ecosystem, and that there could be problem with different libs design to be compiled with different tsconfig settings, thus not a good idea.

However, compiling node modules works great internally for private modules, where someone or a team can guarantee that everyone is using the config. In this case things are just really really easy: just publish TypeScript files and done! With good caching, a `git pull` will only cause a few files to be re-built assuming caching works as expected. Easy as ice cream and pie!

So, here's a super duper awesome idea, although it might take some effort: it might be nice if `ts-node` could switch contexts inside of each node module, and if the module has it's own `tsconfig.json` file, the use that file to compile the files of that node module.

If done properly, this would solve both the ecosystem problem and clashing config problem.

I know, this is probably lots of work, but the result would be _super awesome_. Is it feasible?

(related issues for context: https://github.com/TypeStrong/ts-node/issues/158, https://github.com/TypeStrong/ts-node/issues/155)

blakeembrey:
It might be possible, probably via upfront configuration instead of a default behaviour. Would that help you? One major hurdle we’ve run into is that TypeScript will refuse to compile inside node modules, so we will need to solve this too (I’m sure it’s just a configuration issue).

Another very challenging part is how this would all work together. For instance, some modules rely on also using the config file to do things such as override path resolution. This behaviour would probably be impossible to replicate all the way down at every level and may result in some inconsistencies when two different modules say differences on how paths or types are resolved (even if the types work, runtime may break since it’s acting differently at different levels). For these downsides I’d like to head a more thorough proposal and give myself time to think over any disadvantages of introducing this behaviour could bring.

Let’s start with this though, can you help explain why this would be preferable over having everything either compiled beforehand or transpiled by the same instance?

trusktr:
> I’m sure it’s just a configuration issue

Yep, not problem compiling in node_modules here. In my tsconfig:

```js
    "include": [
        "src/**/*",
        "node_modules/some-package-that-i-want-to-compile/src/**/*"
    ]
```

> some modules rely on also using the config file to do things such as override path resolution

In my case, I remove that from ts-config, then use the [`module-alias`](https://www.npmjs.com/package/module-alias) package. It might be possible to use it programmatically in ts-node to call it based on what is in a project's tsconfig.

> can you help explain why this would be preferable over having everything either compiled beforehand or transpiled by the same instance?

[...]

Issue #409, opened 2017-08-21
Prevent re-register


sixmen:
If I require 'ts-node/register' twice, it does not run well. For examples,

my.ts:
```
export default class My {
  msg: string;
  constructor() {
    this.msg = 'hello';
  }
}
```

index.js:
```
require('ts-node/register');
My = require('./my').default;
console.log(new My().msg);
```

If I run with `node index.js`, it works. But if I run with `ts-node index.js`, it fails.

In the real world, I run app with ts-node.
And config module, which I use, also register ts-node ( https://github.com/lorenwest/node-config/blob/master/lib/config.js#L872 ).

blakeembrey:
There's no great way to do this except for throwing an error and making it impossible for your code to execute. If I silently allow one and not the other, especially in this case, it's pretty likely the compiler options or configuration options are different and that'll result in an inconsistent runtime. I'd recommend you use a different module (there should be no reason for a config module to alter your runtime, it should do transpilation separately) or make an issue on that module.

I'll keep this open to remind me to throw an error if you try to double register in the future. This **will not** solve your problem though.

electroma:
I face similar issue with "config" module. It registers ts-node explicitly https://github.com/lorenwest/node-config/blob/02c98b7f5947ece93357652b69d9cb3da6decfd3/lib/config.js#L878

@blakeembrey, after i read your comment it is clear that fix should be done on "config" side.
Can you please suggest the best way to make sure that ts-node is already registered?

electroma:
@sixmen https://github.com/lorenwest/node-config/pull/458 should fix your issue

ZSkycat:
I face similar issue with webpack-cli.
```
ts-node --project ./tool/tsconfig.json  ./node_modules/webpack/bin/webpack.js --config ./tool/config.webpack.prod.ts
```

I modify the config file path with ts-node. But webpack-cli re-register ts-node.
When CWD equals `./`, it loads the config file `./tsconfig.json`. overwrite my config file `./tool/tsconfig`.

If there is a way to detect whether it is registered, the caller can solve this problem.

blakeembrey:
@ZSkycat Do you have a suggestion for how this could be detected? My only thought is storing a global that can be accessed somehow by all modules, but I have no consensus on where this should be stored.

FWIW, you should look at using environment variables in that case too - it's why they're provided.

ZSkycat:
@blakeembrey 
[...]

Issue #135, opened 2016-06-14
Please remove the following code (v0.9+)


wotsyula:
The `shouldIgnore()` function found here https://github.com/TypeStrong/ts-node/blob/v0.9.0/src/index.ts#L302 prevents anyone from storing any `.ts` or `.tsx` files in `node_modules` directory.

```ts
  function shouldIgnore (filename: string) {
    return relative(cwd, filename).split(sep).indexOf('node_modules') > -1
  }

  function registerExtension (ext: string) {
    const old = oldHandlers[ext] || oldHandlers['.js'] || require.extensions['.js']

    oldHandlers[ext] = require.extensions[ext]

    require.extensions[ext] = function (m: any, filename: string) {
      if (shouldIgnore(filename)) {
        return old(m, filename)
      }

      return loader(m, filename)
    }
  }

```
This will force them to stick to `v0.8` of `ts-node` or shift their support elsewhere.

I think this code is unnecessary. If I require / import a `.ts` file then I obviously want it to be compiled. Otherwise I will get a JavaScript syntax error. I see no scenario where someone would try to include a TypeScript file and not want it compiled. A better algorithm would be to let `ts-node/register` simply try to compile every `.ts` and `.tsx` file it comes across and only choose when it comes accross some other file like `.js`.

blakeembrey:
I don't think so. You should never be publishing modules as `.ts`. I would prefer to allow you to override the ignore function, but I don't think that particular practice should be promoted.

wotsyula:
1. So all I need to do is figure out a way to ensure that they are `.ts` files and you'll implement it?
2. Not all files in `node_modules` have to come from npm.org. I totally agree with your statement. But sometimes I put my own private repos there. Helps me organize my code. `npm install path/to/git` or `npm install path/to/folder`. This helps me to build websites faster not to mention anything in `node_modules` does not require `../../../` which you will see a lot in bigger libraries.
3. If I use something like web-pack, and I feed it pre-compiled `.js` files, I lose debugging info :'-( I'm sure they'll fix it but in the mean time I'd like this to work.

Tell me what you think and how I can help.

blakeembrey:
I think maybe if it supports a `--ignore` option which would override the default check. It could probably be a glob string, or a regexp pattern, whichever seems more reasonable. Then when you use that, you'll be in control.

wotsyula:
[...]

PR #668, opened 2018-08-17
feat: scope extension handlers to register path

jstewmon:
This ensures the compiler settings for a given register call are applied when the files under the registered directory are required.

ref: oclif/config/pull/44


blakeembrey:
This doesn’t seem desirable, can you describe what you’re trying to fix?

jstewmon:
I'm building a plugin-based cli with oclif. Each plugin is defined by a node package in a monorepo.

To facilitate plugin development with TS, oclif registers each plugin root (package in my case) with ts-node.

Since each package has a tsconfig.json file, we need to ensure that when a .ts file is required, it is checked/compiled using its package's tsconfig.json.

Does that make sense?

What seems undesirable to you? I couldn't think how this would be problematic, since requiring a file outside of the registered cwd wouldn't be expected to use the registered settings.

jstewmon:
I looked back over the code and thought of a case where this might not be exactly the right behavior...

The tsconfig.json file is searched for in ancestors of cwd, so I think we should capture that location and check whether the required file is a child of the config file's parent.

This way if you're using ts-node cli in a descendant folder of your project, you can still require files from an ancestral path.

Is that the type of concern you had, or is there more still?

jstewmon:
I figured out that we can use the `fileNames` found when tsconfig is loaded to decide whether to apply the registered settings or not.

blakeembrey:
I think it’s a reasonable change, but it might be an undesirable default behaviour. Some people might rely on this today in the same use case because they extend a global config file. I think it’d be a fantastic addition behind a flag, and relates to another proposal to use different configs per directory too. I can think more about how this configuration should be exposed also.

jstewmon:
Ok, makes sense to me.

How about scope: project | global

blakeembrey:
Happy with even just scope: true or scope: path, but give me some time to think it over this weekend. Maybe scope true just uses CWD by default.

blakeembrey:
The reason it doesn’t help using CWD though is that it doesn’t change over the course of program execution, so we’d need to accept a path or infer that path some how. This can be combined with another flag allowing multiple different scopes to be defined which would solve the other requested use-case too.

jstewmon:
Yeah, I think a path option that defaults to `process.cwd()` makes perfect sense - that would be useful in oclif right away, as it currently has to `chdir` into each plugin root before calling `register`.

I don't have any prior experience using ts-node, so I'm not sure what experienced users expect, but I was surprised that register didn't take a path and bind the discovered settings to the registered path (cwd).

To summarize thus far:
- Add a `path` option to register, which defaults to `process.cwd`
  - The the `path` option is mutually exclusive with `project` option
  - The `path` option is used in place of the `cwd` variable for tsconfig resolution, etc
- Add a `scope` option to register, which defaults to `false`, implying that all required files will be compiled using the same settings.
  - Should we emit a warning or throw an error if `register` is called multiple times with `scope: false`?
  - If scope is true, only apply the registered compiler settings to required files that are in the ts config `fileNames` list after loading and parsing the tsconfig.

Does that accurately capture the direction you want to go?

blakeembrey:
Sounds perfect 👍 I'd even be happy combining `scope` and `path` into a single option, with `boolean | string` input. Internally, seeing `true`, we'd use CWD. Maybe there's a preferred word than scope too, I'll leave that up to you to decide. 

> If scope is true, only apply the registered compiler settings to required files that are in the ts config fileNames list after loading and parsing the tsconfig.

If `scope` is true, I thought we'd want to compile paths only under the specified directly like your PR. Is there a time you'd only want files in `tsconfig.json`? As I write that, it sounds like yes, but beware that today `ts-node` does not use `fileNames` by default. This may be a very tricky one for people to debug also. If you think that's a desired behaviour, and at risk of sounding ridiculous right now, let's add a separate flag for that (unless, again, you feel strongly it should be default when using paths). 

> Should we emit a warning or throw an error if register is called multiple times with scope: false?

For for the error/warning, it's a nice idea. It's been a TODO forever (ref: https://github.com/TypeStrong/ts-node/issues/409) and would be breaking change. Let's push it off for now until the next version.

jstewmon:
> If scope is true, I thought we'd want to compile paths only under the specified directly like your PR. Is there a time you'd only want files in tsconfig.json?

My original issue was that calling register multiple times where each call uses a separate tsconfig.json isn't supported, so the behavior I wanted was for ts-node to use the settings found in tsconfig.json when files to which those should apply are required.

In other words, I want to register project directories with ts-node, and ts-node should do what tsc will do with the ts files in those project directories.

My first pass was to naively apply the compiler settings to only files under the cwd at the time register was called. I later realized that the cwd doesn't directly implicate there the tsconfig settings will be found, since ancestral folders are searched until tsconfig.json is found. So, I determined that the best way to determine which compiler settings should apply to a file is check if the file was in the set of `fileNames` to which typescript determined the tsconfig.json settings apply.

At this point, I see two clear use cases:
- settings are applied globally; register expected to be called once
- settings are project-specific; register scopes compiler settings to fileNames from ts parsing config

I'm not sure what the use case would be for ad-hoc registering compiler settings against a directory, ignoring the fileNames chosen by ts during parsing. Sounds like a bizarre scenario to me.

The `project` and `skipProject` options are at odds with a mode of operation that works like tsc. I've written out ways in which the options could work together, but nothing really made sense.

I suggest we keep it very simple for now and add a new option called `scope`, which can be set to a path. Setting `scope` will have the following effect:
- Settings are loaded from the `tsconfig.json` found in that path's ancestral tree.
- The registered compiler settings only apply to files in the parsed config's `fileNames` list.
- `skipProject` and `project` are ignored or an error is thrown if either are set to indicate invalid usage

>  beware that today ts-node does not use fileNames by default

I'm not exactly sure what effect the `files` option has. The docs that it makes those files load on startup. If it only affects startup behavior, then it makes sense to me that there would be orthogonal settings for specifying the startup behavior and the scope to which a register call applies. I can see how it might be confusing b/c `files` (as an option name) doesn't really express what it does.

jstewmon:
@blakeembrey I think this is ready for more feedback.

I implemented the `scope` option as described in my last comment.

I refactored a few internals to use OOP encapsulation to make it more clear that registerExtensions and shouldIgnore are bound to the settings of a registered compiler instance.

I made it possible to disable a registered compiler, to make testing a little easier (I wanted to ensure only the registered scopes were active in the scope test).

I think in a future, major release, it might be useful if the exported interface was CompilerRegistry or something like that, allowing for introspection/management of all registered compilers.

jstewmon:
Hmm... the typescript 2.0 test configuration is failing b/c I used esModuleInterop in the scope test, but that setting was added in 2.7. Any suggestions for a different setting that can be used to verify that the right settings are applied to the scope?

blakeembrey:
You can definitely add to the exports, not sure on the scope of the changes now though. I’d prefer not to use classes, they aren’t necessary here. I can tidy this up though according to your vision. Can you explain why you want to nullify project options when scope is used? I assume they are valid to be used together - say I want to put all my configs in one directory apart from my code?

jstewmon:
> say I want to put all my configs in one directory apart from my code

I sought to support that, but it's a pain b/c Typescript is strict in its convention that a tscofig.json belongs at the "root" of the "project".

The `configFileName` property passed to `parseJsonConfigFileContent` is significant b/c it serves as the baseUrl for files that are discovered. `configFileName` can be passed as undefined, in which case the `baseUrl` will be used, but a different set of default `compilerOptions` will be selected.

I finally asked myself why I was fighting against the typescript toolchain for an arbitrary requirement, for which I didn't know of any practical use case...

For lack of a better description, using `scope` means behaving like `tsc`. `tsc` requires that you have a `tsconfig.json` file at the project root, which means `scope` is inherently incompatible with `project` and `skipProject`.

blakeembrey:
This will break for people who use a different file name though, which is common. Not just a different directory. I don’t think there’s much fighting to be done, by default we don’t use anything outside compiler options and don’t care what directory things live in. The config path can still be elsewhere in any case.

blakeembrey:
Would you mind if I take another PR attempt at what we have described and share it with you to review?

jstewmon:
Sure, go for it.

> This will break for people who use a different file name though, which is common

`ts.findConfigFile` supports passing an alternate name, but I didn't see a way to support that without making the options very confusing.

Maybe if `scope` were just a `bool`, `project` could be a path to a file or directory, then those two options would work together, but the fileNames wouldn't be correct if the config is outside of the source tree.

blakeembrey:
The file names don’t matter, that’s what I’m trying to say. I’d remove that from even being parsed if I had the time to fix it, since it’s an unnecessary operation for us. As for the path, we already support a directory or file, you’re removing that support when the scope flag is set which is confusing me.

jstewmon:
> The file names don’t matter

Do you mean the filenames in the config object returned by `parseJsonConfigFileContent`? If so, then we disagree (or misunderstand each other) on this point. Those filenames are the files to which the parsed `tsconfig.json` applies, so they are the exact set of files to which the scoped compiler should apply.

jstewmon:
> But why not just have it scope to a directory?

Because that is not what `tsc` will do. All I want is a way of making ts-node behave like tsc.

I tried to provide the context before; maybe it didn't resonate. I'll try to restate it...

I have a monorepo with several packages that will be built with tsc (each package adheres to tsc constraints, having a tsconfig.json file in the package root). Some of those packages are plugins for an oclif-based cli.

When the oclif-based cli is run from the source repository, as a convenience, it dynamically registers each plugin package with ts-node, so that developers can run the cli, picking up changes made to  the ts sources without having to run tsc after editing a file.

Currently, this convenience feature is rather inconvenient for me because the settings in each of my package's tsconfig.json files are not applied by ts-node the same way as they are applied by tsc.

Not sure it makes it more clear, maybe reviewing the discussion on https://github.com/oclif/config/pull/44 would help. That's the PR I opened for the relevant oclif change that would follow this PR.

blakeembrey:
@jstewmon I understand what you want to do. I'm trying to ask why you want one approach over another. Given that this will never work like `tsc` - it takes a script to execute and spiders out from there, why do you want to so strongly tie it to a behaviour node.js can't follow? It may create more problems in the future if we ever want to skip the `fileNames` resolution which is really a wasted look up in the compiler today. Like I said, if that's really what you want to do, that's fine also, but it shouldn't rewrite too much of the package or use classes - let's keep the change simple.

jstewmon:
In my project I have tsconfig.json files in each project root that includes the `src` directory, and I have a tsconfig in each project's test directory (sibling to src). If I make two register calls, one for the project and another for its test directory, then the order in which I register them will affect the settings that are used to compile test files (if I register the project after the tests, my tests would compile using the project's settings).

That doesn't actually impact the issue I have with oclif, but I was aware of it being a case that might violate someone's expectations. Honestly, it never occurred to me that there'd be an objection to using fileNames - I didn't suggest it initially b/c I didn't know it would be available. And, it didn't occur to me that it would be a wasted step, since tsc will transpile all files. I see your point that it isn't necessary for ts-node, since it may not transpile all files, but the cost is only paid when register is called, which I wouldn't expect to be often, since currently only calling once is supported.

I can revise the PR to eliminate the class - it made the code easier for me to grok and allowed for the disabling of a registered compiler, but I can make that work with just functions and a plain object.

blakeembrey:
That’s one of the tricky parts I’d like to better understand here for sure, as I imagine it being the largest use-case. I’m happy to use the file names if we expect there to be no overlap in general use, or even just as an experiment and introduce directories later. In this case do we expect these two different configuration files to apply to distinct sets, or is there some level of overlap? If there’s some level of overlap, that’s my largest concern, because we can’t easily tell if something was already transpiled. That can result in hard to debug issues which regularly get logged today already. We also need to consider how this flag will be used to ensure we aren’t double registering in the future - does it make sense that the list is the source of truth always and any overlap with a second list is an explicit error? 

I’m only trying to optimise, selfishly, for minimum number of issues logged here. Path of least confusion. I agree that up front both feel perfectly fine and I’m happy to experiment with both, just wanted to share some possibilities. I’m very sorry it turned into a lot of discussion instead of straightforward implementation, because in either case I clearly agree it’s a valuable feature. It’s just the unknown usage that I can’t yet imagine that I want to ensure we are using the most obvious and future proof direction.

I believe this feature has extremely large overlap with https://github.com/TypeStrong/ts-node/issues/655 also, so it’s definitely part of the reason I don’t want to fragment any approach here. 

There’s one very tricky problem here also. If we support this, it doesn’t produce real files. That means no definition file is produced. If the main project then requires another which has a slightly different configuration, it won’t have any way to get the type information. This seems like a serious issue.

jstewmon:
> I’m very sorry it turned into a lot of discussion instead of straightforward implementation

No worries - I understand the need to approach this change with caution. I've just started to feel a bit of stress due to the amount of time I'm directing at this issue. I'm eager to figure out the right solution. :-)

I had assumed that the filenames would never overlap. If two register calls include the same filename, I think throwing an error is helpful b/c it's a case where one of the register calls is not going to have the expected result. We could always support an option to not throw in the future if someone has a use case for it.

#655 is interesting - it does overlap, but I think it's a different mode of operation. It's basically asking for dynamic compiler registration. When a file is required, find the compiler that applies to it; if none found, find the relevant tsconfig.json and register a new compiler. This type of feature is what I had in mind when I previously suggested making a CompilerRegistry, which could be used manage the set of registered compilers.

> If we support this, it doesn’t produce real files. That means no definition file is produced. If the main project then another project which has a different instance transpiling, it won’t have any way to get the type information. This seems like a serious issue. Do you have any thoughts on this?

I need to do some testing around this one. I want to say project references will allow for this. If it does, it might violate the expectation that a filename only appears once among all compilers.

blakeembrey:
The tricky part with having multiple instances with different files is because the errors may be different. For instance, in strict mode. We would definitely need to resolve the definition first over the TypeScript file on demand. Based on this, it might be worth combining these two requests and using the compiler instances to look up a definition before we look up the real file in the file system. It’s definitely a more complicated feature, I’m sorry I overlooked this problem from the start. For example, implicit any means we need to get the definition output and not the TypeScript file since it won’t compile with a stricter configuration. Same with any target differences, lib, types, etc. This is primarily because TypeScript will still look up these files to resolve the type definitions, not for the runtime itself.

jstewmon:
I'm very confused by your last comment. I probably just don't have enough context for it to make sense to me.

Are you just talking about the repl behavior?

blakeembrey:
@jstewmon Sorry for the long delay, work has been busy and haven't had time to type out a thoughtful response. What do you think of the new `references` feature and could it be useful for this feature? I typed out some thoughts in https://github.com/TypeStrong/ts-node/issues/135#issuecomment-421522197. Basically, we can use each `tsconfig.json/includes` to decide whether something should compile, and `references` to load up additional contexts such as this PR.

jstewmon:
No worries - I've been busy on another project as well. I should have some time next week to revisit this issue. I do think that `references` will apply to this feature, but I'm not sure how yet.

blakeembrey:
I'm revisiting old PRs and will land something similar to this request, I believe it's not too large of a feature (it's a combination of 2 features, one already landed here: https://github.com/TypeStrong/ts-node/commit/a7df13787853df42b75ba93ed1bab679a37f5438). The only other feature I see in this PR is the ability to enable/disable dynamically. One thing I see missing in this PR is `--scope` avoiding compiling files outside its scope, which I'll also add.

blakeembrey:
Dynamic enable/disable here: https://github.com/TypeStrong/ts-node/commit/85aa2d3b167e98c7deb6e49f8d1ad9eef55e4511.

blakeembrey:
Scoped support enabled here: https://github.com/TypeStrong/ts-node/commit/c1c22f8064bcc3220ab104bbed6875ca406fa81c.

PR #710, opened 2018-10-19
Add prettier screenshot

abraham:


abraham:
Fixed.

fbjork:
@abraham what terminal theme is that? Looks nice.

abraham:
@fbjork I think it was the default Seti on https://carbon.now.sh/

PR #716, opened 2018-10-25
feat: Add custom transformers for CLI

tudou527:
some node frameworks [eg: [egg](https://eggjs.org)], application run in child process, it's difficult to add transformer except cli.


tudou527:
an someone look at this pr? @blakeembrey

src/bin.ts
```diff
@@ -116,7 +120,8 @@ const service = register({
   ignoreDiagnostics: argv.ignoreDiagnostics,
   compilerOptions: parse(argv.compilerOptions) || DEFAULTS.compilerOptions,
   readFile: isEval ? readFileEval : undefined,
-  fileExists: isEval ? fileExistsEval : undefined
+  fileExists: isEval ? fileExistsEval : undefined,
+  transformers: argv.transformers ? require(join(cwd, argv.transformers)) : DEFAULTS.transformers
```
blakeembrey: This should use node module resolution instead of plain cwd join. This would enable the publishing of transformers to NPM. This should also support `argv.transformers` as an array and not just a string.

src/index.ts
```diff
@@ -71,6 +71,11 @@ export interface TypeInfo {
   comment: string
 }
 
+let envTransformer
+if (process.env['TS_NODE_TRANSFORMERS']) {
+  envTransformer = require(join(process.cwd(), process.env['TS_NODE_TRANSFORMERS'] as string))
```
blakeembrey: How would someone specify more than one here? Also, this logic is duplicated from `cli` - could we not just support strings from the options input?

README.md
```diff
@@ -129,6 +129,7 @@ _Environment variable denoted in parentheses._
 * `-P, --project [path]` Path to TypeScript JSON project file (`TS_NODE_PROJECT`)
 * `-D, --ignoreDiagnostics [code]` Ignore TypeScript warnings by diagnostic code (`TS_NODE_IGNORE_DIAGNOSTICS`)
 * `-O, --compilerOptions [opts]` JSON object to merge with compiler options (`TS_NODE_COMPILER_OPTIONS`)
+* `-F, --transformers [path]` Path of .js transformers to pass to TypeScript (`TS_NODE_TRANSFORMERS`)
```
blakeembrey: I'm not comfortable making `-F` the flag for `transformers` here - would prefer to leave it short-flag-less.

PR #720, opened 2018-10-30
ci: test on Node.js 4, 6, 8, 10 and 11

DanielRuf:


blakeembrey:
This doesn’t seem necessary. I don’t need node 4 is supported either.

DanielRuf:
Well, the build is green and there is no `engines: node: >= 6`

https://travis-ci.org/TypeStrong/ts-node/jobs/448207749

DanielRuf:
> That’s fair. Can we just test min and max then?

Makes not much sense as there are many changes in between and we test all the supported even releases up to latest / `node` in other projects too.

blakeembrey:
I understand the concern, but it’s not something encountered in this repo or any other I maintain. Therefore, I’d prefer to keep build time down. The issue is this creates a very large matrix of builds that will only keep growing over time, and the value of this time is not growing proportionally. It’s a lot easier for people to contribute when they see results faster.

blakeembrey:
Personally, I usually only do min and stable. I can also do the latest edge release to ensure forward compatibility. The chance of something breaking in a release between these and being fixed again later is extremely slim for node.js - they usually backport fixes like that to the supported major versions.

DanielRuf:
> when they see results faster.

Well then we should also check if we can use the `cache` setting to speedup builds.
I guess I'll remove Node.js 4 and 6 but keep 8 and 10, the current advertised LTS releases with the highest usage.

Also see https://github.com/nodejs/Release/blob/master/README.md

blakeembrey:
Ok, feel free to rebase/update. I'm good with 6.x being our baseline going forward in 8.0 since it'll be LTS until April.

PR #768, opened 2019-01-25
Resolves performance issue by removing invalid config

### LINKED ISSUE #754: Compilation is unbelievably slow
Hi, for some reason we have performance problems when using ts-node in our mocha tests. It takes about 500 ms to compile an empty ts file. I added some console.logs to measure the time to run [this very line](https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L316): `const output = service.getEmitOutput(fileName)` and this line takes 500 ms to run even though the content of the file is an empty string (the `code` variable on the line 314 is empty string). Actually, all our tests files take too long to process, while the production files take a few milliseconds to compile. So any `*.spec.ts` takes about 500 ms to compile, a regular `*.ts` file takes about 20 ms. Do you have any idea what could be the root cause or how should I debug it more?

We use latest mocha, ts-node and typescript, but we tried some old versions too and the problem persists. The `tsconfig.json`:

```
{
  "compileOnSave": false,
  "compilerOptions": {
    "outDir": "dist",
    "rootDir": ".",
    "declaration": true,
    "target": "es2017",
    "lib": ["es2017"],
    "module": "commonjs",
    "moduleResolution": "node"
  },
  "exclude": [
    "node_modules",
    "dist"
  ]
}
```

garethflowers:
Setting `files` and `includes` to empty arrays creates an invalid `tsconfig.json` file - and seems to subsequently cause performance issues.

This PR will remove the properties altogether.

Fixes #754.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -481,8 +481,8 @@ function readConfig (
 
   // Remove resolution of "files".
   if (!includeFiles) {
-    config.files = []
-    config.includes = []
+    delete config.files
+    delete config.includes
   }
 
   // Override default configuration options `ts-node` requires.


blakeembrey:
@garethflowers I'll take a look. My concern with deleting the files means that TypeScript actually resolves _every_ file in the directory, breaking what this was originally added to solve.

blakeembrey:
I don't think it's https://github.com/TypeStrong/ts-node/commit/e009d792c52b56fe5ba8f5cf8965dfb03916b3dd, but please try it out and let me know if the latest patch improves things!

Issue #765, opened 2019-01-23
Regression in 8.0 - TypeScript not found (when installed globally)


jamonholmgren:
(I looked at #609 and, while I believe it's related, I think my issue is more specific and replicable.)

In my CLI [Gluegun](https://github.com/infinitered/gluegun), if I have `ts-node@7.0.1` it works fine. If I upgrade to `8.0.1`, I get this error:

```
❯ gluegun
Error: Cannot find module 'typescript'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:603:15)
    at Function.resolve (internal/modules/cjs/helpers.js:28:19)
    at Object.register (/Users/jh/Code/Node/gluegun/node_modules/ts-node/src/index.ts:208:28)
    at Object.<anonymous> (/Users/jh/Code/Node/gluegun/bin/gluegun:19:22)
    at Module._compile (internal/modules/cjs/loader.js:721:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:732:10)
    at Module.load (internal/modules/cjs/loader.js:620:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:560:12)
    at Function.Module._load (internal/modules/cjs/loader.js:552:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:774:12)
```

The only change is bumping the `ts-node` version.

(Note this is only when running locally.)

*Steps to reproduce*

```
git clone https://github.com/infinitered/gluegun.git
cd ./gluegun
yarn
yarn upgrade ts-node@7.0.1
yarn link
cd ..
gluegun
# observe that it works fine
cd ./gluegun
yarn upgrade ts-node@8.0.1
cd ..
gluegun
# observe error
```

(Note that it _does_ work fine as long as you run the binary _inside_ the `gluegun` folder.)

This seems like a regression to me, unless there's something I missed in the release notes that I should have done.

blakeembrey:
cc @arcanis from https://github.com/TypeStrong/ts-node/issues/697.

blakeembrey:
@jamonholmgren This is a feature and, I believe, how it was intended to work. It sounds like you're relying on `typescript` being co-located with `ts-node`. Is there any way you can rely on `typescript` being in a different location? Can you describe how it uses `ts-node` and maybe we can find a workaround that works for all use-cases?

jamonholmgren:
@blakeembrey Here is a minimal CLI that exhibits the same behavior:

[...]

PR #793, opened 2019-03-03
feat: Add TypeScript local resolving

Maecenas:
Fix and extend #765

Changing `{ paths: [cwd] }` (BTW, is this by design?) to `{ paths: [cwd, __dirname] }` would make `ts-node` able to resolve both local and global typescript dependency.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -192,7 +192,7 @@ export function register (opts: Options = {}): Register {
   // Require the TypeScript compiler and configuration.
   const cwd = process.cwd()
   const typeCheck = options.typeCheck === true || options.transpileOnly !== true
-  const compiler = require.resolve(options.compiler || 'typescript', { paths: [cwd] })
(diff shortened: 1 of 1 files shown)


blakeembrey:
@Maecenas This seems reasonable. I was hesitant to support this since it makes resolution a little unpredictable. What are you thoughts on doing this by default vs behind a flag?

Maecenas:
@blakeembrey The motivation is to [contribute `ts-node` to Homebrew](https://github.com/Homebrew/homebrew-core/pull/37059). It seems `ts-node` is resolving `typescript` from local (`$ npm i -D`), and then global (`$ npm i -g`). As for `Homebrew`, apps are installed to a path `/usr/local/Cellar` and can not trace seperated local dependency.

The intuition behind this PR is to add installed path to dependencies resolving, as `__dirname` and after `cwd` yet before global path.

PR #801, opened 2019-03-14
expose program

devilj:
expose program from `require('ts-node').register()` as a workaround for https://github.com/Microsoft/TypeScript/issues/25147


blakeembrey:
This has been supported since https://github.com/TypeStrong/ts-node/commit/12ff53d73e5af47321f17f221396f529891cb63e. Thanks for the PR though!

PR #806, opened 2019-03-29
Stop keeping REPL state after successful commands

IvanVergiliev:
Right now, the REPL keeps the input between successful commands, leading
to some unexpected behaviours like the following:

```
~/src  ts-node
> 5
5
> +3
8
> -2
1
```

Instead of treating each line as a separate unary numeric expression,
ts-node treats the first one correctly, but then groups the 1st and
second to get "5 +3" = 8, and then the 2nd and 3rd to get "+3 -2" = 1.

This commit force-resets the input state after a successful command to
prevent this from happening.

I couldn't manage to reproduce this in tests since AFAICT it only
happens when using the REPL with a TTY, and I couldn't get this
environment to work in tests. Very open to trying again if there are
alternatives I missed.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -292,6 +292,7 @@ function replEval (code: string, _context: any, _filename: string, callback: (er
 
   try {
     result = _eval(code)
+    EVAL_INSTANCE.input = ''
   } catch (error) {
     if (error instanceof TSError) {
       // Support recoverable compilations using >= node 6.


IvanVergiliev:
Also, some more interesting side-effects when experimenting with regex matching (which is how I found this):

Screenshot 2019-03-29 at 14 17 19

That is, it's impossible to run a regex test twice in a row, because the second entry is merged with the first, and the whole thing is interpreted as a division because of one of the `/` characters.

blakeembrey:
Does this work with any TypeScript anymore? E.g. the README example. There’s a reason it keeps appending the history.

IvanVergiliev:
@blakeembrey nope, apparently not. Also I guess that means none of the REPL functionality is covered by tests since all tests passed.

So obviously this is the wrong fix, but I still claim that a bug exists. Especially the second example where you can't really run a second regex match is very confusing, and it doesn't happen in the regular `node` REPL. What would a right fix be?

blakeembrey:
@IvanVergiliev Probably handling ASI better. It's kind of difficult to do, but if it evaluates accurately and doesn't end in a semicolon, maybe we just force one to be appended?

IvanVergiliev:
@blakeembrey yeah, that sounds reasonable - thanks for the response. I don't have enough context to productively work on this at the moment, so I'll close this and let you figure out whether this is worth pursuing in a different manner.

PR #815, opened 2019-04-05
Added removing incremental & tsBuildInfoFile compiler options in fixConfig

### LINKED ISSUE #811: Should tsBuildInfoFile and incremental options are removed in fixConfig function?
It looks like `ts-node` removes some compiler options that doesn't work with `ts-node` in https://github.com/TypeStrong/ts-node/blob/8706c31430cc23862f2043b43fba7a3c3ca4b2c6/src/index.ts#L424-L444, right?

Setting `tsBuildInfoFile` compiler option (in my case for composite project) fail build with `Emit skipped` error. Should this option (quite possible that `incremental` should be removed too?) be removed along with other options in that function?

Related issue: #656

timocov:
Fixes #811

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -429,6 +429,8 @@ function fixConfig (ts: TSCommon, config: _ts.ParsedCommandLine) {
   delete config.options.declarationDir
   delete config.options.declarationMap
   delete config.options.emitDeclarationOnly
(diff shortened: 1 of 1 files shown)


timocov:
Hey @blakeembrey, can you please take a look?

PR #818, opened 2019-04-06
Add disableExit option

tenorok:
Ts-node is a development tool, and during development sometimes you need to quickly change the code and check the result. Of course, in the final form the code types should be corrected and this can be controlled by pre-commit hook like `tsc --noEmit`, for example.

Thus, it happens that in the development process, you have to place `// @ts-ignore`, before every error, which is inconvenient. Given that TSC transpile the code even if it presence types errors, I would like to have the same opportunity in the ts-node too.

This pull request shows the `--disableExit` option (and `TS_NODE_DISABLE_EXIT` environment variable), which allows you not to stop the process when typing errors occur, but only to output these errors to the console and continue execution.

It looks like this:
ts-node-disbleexit

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ _Environment variable denoted in parentheses._
 * `--pretty` Use pretty diagnostic formatter (`TS_NODE_PRETTY`, default: `false`)
 * `--skip-project` Skip project config resolution and loading (`TS_NODE_SKIP_PROJECT`, default: `false`)
 * `--skip-ignore` Skip ignore checks (`TS_NODE_SKIP_IGNORE`, default: `false`)
+* `--log-error` Logs errors of types instead of exit the process (`TS_NODE_LOG_ERROR`, default: `false`)
 
 ### Programmatic Only Options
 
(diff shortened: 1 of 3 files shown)


blakeembrey:
@tenorok Can you verify the behavior is what you expect when it's non-type errors (e.g. syntactic instead of semantic)? Also, I think it'd be better to name this feature differently - maybe `--log-error` or something similar?

tenorok:
@blakeembrey Thank you for quick answer! I did two checks, which expected led to process crashing.

1. I called un existing method `console.bar()`
Снимок экрана 2019-04-07 в 16 46 18

2. I simulated situation when iterator of cycle went over the array length limit.
```js
const arr: string[] = [];
for (let i = 0; i <= 1; i++) {
    console.log(arr[i].length);
}
```
Снимок экрана 2019-04-07 в 16 53 04

Now option renamed. I have doubted about naming, thank you for suggestion.

tenorok:
@blakeembrey Can I do anything else?

blakeembrey:
@tenorok No, sorry. This seems great. One thing I did notice is that the TypeScript compiler completely omits the invalid line instead of, in regular JS, it'd throw a syntax error. We may want to separate syntax and semantic issues later.

PR #823, opened 2019-04-24
chore(travis) add node 6/8/10/11/12

zuohuadong:


blakeembrey:
This isn’t necessary. If you are really interested, you can look at the file history and revive support for using a matrix specifically instead of running every test on every version. We really only care about minimum and latest.

PR #649, opened 2018-07-29
unsafe re-order and make .ts > .js

bluelovers:
https://github.com/TypeStrong/ts-node/issues/648

Issue #648, opened 2018-07-28
import self issue bug 


bluelovers:
## file

./strtable.ts
```ts
import * as self from './strtable';
export default self;
console.log(__filename, require.extensions);
```

## cli

```
ts-node strtable.ts
```

#### Expected behavior

should only once , even if .js exists

```
strtable.ts {
  '.ts': [Function],
  '.tsx': [Function],
  '.js': [Function],
  '.json': [Function],
  '.node': [Function] }
```

#### Actual behavior:

will load self x2 ( .js + .ts )

```
strtable.js { '.js': [Function],
  '.json': [Function],
  '.node': [Function],
  '.ts': [Function],
  '.tsx': [Function] }

strtable.ts { '.js': [Function],
  '.json': [Function],
  '.node': [Function],
  '.ts': [Function],
  '.tsx': [Function] }
```

blakeembrey:
Can you create a reproduction I can look at? There's no way, from this information you've given me, such behaviour could be replicated.

bluelovers:
![2018-07-29-10-44-59-6](https://user-images.githubusercontent.com/167966/43362432-9f919cb6-931c-11e8-8dd6-d4cd96b430c2.png)

![2018-07-29-10-45-18-7](https://user-images.githubusercontent.com/167966/43362433-a1d64f1c-931c-11e8-8304-a6be609bf541.png)

bluelovers:
when we use ts-node and import , should use .ts > .js

blakeembrey:
I disagree. Messing with core node.js expected behaviour is not something I intend to support, and having a behaviour that's different in `ts-node` vs `tsc` + node.js goes against the mission of this module.

blakeembrey:
That’s still altering the behaviour of node somewhat, I would need to override the module resolution logic. Plus it wouldn’t solve anything really for you - node always loads Javascript first, so you’re just pushing the issue to the second file that’s loaded. You can replicate this by putting a file before your “self import” and it’ll show it only loads the JavaScript and never TypeScript.

bluelovers:
how about make option?

https://github.com/TypeStrong/ts-node/pull/649

blakeembrey:
But why?

bluelovers:
make sure when we run .ts is new

sometime when .js, not update done or watcher fail for watch and make new .js

blakeembrey:
I still don’t understand why. Can’t you just output JS to a different directory? 

[...]

PR #837, opened 2019-05-28
Support `--prefer-ts-ext` flag

### LINKED ISSUE #529: require imports javascript if exist instead of typescript
When you import/require a module from the typescript file called from ts-node, it imports typescript if javascript doesn't exist, otherwise execute javascript instead of typescript.

In my case, I have a webpage in typescript and compiled to javascript to execute it in development and production, because is faster instead of execute typescript directly. I have some typescript scripts to test actions and modify some data of database. 
The problem is that this scripts use modules from webpage, that have its javascript compiled files, so when I execute scripts from ts-node this imported files are old versions and to use the new ones I need to have a tsc --watch executed, but with watcher I don't need ts-node because I can execute javascript versions of scripts.
I don't want to execute the watcher to not have another process executed with a lot of inotify watching files.

I don't know if someone of you has had this problem. I've tried to solve it in a fork, but when I create a new object with extensions instead of add the new ones to original object, I have an error when ts-node tries to execute the typescript file target, not the imports. The error is basicaly that don't use the typescript register or it doesn't compile well.

G-Rath:
Lets land this bird!

fixes #529 ,#648

superseeds #649

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -436,21 +436,21 @@ function registerExtensions (
   register: Register,
   originalJsHandler: (m: NodeModule, filename: string) => any
 ) {
-  if (!opts.preferTsExts) {
-    return extensions.forEach(ext => registerExtension(ext, ignore, register, originalJsHandler))
+  if (opts.preferTsExts) {
+    extensions.unshift(
+      '.ts',
+      '.tsx',
+      ...Object.keys(require.extensions), // tslint:disable-line
+    )
   }
 
   // @todo a better way with options
-  Array.from(new Set([
-    '.ts',
-    '.tsx',
-    ...Object.keys(require.extensions), // tslint:disable-line
-    ...extensions
-  ])).forEach(ext => {
-    registerExtension(ext, ignore, register, originalJsHandler)
+  Array.from(new Set(extensions))
+       .forEach(ext => {
+         registerExtension(ext, ignore, register, originalJsHandler)
 
-    if (ext in require.extensions) refreshRequireExtension(ext) // tslint:disable-line
-  })
+         if (ext in require.extensions) refreshRequireExtension(ext) // tslint:disable-line
+       })
 }
 
 /**
--- a/src/index.ts
+++ b/src/index.ts
@@ -426,7 +426,7 @@ function registerHandler (
   originalJsHandler: (m: NodeModule, filename: string) => any
 ) {
   if (!opts.preferTsExts) {
-    extensions.forEach(function (ext) {
+    extensions.forEach(ext => {
       registerExtension(ext, ignore, register, originalJsHandler)
     })
 
@@ -436,10 +436,8 @@ function registerHandler (
   // @todo a better way with options
   ['.ts', '.tsx']
     .concat(Object.keys(require.extensions), extensions) // tslint:disable-line
-    .filter(function (element, index, array: string[]) {
-      return array.indexOf(element) === index
-    })
-    .forEach(function (ext) {
+    .filter((element, index, array: string[]) => array.indexOf(element) === index)
+    .forEach(ext => {
       if (extensions.indexOf(ext) !== -1) {
         registerExtension(ext, ignore, register, originalJsHandler)
       }
--- a/src/index.ts
+++ b/src/index.ts
@@ -66,6 +66,7 @@ export interface Options {
   project?: string
   skipIgnore?: boolean | null
   skipProject?: boolean | null
+  preferTsExts?: boolean | null
   compilerOptions?: object
   ignoreDiagnostics?: Array<number | string>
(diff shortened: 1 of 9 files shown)


src/index.ts
```diff
@@ -415,6 +415,36 @@ function shouldIgnore (filename: string, ignore: RegExp[]) {
...
+  register: Register,
+  originalJsHandler: (m: NodeModule, filename: string) => any
+) {
+  if (!opts.preferTsExts) {
+    return extensions.forEach(ext => registerExtension(ext, ignore, register, originalJsHandler))
+  }
+
+  // @todo a better way with options
+  ['.ts', '.tsx']
+    .concat(Object.keys(require.extensions), extensions) // tslint:disable-line
```
blakeembrey: This doesn't seem necessary, `extensions` already contains the extensions we care about and in the correct order. What if we just do a loop directly on that and make this flag `--reorder-exts` (since `--reorder-ts-exts` is not true in this PR, it may contain `.js`.

src/index.ts
```diff
@@ -415,6 +415,36 @@ function shouldIgnore (filename: string, ignore: RegExp[]) {
...
+    .filter((element, index, array: string[]) => array.indexOf(element) === index)
```
blakeembrey: Let's just use `Set`, i.e. `new Set([...extensions, Object.keys(require.extensions)])`.

src/index.ts
```diff
@@ -415,6 +415,36 @@ function shouldIgnore (filename: string, ignore: RegExp[]) {
...
+    .forEach(ext => {
+      if (extensions.includes(ext)) {
```
blakeembrey: It might be nice to make `extensions` a `Set<string>` before passing it in.

src/index.ts
```diff
@@ -415,6 +415,36 @@ function shouldIgnore (filename: string, ignore: RegExp[]) {
...
+        registerExtension(ext, ignore, register, originalJsHandler)
+      }
+
+      const old = require.extensions[ext] // tslint:disable-line
+
```
blakeembrey: Remove whitespace.

src/index.ts
```diff
@@ -399,9 +401,7 @@ export function register (opts: Options = {}): Register {
   const register: Register = { cwd, compile, getTypeInfo, extensions, ts }
 
   // Register the extensions.
-  extensions.forEach(extension => {
-    registerExtension(extension, ignore, register, originalJsHandler)
-  })
+  registerHandler(opts, extensions, ignore, register, originalJsHandler)
```
blakeembrey: `registerExtensions`?

src/index.ts
G-Rath: I didn't think Set guaranteed preserving order, which is why I didn't use it.

blakeembrey: It should based on insertion order. Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

src/index.ts
blakeembrey: Sorry, or `preferExts` if that sounds reasonable. Forgot the flag name 😄

G-Rath: I think it should either be `--reorder-exts` or `--prefer-ts-exts`, since `prefer extensions` doesn't really mean anything by itself (i.e you're prefering to use extensions..?)

Thats also why I suggested `--prefer-ts-exts`, as I felt it gave you enough useful information about what it does - it reorders the order of extensions so that `.ts` extensions are "prefered" in situations where there is both a `.ts` & `.js` file with the same name.

But so if you'd prefer "reorder", then I'll make it `--reorder-exts` 😄

blakeembrey: Ah, I think I see - because of the concat order it doesn't actually re-order `.js` and `.jsx`. `--prefer-ts-exts` makes sense then. I'd maybe skip over the select, delete, assign if the property doesn't exist to start with to avoid setting `.tsx = undefined`.

G-Rath: > I'd maybe skip over the select, delete, assign if the property doesn't exist to start with to avoid setting .tsx = undefined.

So, what change should I make...?

I tried changing the part at the bottom, the:

```
      const old = require.extensions[ext] // tslint:disable-line
      delete require.extensions[ext] // tslint:disable-line
      require.extensions[ext] = old // tslint:disable-line
```

but it for some reason breaks without it. I can't figure out why :/ 
Like I compared the before & after values, and they're apparently equal and everything, but for some reason if we don't select, delete & assign it doesn't work.

src/index.ts
G-Rath: Can you confirm if you wanted this whitespace removed? I can't tell if github has pulled this comment across b/c it thinks the line hasn't changed :/

PR #849, opened 2019-07-02
Change node_modules file ignore to only affect node_modules subdirectories of cwd

Yuudaari:
This is kind of silly, but due to https://github.com/Microsoft/vscode/issues/25312, (can't get TS types from `npm link`'d module), I've moved my local module to the `node_modules` directory of the project I'm working on. However, this causes my `ts-node` `gulpfile.ts` to not work, due to the `node_modules` ignoring.

It's an easy fix, all that needs to be changed is that the files ignored with the `/node_modules/` thing should only be ignored if they're in a `/node_modules/` *subdirectory*. My solution is to just include the cwd in the beginning of the ignore regex.

I can understand if you don't want to merge this considering how silly my reason for this change is, and the fact that there are the `skipIgnore` and `ignore` options. (Altho, do I have control over those when using `ts-node/register` w/ a `gulpfile.ts`? TBH I don't even understand how that magic works 😅)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -204,12 +204,13 @@ export function register (opts: Options = {}): Register {
     ...(options.ignoreDiagnostics || [])
   ].map(Number)
 
+  const cwd = process.cwd()
+
   const ignore = options.skipIgnore ? [] : (
-    options.ignore || ['/node_modules/']
+    options.ignore || [cwd + '.*/node_modules/']
   ).map(str => new RegExp(str))
 
   // Require the TypeScript compiler and configuration.
-  const cwd = process.cwd()
   const typeCheck = options.typeCheck === true || options.transpileOnly !== true
   const compiler = require.resolve(options.compiler || 'typescript', { paths: [cwd, __dirname] })
   const ts: typeof _ts = require(compiler)


blakeembrey:
@Yuudaari Does `skipIgnore` solve the issue? E.g. have you tested this change? I'm concerned about it marginally because a node.js import does not need to come from a sub-module, and this wouldn't be changing type resolution - only which files are compiled as TypeScript. It may be valid to remove the entire thing if you and others have no issues.

Yuudaari:
Is there a way to pass ts-node args such as `skipIgnore` when using the auto-registering gulp stuff? The documentation for using ts-node with gulp is [very sparse](https://www.npmjs.com/package/ts-node#gulp). If that is a thing, then yeah this can definitely be safely closed.

I'm sure this PR will fix my specific issue, because I'm currently using it in my environment (edited the JS manually before even submitting the pull) — but I have no idea if this would break other environments. I don't understand 100% what all it does, after all, was just a quick monkey patch. 

It just felt a little odd that it doesn't work at all to use ts-node + gulp on a module that is located in a `node_modules` folder.

blakeembrey:
`TS_NODE_SKIP_IGNORE`, documented in https://www.npmjs.com/package/ts-node#cli-and-programmatic-options.

Yuudaari:
How do you recommend setting environment variables for this single module? I'm guessing that neither ts-node nor gulp has an environment-variable-loading module like `dotenv` (considering that setting it via a `.env` file doesn't work). I'm running windows, if it matters.

blakeembrey:
Closing this for now as it won't be supported. I did however change the the ignore to be based on the relative path, so that may help you. I'll look into making `--ignore` easier to use in the next release with something like mini match so you can easily negate matching the things you care about.

PR #851, opened 2019-07-09
Readme typo fix

MatissJanis:


blakeembrey:
Actually, this isn't a typo - it's an example. This flag is unnecessary if you just want to use the default `typescript` compiler.

Issue #771, opened 2019-01-30
Adding environment variable TS_NODE=true


thetutlage:
## Proposal
I would like to propose, if we can add `TS_NODE` environment variable, every time, the process is started using `ts-node`.

## Why it's needed?
Let's say I am building a script, which behaves differently based upon the runtime (Typescript or Javascript), using `TS_NODE` I can make the decision. For example: 

- internally loading file with extension `.ts` or `.js` 
- Differently handling the exports of Typescript and regular Javascript
- and so on.

blakeembrey:
What is the definition of "started using `ts-node`". Is it using the register hook, calling register in code or explicitly using `ts-node` via the CLI?

thetutlage:
I believe both. The idea is to drop an artifact, which can help scripts/modules at runtime distinguish the existence of `ts-node`

blakeembrey:
How would you feel about something added to `process` instead? E.g. `process.ts` or `process[TS_NODE_SYMBOL]`?

blakeembrey:
@thetutlage Do you have any issues with it being a symbol vs a property? I'd be a lot more comfortable starting with a symbol to test this out, since it won't affect anyone else or runtimes.

thetutlage:
Yeah, I am okay with the symbol, until it is accessible using `Symbol.for()`

thetutlage:
@blakeembrey Did you get chance to look into it? I am happy raising a PR for same :)

blakeembrey:
@thetutlage I did not, happy to accept a PR!

blakeembrey:
Isn't `Symbol.for()` global though? Shouldn't we create a `Symbol()` regularly?

blakeembrey:
Closed with https://github.com/TypeStrong/ts-node/commit/b681e2457c7acc28fcbba638a8dac6ecf27846e3.

PR #858, opened 2019-07-26
Provide TS_NODE=true when running with ts-node

### LINKED ISSUE #846: Detect when running in ts-node
Is there any canonical way for a script to detect when it is running via `ts-node`?

Sometimes I need to use slightly different paths inside my code (like to reference `src/*.ts` source, or `dist/*.js` build files, for example in TypeORM configuration) depending on whether a script is being executed with `ts-node` or not.

My workaround at the moment is to manually set an environment variable (`TS_NODE=true`) in my package scripts wherever I run `ts-node`, but it would be cleaner if this wasn't necessary.

I had a quick look through the source and couldn't see anything obvious e.g. added to global `process` object or elsewhere that would permit `ts-node` to be detected but maybe I missed something. If not, perhaps it would be worthwhile to add?

Nox-404:
Provides a way to know if a program is executed via `ts-node` or not.
This allow a program to select paths to `src` or `dist` depending on the executing environment (compiled or not)

It should fix #846

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ echo "console.log('Hello, world!')" | ts-node
 
 ![TypeScript REPL](https://github.com/TypeStrong/ts-node/raw/master/screenshot.png)
 
+> An environment variable `TS_NODE=true` is provided for conveniance when running with `ts-node` or registering it.
+
 ### Programmatic
 
 You can require `ts-node` and register the loader for future requires by using `require('ts-node').register({ /* options */ })`. You can also use file shortcuts - `node -r ts-node/register` or `node -r ts-node/register/transpile-only` - depending on your preferences.
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -32,6 +32,26 @@ describe('ts-node', function () {
       })
     })
 
+    it('should provide TS_NODE environment variable', function (done) {
+      exec(`${BIN_EXEC} tests/env`, function (err, stdout) {
+        expect(err).to.equal(null)
+        expect(stdout).to.equal('true\n')
+
+        return done()
+      })
+    })
+
+    it('should provide TS_NODE environment variable on register', function (done) {
+      exec(`node -r ../register env.ts`, {
+        cwd: TEST_DIR
+      }, function (err, stdout) {
+        expect(err).to.equal(null)
+        expect(stdout).to.equal('true\n')
+
+        return done()
+      })
+    })
+
     it('should register via cli', function (done) {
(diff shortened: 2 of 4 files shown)


blakeembrey:
I'd prefer putting a symbol on `process` if that's alright, not mutating `process.env` in any way. I'd follow the way node.js does this with something like `process[Symbol.for('ts-node.register.config')] = tsconfig`. This would also resolve a separate issue where packages would prefer to re-use the existing loaded `tsconfig.json` files and other configuration settings.

Nox-404:
Sure, want me to edit that ?

Nox-404:
@blakeembrey Adding a symbol onto process requires to change process' type, that seems like a lot of trouble to me or am I missing out something ? Also I don't fully grasp where I can get that tsconfig.
I'll be happy to edit that PR if you could give me some hints/advices :)

Nox-404:
Referencing #771

blakeembrey:
You can augment any type you want, it's not much work. The config is available after reading it: https://github.com/TypeStrong/ts-node/blob/629525474fb548fddb7a90a48ab795a9199b1997/src/index.ts#L219.

Nox-404:
@blakeembrey Sorry I've been away for a while. I don't fully grasp how I can augment process like that, could you give me some hint ?

3mard:
@blakeembrey  can you explain why would prefer using Symbol over porcess.env.TS_NODE

thetutlage:
Any plans on getting this PR merged?

dilame:
Wait for it!

blakeembrey:
> can you explain why would prefer using Symbol over porcess.env.TS_NODE

This change:

1. Mutates something that the user doesn't expect to be mutated
2. Can't provide more information than just a string
3. Would break with sub-processes since they re-use the parent `.env` by default, so you'd get `TS_NODE=true` even though there is no `ts-node` instance loaded

Nox-404:
I'll change this PR then, but I'm rather short on free time these days.
Maybe later this week.

ORESoftware:
I agree, although users can do this manually:

```
const k = cp.spawn('bash', [], {
   env: Object.assign({}, process.env, {
       TS_NODE: 'nope'
   })
})
```

I prefer yes/nope to  true/false for env vars, and in general, Linux beards recommend lowercase env vars for all but the most system-y values.

blakeembrey:
Closed with https://github.com/TypeStrong/ts-node/commit/b681e2457c7acc28fcbba638a8dac6ecf27846e3.

PR #867, opened 2019-08-16
change value to pass as an argument (for --prefer-ts-exts)

nana4gonta:
The `--prefer-ts-exts` option ( #837 ) seemed to work for CLI arguments but not for environment variables.
There are values that merge environment variables and CLI arguments ([index.ts#L197](https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L197)), but `--prefer-ts-exts` processing seems to use only the values received by CLI.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -401,7 +401,7 @@ export function register (opts: Options = {}): Register {
   const register: Register = { cwd, compile, getTypeInfo, extensions, ts }
 
   // Register the extensions.
-  registerExtensions(options, extensions, ignore, register, originalJsHandler)
+  registerExtensions(options.preferTsExts, extensions, ignore, register, originalJsHandler)
 
   return register
 }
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -401,7 +401,7 @@ export function register (opts: Options = {}): Register {
   const register: Register = { cwd, compile, getTypeInfo, extensions, ts }
 
   // Register the extensions.
-  registerExtensions(opts, extensions, ignore, register, originalJsHandler)
+  registerExtensions(options, extensions, ignore, register, originalJsHandler)
```
blakeembrey: Can we pass through only `options.preferTsExts`, I think that's the only thing used.

nana4gonta: Right, I'll fix it.

nana4gonta:
@blakeembrey If this commit ( 405dcf6 ) is fine, I would be very happy if you could merge it.

PR #879, opened 2019-09-03
Introduce programTransformers option

### LINKED ISSUE #792: [request] getCustomTransformer(program: ts.Program) just like as ts-loader's
In order to use a transformer that uses ts.Program / ts.TypeChecker, for example https://github.com/kimamula/ts-transformer-keys 

See https://github.com/TypeStrong/ts-loader#getcustomtransformers--program-program---before-transformerfactory-after-transformerfactory-- for details.

It might be somewhat difficult because TypeScript compiler restricts the use of TypeChecker in transformers: https://github.com/Microsoft/TypeScript/issues/25147 , though.

saitonakamura:
Here's the story, I have a transformer that can generate a runtime validation for json based on a generic type parameter. For that I need to have an access to a type checker, but vanilla transformers can't do that. I came across a solution in `ts-loader`: instead of passing an object literal, you pass a function that accepts a `Program` and should return that object with transformers. So I implemented the same api in `ts-node` considering the backwards compatibility (it's a brand new separate option). While implementing I came across chicken-egg problem cause `LanguageServiceHost` accepts transformers function, but you can only get an access to program once you create a `LanguageService` itself, so i'm kinda using the variable that will be initialized later. It's working, though it can be fragile with different typescript versions.

I'm also not sure about the name of the option `programTransformers`. It doens't sound like a perfect one, but i'm yet to find out the better alternative. @blakeembrey maybe you have some ideas?

Fixes #792

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,8 @@ _Environment variable denoted in parentheses._
 
 ### Programmatic Only Options
 
-* `transformers` An array of transformers to pass to TypeScript
+* `transformers` An object with transformers to pass to TypeScript
+* `programTransformers` A function that accepts a program and return and object with transformers to pass to TypeScript. Not available with `transpileOnly` flag
 * `readFile` Custom TypeScript-compatible file reading function
 * `fileExists` Custom TypeScript-compatible file existence function
 
--- a/README.md
+++ b/README.md
@@ -140,8 +140,7 @@ _Environment variable denoted in parentheses._
 
 ### Programmatic Only Options
 
-* `transformers` An object with transformers to pass to TypeScript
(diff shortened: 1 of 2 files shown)


saitonakamura:
@blakeembrey sorry for bugging, what are your thoughts on this?

blakeembrey:
@saitonakamura This looks great, but how about we just make it a union signature of `Transformers | (program) => Transformers`? That's also backward compatible.

saitonakamura:
@blakeembrey yeah, totally, I don't know why I haven't thought about it in the first place, too much C# I guess 😃 . Thank you, I've done it and it helped to simplify the code!

saitonakamura:
Thanks for merging and refactoring it!

PR #902, opened 2019-11-06
Add support for script mode (tsconfig relative to executing ts file)

n1kk:
I've had a specific use case for ts-node for running project scripts written in ts in command line so I could reuse project classes and utilities and benefit from code completion, type safety and ts refactoring capabilities. The problem was that project is configured for `esnext` so I had to create a separate config in my `./scripts/` folder and extend main config to override module target to `commonjs`. Due to this every time I wanted to use my scripts I needed to specify the correct config to use.

In this PR I've made few changes to add a flag `-s | --script-mode` that changes the path to look for `tsconfig.json` from current working directory to the executed script file directory. Also I've created new cli binding `ts-script` to be able to use it in shebang line, all it does just makes sure that `-s` flag is set.

This allows me to just call my ts scripts from command line without any extra hassle (if shebang like is set) and just pass some args to it: `./scripts/run_task.ts --dev --cleanup`. It will use the tsconfig.json that is next to the script file rather than my projects default one.


package.json
```diff
@@ -5,7 +5,8 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "bin": {
-    "ts-node": "dist/bin.js"
+    "ts-node": "dist/bin.js",
+    "ts-script": "dist/scriptMode.js"
```
blakeembrey: Use dash case (`-`) for file names.

src/bin.ts
```diff
@@ -129,6 +134,8 @@ const service = register({
   project,
   preferTsExts,
   logError,
+  scriptMode,
+  fileToExecute,
```
blakeembrey: Instead of adding a new option with conditions, why not just make `-s` behave like `cwd: dirname(fileToExecute)`?

src/bin.ts
```diff
@@ -69,7 +71,8 @@ const {
   '--skip-ignore': skipIgnore = DEFAULTS.skipIgnore,
   '--prefer-ts-exts': preferTsExts = DEFAULTS.preferTsExts,
   '--log-error': logError = DEFAULTS.logError,
-  '--build': build = DEFAULTS.build
+  '--build': build = DEFAULTS.build,
+  '--script-mode': scriptMode = DEFAULTS.scriptMode
```
blakeembrey: I’d be happy to call this `script` or even just `rel` (`relative`) since that’s all this is really doing. Script mode seems a bit expansive for the simplicity of the feature.

src/scriptMode.ts
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
+
+// add script mode flag in not present
+if (!(process.argv.includes('-s') || process.argv.includes('--script-mode'))) {
+  process.argv.splice(2, 0, '-s')
```
blakeembrey: I’m not super comfortable with this. I think we could probably export a function from `bin` instead, and use it here, to avoid mutating `process.argv`. But LGTM anyway, I’ll make any changes after we merge.

blakeembrey:
Feel free to rebase the PR on top of https://github.com/TypeStrong/ts-node/tree/8.x, it'll land faster that way. Sorry for the pain, excited to land the changes! 😄

package.json
n1kk: alright

src/bin.ts
n1kk: yeah, wasn't happy with the name wither, but just wanted to make it work
`relative` sounds better since `script` can be ambiguous in `typescript` context :)

src/bin.ts
n1kk: I see what you mean, I'll change this

src/scriptMode.ts
n1kk: Alright, I just wanted to achieve it with minimal changes, but there's certainly a more proper way to do it. The bin can be exported as function and have a code that autoruns itself if it is being called as main module:
```js
function bin(args) {
    // code
}

module.exports = bin;

if (require.main === module) {
    bin();
}
```

blakeembrey:
Closed with https://github.com/TypeStrong/ts-node/commit/1ad44bf82b7e33a4dab8fcb997c891636309e89a.

n1kk:
Hi, sorry for delay, was away on a trip.

So you already implemented all the changes in https://github.com/TypeStrong/ts-node/commit/1ad44bf82b7e33a4dab8fcb997c891636309e89a ?

Btw, about the `cwd` I remembered why I went with that route, just changing `cwd` will mean that launching script from cli and passing path to a file to it will make script look for a file relatively to itself always, which is not what you expect from standard cli behavior. So instead of `./scripts/convert.ts thisFile.json` it will have to be `./scripts/convert.ts ../thisFile.json`,  and if you rely on your shell completion it wont work. The only file I want to be resolved relatively to current script was tsconfig, that's why I've put it in `readConfig` function.

n1kk:
Alright, disregard that. Just checked on your master with some scripts, `cwd` works fine, it resolves to `cwd` of commandline. Guess I was worried over nothing.

But I discovered following: `console.log(process.argv);` from a script file with `#!/usr/bin/env ts-script` shebang line will print out (running git-bash on win10)
```bash
user@workstation MINGW64 /c/dev/ts-cli (master)
$ src/ts-cli.ts ./package.json
[
  'C:\\Users\\user\\AppData\\Roaming\\npm\\node_modules\\ts-node\\dist\\script.js',
  'C:\\dev\\ts-cli\\src\\ts-cli.ts',
  './package.json'
]
```
The first argument is ts-node's script.js. Can this be fixed? It messes with my yargs setup, since $0 is resolved to script.js and not my ts script.

n1kk:
Well, it seems like all works fine, `argv` is just a misconfiguration of my yargs.

Thanks for your work! Sorry for being useless :(

PR #915, opened 2019-11-26
fix: pass the right diagnostics list in transpileOnly mode

Weakky:
## Description

The wrong `diagnoticList` is passed to `reportTSError`, thus making all errors lacking of any relevant stack trace in `transpileOnly` mode.

I think the origin of that problem is a simple typo. It's currently referencing a variable in the outer scope with a similar name.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -213,6 +213,18 @@ describe('ts-node', function () {
       })
     })
 
+    it('should throw error even in transpileOnly mode', function (done) {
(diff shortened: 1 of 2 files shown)


blakeembrey:
Thanks for the catch! Are you interested in adding a test case?

blakeembrey:
@Weakky Not sure if you can change the parent branch to `8.x` too? If you can't I'll pick this branch across and commit it for you.

Weakky:
closing in favor of #917 to change the parent branch 👍

PR #921, opened 2019-12-02
Configure ts-node via tsconfig

### LINKED ISSUE #4: Add ability to configure TypeScript Node from tsconfig.json
Since the module could end up being used in a lot of different environments (E.g. `gulp`), there should be a way to configure it via config file or environment. Since `tsconfig.json` already exists, I'd say we piggy back off it and add a `typescript-node` property to the object.

cspotcode:
Closes #4.  Still WIP, but I still welcome early code review.


cspotcode:
@blakeembrey @G-Rath  This is ready for a review.  It still needs docs updates but I'd rather write them after the code is reviewed.

package.json
```diff
@@ -11,11 +11,17 @@
...
+    "tsconfig.schemastore-schema.json"
   ],
   "scripts": {
     "lint": "tslint \"src/**/*.ts\" --project tsconfig.json",
-    "build": "rimraf dist && tsc",
+    "lint:fix": "tslint \"src/**/*.ts\" --project tsconfig.json --fix",
+    "clean": "rimraf dist && rimraf tsconfig.schema.json && rimraf tsconfig.schemastore-schema.json",
+    "build": "npm run clean && npm run build:tsc && npm run build:configSchema",
+    "build:tsc": "tsc",
+    "build:configSchema": "typescript-json-schema --topRef --refs --validationKeywords allOf --out tsconfig.schema.json tsconfig.json TsConfigSchema && node --require ./register ./scripts/create-merged-schema",
```
blakeembrey: Nit: Consistent dash casing is currently used.

scripts/create-merged-schema.ts
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env ts-node
```
blakeembrey: I'm assuming this line doesn't work, since we use `node --require` in the script?

scripts/create-merged-schema.ts
```diff
@@ -0,0 +1,58 @@
...
+/*
+ * Create a complete JSON schema for tsconfig.json
+ * by merging the schemastore schema with our ts-node additions.
+ * This merged schema can be submitted in a pull request to
+ * SchemaStore.
+ */
+
+import axios from 'axios';
+import * as Path from 'path';
```
blakeembrey: Nit: Casing on `Path` should be `path`, but you could just import `resolve` only.

scripts/create-merged-schema.ts
```diff
@@ -0,0 +1,58 @@
...
+                  $ref: '#/definitions/compilerOptionsDefinition/properties/compilerOptions'
+                }]
+              }
+            }
+          }
+        }
+      },
+    },
+    allOf: [
+      ...schemastoreSchema.allOf.slice(0, 4),
```
blakeembrey: What does this mean? Why 4?

scripts/create-merged-schema.ts
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env ts-node
+/*
+ * Create a complete JSON schema for tsconfig.json
+ * by merging the schemastore schema with our ts-node additions.
+ * This merged schema can be submitted in a pull request to
+ * SchemaStore.
+ */
```
blakeembrey: Is this script necessary? Why can't we specify a URL within our schema to have it already be up-to-date with the remote schema? I'm worried this will mean that every TypeScript version I'll need to be generating and releasing the updated schema. You're meant to be able to use `$ref` for this use-case: https://cswr.github.io/JsonSchema/spec/definitions_references/.

src/bin.ts
```diff
@@ -79,29 +79,71 @@ export function main (argv: string[]) {
...
-    '--ignore': ignore = DEFAULTS.ignore,
-    '--transpile-only': transpileOnly = DEFAULTS.transpileOnly,
-    '--pretty': pretty = DEFAULTS.pretty,
-    '--skip-project': skipProject = DEFAULTS.skipProject,
-    '--skip-ignore': skipIgnore = DEFAULTS.skipIgnore,
-    '--prefer-ts-exts': preferTsExts = DEFAULTS.preferTsExts,
-    '--log-error': logError = DEFAULTS.logError,
-    '--emit': emit = DEFAULTS.emit
-  } = args
+  const flagOptions = {
```
blakeembrey: I'm _really_ not a fan of these changes, it's super opaque and confusing. This is a large reason I said we should probably change the ordering in a follow up PR, it's taking away from the actual feature. In general, I think we should revert all these changes and tackle it in a follow up PR. If you still want to tackle it, move `DEFAULTS` into `index.ts` _only_ and don't use them in `bin`. The current code would be better represented by just extending `options, tsconfig, DEFAULTS` where each option is previously `undefined`.

src/bin.ts
```diff
@@ -79,29 +79,71 @@ export function main (argv: string[]) {
...
+  }
+  const defaultOptions = {
+    ...DEFAULTS,
+    help: false,
+    scriptMode: false,
+    version: 0,
+    requires: [],
+    code: undefined,
+    print: false,
+    interactive: false
```
blakeembrey: None of these options should be passed to `register`, they're all CLI only.

src/index.ts
```diff
@@ -99,6 +158,81 @@ export interface RegisterOptions extends CreateOptions {
   preferTsExts?: boolean | null
 }
 
+/*
+ * This interface exists solely for generating a JSON schema for tsconfig.json.
```
blakeembrey: Can we put this in another file if it's not used here?

src/index.ts
```diff
@@ -99,6 +158,81 @@ export interface RegisterOptions extends CreateOptions {
...
+export interface TsConfigOptions
+  extends Omit<RegisterOptions,
+    | 'transformers'
+    | 'readFile'
+    | 'fileExists'
+    | 'skipProject'
+    | 'project'
+    | 'dir'
+  > {
+  requires?: Array<string>
```
blakeembrey: Why this this allowed? It's not used except via CLI. I think we should not include any option support that isn't part of `register`.

src/index.ts
```diff
@@ -99,6 +158,81 @@ export interface RegisterOptions extends CreateOptions {
...
+}
+
+/**
+ * Helper for obtaining ts-node options from multiple sources and merging them.
+ * @internal
+ */
+export class OptionsHelper {
```
blakeembrey: Remove this helper, just use `defaults()` in `register`.

src/index.ts
```diff
@@ -219,9 +355,20 @@ function cachedLookup <T> (fn: (arg: string) => T): (arg: string) => T {
  * Register TypeScript compiler instance onto node.js
  */
 export function register (opts: RegisterOptions = {}): Register {
-  const options = { ...DEFAULTS, ...opts }
+  return registerInternal(new OptionsHelper({
```
blakeembrey: Can we keep only a single `register` function - not clear why we need a layer of indirection. If we implement options clearer it should be resolved.

src/index.ts
```diff
@@ -242,25 +389,55 @@ export function register (opts: RegisterOptions = {}): Register {
  * Create TypeScript compiler instance.
  */
 export function create (options: CreateOptions = {}): Register {
+  return createInternal(new OptionsHelper({
+    explicitOptions: options
+  }))
```
blakeembrey: Ditto here, remove the extra layer of indirection.

src/index.ts
```diff
@@ -647,9 +832,29 @@ function readConfig (
...
+    TS_NODE_COMPILER_OPTIONS
+  )
+
+  const fixedConfig = fixConfig(ts, ts.parseJsonConfigFileContent(config, ts.sys, basePath, undefined, configFileName))
+
+  // Fix ts-node options that come from tsconfig.json
+  const tsconfigOptions: TsConfigOptions = {
+    ...config['ts-node']
+  }
+  if (tsconfigOptions.requires) {
```
blakeembrey: If we want this we should do it as part of a follow up PR.

src/index.ts
```diff
@@ -242,25 +389,55 @@ export function register (opts: RegisterOptions = {}): Register {
...
+    const isScoped = options.scope ? (fileName: string) => relative(cwd, fileName).charAt(0) !== '.' : () => true
+    const compiler = require.resolve(options.compiler || 'typescript', { paths: [cwd, __dirname] })
+    const ts: typeof _ts = require(compiler)
+    const readFile = options.readFile || ts.sys.readFile
+    const fileExists = options.fileExists || ts.sys.fileExists
+    return { cwd, isScoped, compiler, ts, readFile, fileExists }
+  }
+
+  // compute enough options to read the config file
+  let { cwd, isScoped, compiler, ts, fileExists, readFile } = recomputedOptions()
```
blakeembrey: I don't think we should be doing `recomputedOptions` like this, it feels wasteful. There's only two options here (`isScoped` and `compiler`) that can change from `tsconfig.json`. `isScoped` could be populated _after_ `tsconfig.json` is loaded, therefore this is really only here to load `compiler` again.

src/index.ts
```diff
@@ -242,25 +389,55 @@ export function register (opts: RegisterOptions = {}): Register {
...
+
+  // Read config file
+  const { config, options: tsconfigOptions } = readConfig(cwd, ts, fileExists, readFile, options)
+  if (tsconfigOptions) {
+    optionsHelper.tsconfigOptions = tsconfigOptions
+  }
+
+  // Merge default options, tsconfig options, and explicit --flag options
+  options = optionsHelper.merge()
```
blakeembrey: Use `defaults` directly, there's a layer of indirection and mutation here that's not necessary from what I can tell.

tsconfig.json
```diff
@@ -1,4 +1,6 @@
 {
+  "$schema": "./tsconfig.schemastore-schema.json",
+  "ts-node": {},
```
blakeembrey: Is it empty for a reason?

src/index.ts
```diff
@@ -242,25 +389,55 @@ export function register (opts: RegisterOptions = {}): Register {
...
+    const readFile = options.readFile || ts.sys.readFile
+    const fileExists = options.fileExists || ts.sys.fileExists
+    return { cwd, isScoped, compiler, ts, readFile, fileExists }
+  }
+
+  // compute enough options to read the config file
+  let { cwd, isScoped, compiler, ts, fileExists, readFile } = recomputedOptions()
+
+  // Read config file
+  const { config, options: tsconfigOptions } = readConfig(cwd, ts, fileExists, readFile, options)
```
blakeembrey: We could be explicit about `options` being passed to `readConfig` here. One feature I think is broken due to the current design is the `files` option, that functionality could be moved into this function instead of `readConfig`.

src/index.ts
```diff
@@ -99,6 +158,81 @@ export interface RegisterOptions extends CreateOptions {
...
+ * Like Object.assign or splatting, but never overwrites with `undefined`.
+ * This matches the behavior for argument and destructuring defaults.
+ * @internal
+ */
+export function defaults<T> (...sources: Array<T>): T {
+  const merged: any = {}
+  for (const source of sources) {
+    for (const key of Object.keys(source)) {
+      const value = (source as any)[key]
+      if (value !== undefined) merged[key] = value
```
blakeembrey: Probably want to do `!= null` since some options can be `null`.

scripts/create-merged-schema.ts
cspotcode: Correct, I'm generating 2 schemas.  First is `tsconfig.schema.json`, which *does* use `$ref`.  That was my initial approach, same thinking as you.

@G-Rath suggested patching ts-node's additions into the official file on Schemastore.  This script produces that merged schema.  The idea is that it would live in SchemaStore, so editors would tab-complete the ts-node fields without any configuration on the user's part.  I'm not sure if they'll ever merge it, though, because if someone tab-completes the ts-node options, they'll still need to install ts-node.  It could be nice publicity for the library; I dunno.

G-Rath: They will merge it, since it's not a breaking change :)

The configuration options are harmless if `ts-node` isn't there.

scripts/create-merged-schema.ts
cspotcode: Just so it gets spliced into the array at a point that looks nice.  https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/tsconfig.json#L630-L631

scripts/create-merged-schema.ts
cspotcode: It should work if you have `ts-node` installed globally, but that's not technically true all the time, so I put `node --require` in the package script.

src/index.ts
cspotcode: Yeah, it could be confusing.  I was thinking there might be a situation where a project bootstraps the environment by convention, maybe to tweak the source-map-support or something, so they put ts-node in the shebang and expect it to `require()` a bootstrapper.  I guess I was thinking that ts-node is, itself, bootstrapping the node environment to do something special, so it makes sense to allow that bootstrapping to be extended.

I can remove it if it's a blocker.

tsconfig.json
cspotcode: No reason, I forgot I added this.  I'll remove it.

src/index.ts
cspotcode: Do we want a `null` to be able to override a non-null value specified elsewhere?  The idea was `undefined` means the option is unspecified, whereas `null` means we explicitly want to override the default with `null`.  I was mimicking the JS language behavior of defaults in arguments and destructuring.  Will that cause problems?

src/index.ts
cspotcode: I'll move `cwd` and `isScoped` outside of this function.

The other 4 -- `compiler`, `ts`, `readFile`, and `fileExists` -- are all needed both before and after `readConfig()`.  Should I rename the function to something like `loadCompiler`?  I can also get rid of the destructuring if you want; just assign straight to local `let` variables.

cspotcode:
Thanks for the speedy review.  I started working on the necessary changes before realizing I won't be able to get to this till at least next week.  I pushed what I have, and I'll do the rest as soon as I can.

src/bin.ts
cspotcode: I'll roll them back.  If I understand you correctly, the implementation you're proposing is to never pass defaults from bin.ts into `register()` nor from `register()` into `create()`?  Instead, read them only when actually used?

blakeembrey: Right, I believe we can read them only in `create()` if we use the `defaults()` function you created. This is because the values are all from `process.env` and should be undefined (or `null` in the case of `yn` being used).

src/index.ts
blakeembrey: Yeah, for `readFile` and `fileExists`, we should probably just fallback on `ts.sys` within the `readConfig` function directly - I don't think those two functions are actually used consistently when I looked at the code, so it wouldn't be a loss to lose the functionality altogether.

blakeembrey:
@cspotcode I can also take a stab at landing the PR this weekend too. Since you've done the core functionality, I'd really be fixing all my own nit-picking since I'm particular about the structure 😄

src/index.ts
blakeembrey: > Will that cause problems?

No, I think your reasoning sounds solid. I hadn't considered someone would use `null`, it's historically just an input because of using the `yn` library which returns `null` when a value is neither true or false.

cspotcode:
@blakeembrey That is totally fine with me, if you've got the free time.  Thanks!

src/bin.ts
cspotcode: I'm upgrading yn to the latest version, which returns `undefined` instead of `null`.  This is good cuz it matches JS semantics for arg and destructuring defaults.

src/index.ts
cspotcode: Referring specifically to specifying `requires` in a tsconfig file?

src/index.ts
cspotcode: Good catch.  I think we need to set empty arrays for `files` and `includes` *within* `readConfig` because otherwise TypeScript's `parseJsonConfigFileContent` will do a bunch of globbing filesystem calls to discover the full list of files.  Does that sound right?

src/index.ts
cspotcode: I see what you mean.  `ts.parseJsonConfigFileContent` accepts a `ParseConfigHost`, which has `readFile`, `fileExists`, and `readDirectory`.  We can add a `readDirectory` option to `RegisterOptions`, and then we can consistently use `readFile` and `fileExists` in a way that makes sense.

Who is using these options, and what do they use them for?

cspotcode: It might be cleaner if ts-node accepted a `*Host`-style object instead of `readFile` and `fileExists`.  That way consumers of the API could override any of the various fs-related methods on `ts.sys`.  Looks like there are a bunch, like `directoryExists`, that are being used by ts-node but not overridable via the API.

cspotcode:
@blakeembrey This is ready for another review.  I think I've fixed all the issues:

* No more `OptionsHelper`; defaults are applied within `create()` and `readConfig()` as needed.
* Fixed inconsistent usage of `readFile` and `fileExists` in `readConfig()`
* Moved `"requires"` support into a follow-up PR
* Upgraded `yn` so that unspecified env vars will be `undefined` to be compatible with `defaults()` and JS semantics
* Fixed `files` handling in `readConfig`

cspotcode:
@blakeembrey bump!  If you're still on holiday, no rush, but when you have time I'd love to get this feature released.  Thanks in advance.

cspotcode:
@blakeembrey I hope it's ok if I bump this again.  Is there anything I can do to make this easier to review, rebasing or adding tests or whatever?  I'd love to get it into a release.  Anyone else I should request a review from if you're busy?

blakeembrey:
@cspotcode Taking a look now. Tried to have a proper vacation 😄 In the meantime I'll also add you as a collaborator in GitHub.

cspotcode:
@blakeembrey Thanks!  I didn't mean to interrupt your vacation if it's still going.

blakeembrey:
@cspotcode Nope, it's over. I intended to review this now.

src/index.ts
```diff
@@ -336,37 +336,32 @@ export function register (opts: RegisterOptions = {}): Register {
...
     return { compiler, ts }
   }
 
-  // compute enough options to read the config file
-  let { compiler, ts } = loadCompiler()
+  // Compute minimum options to read the config file.
+  let { compiler, ts } = loadCompiler(inputOptions.compiler)
 
-  // Read config file
-  const { config, options: tsconfigOptions } = readConfig(cwd, ts, optionsWithoutDefaults)
```
blakeembrey: Looks like the usage of `optionsWithoutDefaults` here is what broke https://github.com/TypeStrong/ts-node/issues/938. Fixing and adding a test now.

PR #925, opened 2019-12-14
Allow specifying "require" option via tsconfig

cspotcode:
A follow-up to #921, moved into a separate PR per review comments.  The diff for this will look much better after #921 is merged.

Allows specifying `--require` options via tsconfig.

### Questions

* [x] `create()` and `register()` will parse the option from tsconfig.  Should they also do the `require()`s?  Right now that's only done by `bin`


cspotcode:
I came across another use-case for this: using ts-node with tsconfig-paths via shebang.

```
#!/usr/bin/env ts-script
// Shebang loads ts-node options from local tsconfig.json
import * as localFooPackage from 'packages/foo';
```

```
// in tsconfig.json
{
  "ts-node": {
    "requires": ["tsconfig-paths/register"]
  },
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "packages/*": "./local-packages/*" // This example might not be perfect; I haven't used "paths" in a while
    }
  }
}
```

The alternative is to import `tsconfig-paths/register` at the top of every entry-point file.  Which I'll grant isn't the worst thing in the world, but it's a little bit messy.

ArkadyDR:
Hi @cspotcode,

I've found myself wanting this behaviour in one of my projects - for a similar reason to the use case you outlined above. I'd like to use `tsconfig-paths` with some paths defined in `tsconfig.json`, but want developers to be able to run some loose scripts (that may import some shared code using paths) via `npx ts-node <script path>` without needing to bolt on the `-r tsconfig-paths/register`. As you mentioned I could put the register at the top of each of the scripts, but given the `tsconfig.json` file specifies the paths it'd be nice if any execution using that project could load the requisite module for understanding the project (and perhaps any other bootstrapper code common to the project).

Given the age, are you likely to finish this PR? If not I might attempt something similar when I get some free time!

cspotcode:
@ArkadyDR I've merged the latest master so this PR is ready for another review.

blakeembrey and I didn't agree whether or not this should be a feature, so I shelved it in order to get #921 released.  I still think it's a good feature for the reasons that you cite, but we'll need to come to agreement.

@blakeembrey, want to revisit this?

cspotcode:
CI seems to be stuck but as far as I can tell all tests are fixed.  It properly resolves `require()` targets relative to the tsconfig file.  And I renamed the option from `requires` to `require` to match the CLI flag.

blakeembrey:
> want to revisit this?

Absolutely, sounds like a good idea now that we shipped the basic feature.

src/bin.ts
```diff
@@ -199,6 +199,8 @@ export function main (argv: string[]) {
       : undefined
   })
 
+  const requires = argsRequire.length !== 0 ? argsRequire : service.options.require || []
```
blakeembrey, line 202: Should we merge these lists together? Seems odd to have the CLI entirely overwrite the options when specified.

src/index.ts
```diff
@@ -991,6 +1005,14 @@ function readConfig (
     useCaseSensitiveFileNames: ts.sys.useCaseSensitiveFileNames
   }, basePath, undefined, configFileName))
 
+  if (tsconfigOptions.require) {
```
blakeembrey, line 1008: Thoughts on doing the require for these options within `index.ts` so configuration flows in a single direction? It might make it easier instead of needing to resolve and pass it back to the CLI which requires. Also, this would ensure we cover the programmatic usage of `register()` and via `ts-node/register`.

src/bin.ts
cspotcode: Yeah, that makes sense.  I can't think of a good use-case where you'd want the `--require` flag to suppress whatever was specified in the tsconfig.

src/index.ts
cspotcode: Yeah I like that.

The CLI uses `Module._preloadModules`; do you know if that function is doing anything special that will break if we switch to a normal `require()` call?  There's nothing wrong with `_preloadModules` but I don't really understand what it does other than `require()` stuff.

blakeembrey: Yeah, I recall I tried to change it one time. Feel free to keep the CLI logic as it is and build the internal one separately for now so there's no regression. Here's the implementation in node.js: https://github.com/nodejs/node/blob/ec2ffd6b9d255e19818b6949d2f7dc7ac70faee9/lib/internal/modules/cjs/loader.js#L1400-L1417. 

Here's where I had changed it at one point, but I can't seem to find the issue. I should have linked things better in the past 🤕 https://github.com/TypeStrong/ts-node/commit/2e99c5065f07293bb40649029990498515ba9616#diff-ee45801c137c3b9f9d8ec7ece340a514R197.

blakeembrey: The main thing it's trying to ensure is that the `require()` is occurring from the place where you run the script, not where `ts-node` lives.

ArkadyDR:
> @ArkadyDR I've merged the latest master so this PR is ready for another review.

Thanks @cspotcode and @blakeembrey for jumping on this so quickly!

> Yeah, that makes sense. I can't think of a good use-case where you'd want the --require flag to suppress whatever was specified in the tsconfig.

If it helps, my expectation would be in line with where you're going - I would think that any CLI requires would be in addition to what is in the config, although you may want to de-duplicate them if there would be an overhead in processing something twice? If someone had a case where one of the configured requires was innapropriate for a script, the solution would probably be to spin off one or more extra tsconfig files (using the extension behaviour).

src/index.ts
cspotcode: Ok, I don't think there's anything wrong with calling `_preloadModules` from within `register()`, so I'll keep doing that.

Should it happen in `register()` or in `create()`.  I feel like `register()` makes more sense because that's where we "install" the environment.  But source map support is installed in `create()`.

cspotcode:
Ok, here's how the new behavior works:

`create()` will accept a `require` array in the programmatic options, merge it with the array it pulls from `tsconfig.json`, but will *not* `require()` anything.

`register()` will `_preloadModules()` the array it gets from `create()`.

When `create()` merges the arrays, tsconfig items are first, then the programmatic items / those from the CLI `--require`.  TBH I couldn't think of a good reason to prefer one ordering over the other.

Items pulled from `tsconfig.json` will be eagerly resolved at parse-time to absolute paths via `createRequire(pathToTheTsConfigFile).resolve(item)`.  It will be as if the tsconfig file had required them.

Programmatically passed items are *not* resolved and are passed verbatim to `_preloadModules`.  This matches the current CLI behavior for `--require`.  Docs point out that you can pre-resolve and pass us absolute paths for best results.

Issue #782, opened 2019-02-13
Custom typings not working with ts-node 8.0.2


Patcher56:
After updating to ts-node version 8.0.2 the custom typings (currently placed in ./types) don't work anymore.
When I am changing the version to 8.0.1 the typings are working fine.

Typescript version: 3.3.3

Current `tsconfig.json`:
```json5
{
  "compilerOptions": {
    "sourceMap": true, // allow sourcemap support
    "strictNullChecks": true, // enable strict null checks as a best practice
    "strict": true,
    "target": "esnext", // specify ECMAScript target version
    "module": "commonjs",
    "moduleResolution": "node",
    "noImplicitAny": true,
    "declaration": false,
    "typeRoots": ["./@types"],
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "outDir": "./build"
  },
  "include": ["./src/", "./types"],
  "exclude": ["node_modules", ".logs", "upload"]
}
```

Error Message:
```
uncaughtException: ⨯ Unable to compile TypeScript:
src/import/law.ts(2,28): error TS7016: Could not find a declaration file for module 'csv-string'. '/home/patrick/dev/bodner-api/node_modules/csv-string/index.js' implicitly has an 'any' type.
  Try `npm install @types/csv-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'csv-string';`

TSError: ⨯ Unable to compile TypeScript:
src/import/law.ts(2,28): error TS7016: Could not find a declaration file for module 'csv-string'. '/home/patrick/dev/bodner-api/node_modules/csv-string/index.js' implicitly has an 'any' type.
  Try `npm install @types/csv-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'csv-string';`

    at createTSError (/home/patrick/dev/bodner-api/node_modules/ts-node/src/index.ts:228:12)
    at getOutput (/home/patrick/dev/bodner-api/node_modules/ts-node/src/index.ts:334:40)
    at Object.compile (/home/patrick/dev/bodner-api/node_modules/ts-node/src/index.ts:367:11)
    at Module.m._compile (/home/patrick/dev/bodner-api/node_modules/ts-node/src/index.ts:413:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/home/patrick/dev/bodner-api/node_modules/ts-node/src/index.ts:416:12)
[...]

PR #930, opened 2019-12-26
doc: add precision regarding `--files` option

ledenis:
I suggest to add a little precision about `--files` options, since it does not seem very clear https://github.com/TypeStrong/ts-node/issues/782#issuecomment-561257824


README.md
```diff
@@ -215,7 +215,7 @@ An alternative approach for definitions of third-party libraries are [triple-sla
 import UntypedJsLib from "untyped_js_lib"
 ```
 
-**Tip:** If you _must_ use `files`, enable `--files` flags or set `TS_NODE_FILES=true`.
+**Tip:** If you _must_ use `files` or `include`/`exclude`, enable `--files` flags or set `TS_NODE_FILES=true`.
```
cspotcode: Nit: how about this phrasing, since some people combine "files" with "include"?

> If you _must_ use `files`, `include`, or `exclude`, enable ...

ledenis: Thanks, done. Also reworded the phrase above

PR #942, opened 2020-01-13
Downgrade to yn v3 to avoid v4's node 10 requirement

### LINKED ISSUE #940: Node engine V8 compatibility is broken
ts-node is no longer work on Node V8, because [yn](https://github.com/sindresorhus/yn) is enforcing Node V10 compatibility.
Caused by: [yn#26](https://github.com/sindresorhus/yn/pull/26)

The motivation for changing the V8 requirement to V10 is because Node.js V8 is no longer supported. I believe node engine requirement should indicate the engine(s) the module **works** with.

cspotcode:
Fixes #940


src/index.ts
```diff
@@ -26,6 +26,15 @@ declare global {
  */
 export const INSPECT_CUSTOM = util.inspect.custom || 'inspect'
 
+/**
+ * Wrapper around yn module that returns `undefined` instead of `null`.
+ * This is implemented by yn v4, but we're staying on v3 to avoid v4's node 10 requirement.
+ */
+function yn (value: string | undefined) {
+  const parsed = ynModule(value)
+  return parsed === null ? undefined : parsed
```
blakeembrey: We could also just use bullish coalescing within the DEFAULTS object nowadays.

Borewit:
Thank you so much @cspotcode & @blakeembrey!

PR #943, opened 2020-01-13
Fix bash error in usage example on README page

abraj:
On Linux bash terminal, executing
`echo "console.log('Hello, world!')"`
gives the following error:
`bash: !': event not found`

To fix the issue, changing the log string from
`Hello, world!` to `Hello, world`

Tested on following systems:
* Ubuntu 18.04.3 LTS
* Linux Mint 19

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ ts-node -e 'console.log("Hello, world!")'
 ts-node -p -e '"Hello, world!"'
 
 # Pipe scripts to execute with TypeScript.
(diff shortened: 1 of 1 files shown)


cspotcode:
@abraj Good catch, thanks.  Can you think of any problems with switching the quotes around to keep the exclamation mark?

```
echo 'console.log("Hello, world!")' | ts-node
```

abraj:
@cspotcode Even better! I really think that it should be the way to go.
Works perfectly without any bash errors.

PR #948, opened 2020-01-24
Deprecate ts node script

G-Rath:
Following on from #939, this deprecated `ts-script` in favor of `ts-node-script`.

(I've based this off #939 to avoid a merge conflict, so this should be merged after that PR is merged).


package.json
```diff
@@ -6,7 +6,9 @@
   "types": "dist/index.d.ts",
   "bin": {
     "ts-node": "dist/bin.js",
-    "ts-script": "dist/script.js"
+    "ts-script": "dist/script-deprecated.js",
+    "ts-node-script": "dist/script.js",
```
blakeembrey: Nit: Let's move to `bin-*.js` to clarify intent, but LGTM.

PR #958, opened 2020-02-13
Add instructions to readme for intellij / webstorm

bertyhell:
based on: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006528060-New-Webstorm-is-missing-debug-integration-with-Typescript

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,19 +94,11 @@ Create a new node.js configuration, add `-r ts-node/register` to node args and m
 
 **Note:** If you are using the `--project <tsconfig.json>` command line argument as per the [Configuration Options](#configuration-options), and want to apply this same behavior when launching in VS Code, add an "env" key into the launch configuration: `"env": { "TS_NODE_PROJECT": "<tsconfig.json>" }`.
 
-### Intellij (and Webstorm)
+### IntelliJ (and WebStorm)
 
-Run > Edit Configurations > click the plus icon at the top left > Node.js
+Create a new Node.js configuration and add `-r ts-node/register` to "Node parameters."
 
-Node interpreter: node.js (the default should be fine)
-Node parameters: `--inspect=0.0.0.0:9229 --require ts-node/register`
-Working directory: default should be fine
-Javascript file: select the main ts file you want to run. eg: `server/index.ts`
-Application parameters: Extra options for the ts-node process. eg: `--project tsconfig.json`
-Environment variables: `TS_NODE_TRANSPILE_ONLY=true`
-
-Apply and OK
-Click the green triangle at the top of the IntelliJ window to run or the green bug icon to debug.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,6 +94,20 @@ Create a new node.js configuration, add `-r ts-node/register` to node args and m
 
 **Note:** If you are using the `--project <tsconfig.json>` command line argument as per the [Configuration Options](#configuration-options), and want to apply this same behavior when launching in VS Code, add an "env" key into the launch configuration: `"env": { "TS_NODE_PROJECT": "<tsconfig.json>" }`.
 
+### Intellij (and Webstorm)
+
+Run > Edit Configurations > click the plus icon at the top left > Node.js
+
+Node interpreter: node.js (the default should be fine)
+Node parameters: `--inspect=0.0.0.0:9229 --require ts-node/register --require tsconfig-paths/register`
```
cspotcode: We don't mention tsconfig-paths/register in the VSCode example, so this is slightly inconsistent.  I would omit the extra --require and limit it to `ts-node/register`

README.md
```diff
@@ -94,6 +94,20 @@ Create a new node.js configuration, add `-r ts-node/register` to node args and m
...
+Working directory: default should be fine
+Javascript file: select the main ts file you want to run. eg: `server\index.ts`
```
cspotcode: Nit: use a forward slash in path

README.md
```diff
@@ -94,6 +94,20 @@ Create a new node.js configuration, add `-r ts-node/register` to node args and m
...
+Application parameters: Extra options for the ts_node process. eg: `--project tsconfig.json --r`
```
cspotcode: Nit: ts-node should be hyphenated.

`--r` is not a valid flag.

README.md
```diff
@@ -94,6 +94,20 @@ Create a new node.js configuration, add `-r ts-node/register` to node args and m
...
+Environment variables: `TS_NODE_TRANSPILE_ONLY=true`
+
+Apply and OK
+Click the green triangle at the top of the intellij window to run or the green bug icon to debug
```
cspotcode: Nit: Fix capitalization of brand names, add a period at the end of the sentence.

README.md
cspotcode: Is binding to 0.0.0.0 ok?  That'll allow external machines to connect to the process, right?

bertyhell: > Is binding to 0.0.0.0 ok? That'll allow external machines to connect to the process, right?

yes, i think it's because we sometimes run it inside a docker instance

cspotcode:
The configuration wasn't going to work because passing `--project tsconfig.json` doesn't work when requiring `ts-node/register`.  I simplified the configuration to match the level of detail in the VSCode example.  The only thing you need to specify is `--require ts-node/register`.  IntelliJ takes care of everything else.

Issue #266, opened 2017-01-25
mocha --watch not working with ts-node


sweetim:
I am using Typescript and node-ts for the mocha unit testing, but the mocha --watch option is not working with ts-node

Below is the command i pass to run the unit test

```
mocha -r ts-node/register test/**/**/*.test.ts --watch -R min
```

The version I am currently using

```
"mocha": "^3.2.0",
"ts-node": "^1.7.3",
"typescript": "^2.1.5"
```

It works on the first time, but when I change the file it doesnt run again?

How to fix this problem?

blakeembrey:
Is this issue a duplicate of https://github.com/TypeStrong/ts-node/issues/113?

sweetim:
@blakeembrey i could compile the code without problem

blakeembrey:
Do you have an example project you'd be able to share? The only reason I can think of this happening is if there's a compilation error blocking the re-require. I don't think mocha does anything else that's special.

sweetim:
I created a simple example to showcase this problem, below is the code:

Below, is my project directory
```
src
   > student.ts
test
   > mocha.opts
   > student.test.ts
```

```
// student.ts
export class Student {
    constructor(
        private _name: string,
    ) {    }

    get name(): string {
        return this._name;
    }
}
```

```
// test/student.test.ts
import * as assert from 'assert';
import { Student } from '../src/student';

describe('Student', () => {
    describe('#getName()', () => {
        it('should return name', () => {
            let s = new Student('ali', 12);
            assert.strictEqual(s.name, 'ali');
        })
    });
});
```
```
// package.json
"scripts": {
    "test": "mocha",
    "test:watch": "npm run test -- --watch -R min"
  },
  "devDependencies": {
    "@types/mocha": "^2.2.38",
    "@types/node": "^7.0.4",
    "mocha": "^3.2.0",
    "ts-node": "^2.0.0",
    "typescript": "^2.1.5"
  }
```

```
// tsconfig.json
{
    "compilerOptions": {
        "target": "ES6",
        "module": "commonjs",
        "noEmitOnError": true,
        "noImplicitAny": true,
        "experimentalDecorators": true,
        "sourceMap": true,
        "sourceRoot": "src",
        "outDir": "bin",
        "typeRoots": [
            "node_modules/@types"
        ]
    }
}
```

I try to run the watch mode, it will just get stuck there and it doesnt reload whenever I saved the file, or add and remove new file

[...]

PR #961, opened 2020-02-16
Add example for mocha 7, since they removed some watch flags in v7

cspotcode:
This adds a mocha example that works for mocha 7.

They removed the --watch-extensions flag in v7.
https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#700--2020-01-05

I asked in #266 if anyone's able to test it, since I don't use mocha's watch mode on any of my projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,16 @@ You can require `ts-node` and register the loader for future requires by using `
 
 ### Mocha
 
+Mocha 6
+
 ```sh
 mocha --require ts-node/register --watch-extensions ts,tsx "test/**/*.{ts,tsx}" [...args]
+```
+
+Mocha 7
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,10 +54,18 @@ You can require `ts-node` and register the loader for future requires by using `
...
 mocha --require ts-node/register --watch-extensions ts,tsx "test/**/*.{ts,tsx}" [...args]
 ```
 
+Mocha 7
+
+```sh
+mocha --require ts-node/register --extensions ts,tsx --watch --watch-files src 'tests/**/*.{ts,tsx}' [...args]
+```
+
 **Note:** `--watch-extensions` is only used in `--watch` mode.
```
blakeembrey: Move the note back below mocha 6 or delete it?

PR #963, opened 2020-02-16
Implement getProjectVersion

cspotcode:
Due to Microsoft/TypeScript#36748, TypeScript's language service always thinks the `Program` is out-of-date and rebuilds it 3 times for every `.ts` file that we load: once each for `getEmitOutput`, `getSemanticDiagnostics`, and `getSyntacticDiagnostics`.

We can workaround this bug by implementing `getProjectVersion`.  If `getProjectVersion` stays the same, TypeScript will reuse the existing `Program`.  It'll still need to be rebuilt every time we modify the rootFiles array, but that's only once instead of 3 times.

I also added debug log statements that double-check when the `Program` instance is rebuilt.  This should be helpful if we ask for TS_NODE_DEBUG logs from users.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -538,8 +538,9 @@ export function create (rawOptions: CreateOptions = {}): Register {
         updateMemoryCache(code, fileName)
 
         const programBefore = service.getProgram()
-        if (programBefore !== previousProgram)
-        debug(`compiler rebuilt Program instance when getting output for ${ fileName }`)
+        if (programBefore !== previousProgram) {
+          debug(`compiler rebuilt Program instance when getting output for ${ fileName }`)
+        }
 
         const output = service.getEmitOutput(fileName)
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -470,7 +470,7 @@ export function create (rawOptions: CreateOptions = {}): Register {
 
       // Create the compiler host for type checking.
       const serviceHost: _ts.LanguageServiceHost = {
-        getProjectVersion: () => `${ projectVersion }`,
+        getProjectVersion: () => String(projectVersion),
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -505,27 +507,52 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
+        let shouldIncrementProjectVersion = false
         const fileVersion = fileVersions.get(fileName) || 0
+        const isFileInCache = fileVersion !== 0
 
         // Add to `rootFiles` when discovered for the first time.
-        if (fileVersion === 0) rootFileNames.push(fileName)
+        if (!isFileInCache) {
+          rootFileNames.push(fileName)
+          // Modifying rootFileNames means a project change
+          shouldIncrementProjectVersion = true
```
blakeembrey: Nit: You could just do `projectVersion++` here - it's essentially the same, an extra +1 every now and then isn't a big deal.

src/index.ts
```diff
@@ -469,6 +470,7 @@ export function create (rawOptions: CreateOptions = {}): Register {
 
       // Create the compiler host for type checking.
       const serviceHost: _ts.LanguageServiceHost = {
+        getProjectVersion: () => `${ projectVersion }`,
```
blakeembrey: Nit: Personally I'd just do `String(projectVersion)` since you just want to type cast.

cspotcode:
> I'm not sure how useful the debug statements are if the conditions are as mentioned in the comment - maybe better as an `invariant` check that throws?

Yeah, I wasn't sure about that.  I really wanted to avoid ts-node crashing if new versions of the TypeScript compiler misbehaved.  I figured better for us to remain functional with degraded performance and let the debug logs point out why.

PR #970, opened 2020-03-02
Override `isExternalLibraryImport` as needed; re-add `realpath`

### LINKED ISSUE #797: 8.0.3 breaks project with: Unable to require `.d.ts` file
Hey, today I upgraded the ts-node package to the newest version 8.0.3 and it now throws an exception: 

```
TypeError: Unable to require `.d.ts` file.
This is usually the result of a faulty configuration or import. Make sure there is a `.js`, `.json` or another executable extension and loader (attached before `ts-node`) available alongside `graphql.schema.ts`.
    at getOutput (backend/server/node_modules/ts-node/src/index.ts:343:15)
    at Object.compile (backend/server/node_modules/ts-node/src/index.ts:368:11)
    at Module.m._compile (backend/server/node_modules/ts-node/src/index.ts:414:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:745:10)
    at Object.require.extensions.(anonymous function) [as .ts] (backend/server/node_modules/ts-node/src/index.ts:417:12)
    at Module.load (internal/modules/cjs/loader.js:626:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:566:12)
    at Function.Module._load (internal/modules/cjs/loader.js:558:3)
    at Module.require (internal/modules/cjs/loader.js:663:17)
    at require (internal/modules/cjs/helpers.js:20:18)
```

The version 8.0.2 works. I had no time to find the exact cause.

### LINKED ISSUE #819: imports from symlinked and paths referenced files get resolved not following the symlink's location upwards
https://github.com/MarcWeber/tsmono
if you run test/test.ts you'll end up with tmp/tool.
If you try tsc -p .  -> everything should be fine
If you try running tool/tool.ts it says ```cannot resolve "object-hash"```

```
tsmono/links/lib-name -> ../../../src/lib-name
../../../src/lib-name/lib.ts   contents: import * from "object-hash"
node_modules/object-hash (along with types)

tsconfig paths:  lib-name/*  : [ 'tsmono/links/lib-name/*' ]
```

vscode is fine, tsc is fine, but ``` node -r tsconfig-paths/register  -r 'ts-node/register/transpile-only' tool/tool.ts```  looks for object-hash by walking up from the symlinked destination, not the location.

I don't know what should be correct. The behavior is different which is the problem.
I'd prefer it walking up from the symlinked location.

What am I trying to do ?
Having to complie .ts -> .js is tiresome. You need "references", then you still need to link to node_modules using fyn, so multiple operations to an action which should be one: "Get me that dependencies so that I can edit and fix bugs, too".
That's what tsmono is about. just say dependency "foo" and git it from local dir falling back to npm resolving.

Neither does running within tmp/tool ```../../node_modules/.bin/ts-node -P tsconfig.json  tool/tool.ts``` work. It fails to find the .ts files configured by "paths" in tsconfig.json

Tested on linux only for now.

### LINKED ISSUE #876: Problems with pnpm due to removal of realpath function
Issue https://github.com/TypeStrong/ts-node/issues/797 was resolved in commit https://github.com/TypeStrong/ts-node/commit/9691206def28405d4e53c5a7753f96d15be46e79 by removing the `realpath` option from the `host` service parameter for the TypeScript language server. However, this breaking change is not compatible with the package manager [pnpm](https://github.com/pnpm/pnpm), which makes extensive use of symlinks in `node_modules`.

This is a simple example to reproduce this problem:
```shell
pnpm init -y
pnpm install --save ts-node typescript stellar-sdk @types/node 
```

Then create a file `index.ts` with the content:
```ts
import * as stellar from "stellar-sdk";
console.log(stellar.Network);
```

The TypeScript compiler will run without any problems, i.e., executing the following command will finish successfully:

```
./node_modules/.bin/tsc index.ts --noEmit
```

However, when running
```
./node_modules/.bin/ts-node index.ts
```
then this will yield the error message:

```
TSError: ⨯ Unable to compile TypeScript:


2 console.log(stellar.Network);
                      ~~~~~~~
```

## Reason
The package `stellar-sdk` has another dependency `stellar-base`. The class `Network` is defined in `stellar-base` and re-exported by `stellar-sdk`. pnpm will structure `node_modules` as follows (arrows indicate symlinks):

```
node_modules
 |
 +-- stellar-sdk -> .registry.npmjs.org/stellar-sdk/2.3.0/node_modules/stellar-sdk
 |
 +-- .registry.npmjs.org
     |
     +-- stellar-sdk/2.3.0/node_modules
     |   |
     |   +-- stellar-base -> ../../../stellar-base/1.1.2/node_modules/stellar-base
     |   |
     |   +-- stellar-sdk (actual package content)
     |
     +-- stellar-base/1.1.2/node_modules
         |
         +-- stellar-base (actual package content)
```

When running `ts-node` the TypeScript language server is not able to load `stellar-base` as symlinks are not correctly resolved: TypeScript will first resolve the module name `stellar-sdk` found in `index.ts` to `<project>/node_modules/stellar-sdk/lib/index.d.ts` but then is unable to resolve the real path of this file (see [here](https://github.com/microsoft/TypeScript/blob/01e1b1bb276f48eaa0612a34dd8f1eec5d86bd49/src/compiler/moduleNameResolver.ts#L975) and [here](https://github.com/microsoft/TypeScript/blob/01e1b1bb276f48eaa0612a34dd8f1eec5d86bd49/src/compiler/moduleNameResolver.ts#L958)). When reading `<project>/node_modules/stellar-sdk/lib/index.d.ts`, it finds a reference to `stellar-base` and tries to resolve it but the module resolution algorithm will not be able to find a package `stellar-base` inside a `node_modules` folder that is a parent (or grandparent) directory of `<project>/node_modules/stellar-sdk/lib/index.d.ts`.

Observe that `stellar-base` can be found if `<project>/node_modules/stellar-sdk/lib/index.d.ts` is correctly resolved to `<project>/node_modules/.registry.npmjs.org/stellar-sdk/2.3.0/node_modules/stellar-sdk/lib`

### LINKED ISSUE #956: Unable to require .ts nested within a node_modules directory
### Expected Behavior
No errors.

### Actual Behavior
"Unable to require file" error.

1. Create a package (e.g. using `npm pack`) with the following files:
- `npm-postinstall.ts`:
```
console.log('hello');
import * as x from './x';
console.log('hello2', typeof x);
```

- `package.json`:
```
{
  "name": "package1",
  "version": "0.0.0",
  "scripts": {
    "postinstall": "ts-node --skip-ignore npm-postinstall"
  },
  "dependencies": {
    "ts-node": "^8.6.2",
    "typescript": "^3.7.5"
  }
}
```

- `x.ts`:
```
export const x = 1;
```

2. Install that package: `npm install package1-0.0.0.tgz`

Works fine on Windows

### Specifications

* ts-node version: 8.6.2
* node version: 12.15.0
* Operating system and version: Ubuntu

### LINKED ISSUE #990: failed to import TS file from node_modules on Linux & Mac
## Expected Behavior

I want to import everything (including **TS source codes** in `node_modules/`) :

```javascript
register({
    ignore: [],
    compilerOptions: { module: 'CommonJS' }
});
```
https://github.com/EasyWebApp/MarkCell/blob/7b43ea0/source/File.ts#L26-L29

## Actual Behavior

| OS           | result                                                                                                     |
|:----------:|:---------------------------------------------------------------------------:|
| Windows | everything is OK                                                                                   |
| Linux       | `TypeError: Unable to require file: node_modules/xxx/source/yyy.tsx` |
| Mac        | `TypeError: Unable to require file: node_modules/xxx/source/yyy.tsx` |

### error on Linux

https://travis-ci.com/github/EasyWebApp/mark-wiki/builds/155135722

### error on Mac

image

## Steps to reproduce the problem

```shell
git clone https://github.com/EasyWebApp/mark-wiki.git
cd mark-wiki/
npm install
npm run build
```

## Minimal reproduction

https://github.com/TypeStrong/ts-node-repros/pull/1

## Specifications

* ts-node version: 8.8.1
* TypeScript version: 3.8.3
```json
{
    "compilerOptions": {
        "module": "ESNext",
        "moduleResolution": "Node",
        "resolveJsonModule": true,
        "experimentalDecorators": true,
        "jsx": "react",
        "jsxFactory": "createCell",
        "target": "ES5",
        "lib": ["ES2019", "DOM"]
    },
    "include": ["source/**/*"]
}
```
* node version: 12.16.1
* If Windows, are you using WSL or WSL2?: No

### LINKED ISSUE #1093: Different compilation results from tsc
### Expected Behavior

no errors

### Actual Behavior

`src/api/main.ts:12:18 - error TS2339: Property 'sendFormatted' does not exist on type 'Response'.`

try to augment 'express-serve-static-core' module through a d.ts file

### Code

```ts
// types.d.ts

declare module 'express-serve-static-core' {
    import { ResponseFormatterData } from './middlewares/response';

    interface Response {
        sendFormatted(data?: ResponseFormatterData) : Response<any>
    }
}
```

### Details

node : v14.7.0
npm : 6.14.7
tsc : 3.9.7
ts-node : v8.10.2

Win 10

### Related Issues
#391

cspotcode:
The various `realpath` issues we've seen don't strictly relate to symlinks.

They're caused by TypeScript considering a file to be an "external library import" when it's nested within a `node_modules` directory, and refusing to emit `.js` for it.  Sometimes a symlink's path does not include `node_modules` and its resolved path does, or vice versa, so enabling or disabling `realpath` causes or fixes this bug.

The "external library import" determination is handled by the module resolver.  If we implement our own in the `LanguageServiceHost`, we can tell the compiler to consider all `rootFileNames` to be "internal" no matter what.

This lets us re-enable `realpath`.

Related / should fix:
fix #876 
fix #797 
fix #956 
fix #819 
fix #990 
fix #1093

---

## Notes for code review

It's been a while since I started this PR, and I've since learned more about the compiler.

`rootFiles` are always emitted, *but* we want to avoid modifying `rootFiles` as much as possible, because it throws away a bunch of compiler state.  How much, precisely, I'm not sure, but modify `rootFiles` is way slower than leaving it as-is `rootFiles`.

This PR hooks into TypeScript's resolver and does 2x things:
(a) forces some modules to be classified internal using our own logic, overriding TypeScript's logic
(b) keeps track of modules that have *already* been classified as internal

(a) enables compiling files nested within ./node_modules
(b) allows us to avoid modifying rootFiles unnecessary

TS files in node_modules are always forced to be internal, because I can't see any downsides to this.
JS files in node_modules are *only* forced internal on-demand (when required), because the "internal/external" distinction affects maxNodeModuleJsDepth, which affects typechecking.

---

## Old, irrelevant notes

*I hate deleting old research in case I need it later, but stuff below this line is not relevant anymore.*

#### Gotchas / notes

*(this is not true anymore)* Implementation overrides `isExternalLibraryImport` according to our `ignored()` logic, which is the "nuclear option" described here:
https://github.com/microsoft/TypeScript/issues/28432#issuecomment-584405985
There's [another option](https://github.com/microsoft/TypeScript/issues/28432#issuecomment-437498123): `node_modules` nested *within* the `rootDir` are excluded, but if `rootDir` has `node_modules` in its own path, it is still compiled.


arcanis:
Of note, TS has itself a bug with workspaces being marked as external sources (cf https://github.com/microsoft/TypeScript/issues/37270). Maybe the `isExternalLibraryImport` override here could workaround that as well?

cspotcode:
https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L688-L710

I noticed some weirdness in the `compilerHost` codepath which I'll need to fix.  Don't have time now, so I'm jotting it down.

The bug is: `rootFileNames` are updated and the program rebuilt *only* when the builder program doesn't already have a sourcefile for the require()d module.

The problem is that a sourcefile might exist, but it might be considered "external."  When this happens, typescript refuses to emit .js for it.  We can force TS to emit `.js` by adding the file to `rootFileNames`.  Unfortunately this is slow because it triggers a program rebuild that should not otherwise be necessary.  (see #970)

If #970 is merged, this problem goes away.  Alternatively, we keep track of which files are classified "external" and we add them to `rootFileNames` only when they must be reclassified as internal.

cspotcode:
Importantly, we must ensure that tests properly trigger all possible patterns for a file being classified external, being omitted from rootFileNames, being unknown to the compiler, etc.

* `require()` outside the typechecker
* `import` from `node_modules`
* something in `node_modules` import something else in `node_modules` (the latter is classified external b/c is not in `rootFileNames`)

development-docs/isExternalLibraryImport.md
```diff
@@ -0,0 +1,63 @@
+## How we override isExternalLibraryImport
```
cspotcode, line 1: This file is notes to myself and any other ts-node developers.  Given the audience, it's not as polishes as the rest of ts-node, and I can remove it if needed.

src/index.spec.ts
```diff
@@ -545,20 +546,29 @@ describe('ts-node', function () {
...
+          jsmre: { name: 'javascript-module-required-externally' }
+        }, null, 2))
+        done()
       })
     })
 
-    it('should transpile files inside a node_modules directory when not ignored', function (done) {
-      exec(`${cmd} --skip-ignore tests/from-node-modules/from-node-modules`, function (err, stdout, stderr) {
+    it('should respect maxNodeModulesJsDepth', function (done) {
+      exec(`${cmdNoProject} --script-mode tests/maxnodemodulesjsdepth`, function (err, stdout, stderr) {
```
cspotcode, line 571: The magic happens in `./tests/maxnodemodulesjsdepth/`.  As long as typechecking succeeds, the test passes.

src/index.spec.ts
```diff
@@ -567,17 +577,17 @@ describe('ts-node', function () {
...
     let registered: tsNodeTypes.Register
+    let moduleTestPath: string
     before(() => {
       registered = register({
         project: PROJECT,
         compilerOptions: {
           jsx: 'preserve'
         }
       })
+      moduleTestPath = require.resolve('../tests/module')
```
cspotcode, line 588: Frankly I don't know how this was passing before this change.  It should only resolve after we've registered ts-node.  Maybe a race condition, where another test case already installed registered?

src/index.spec.ts
```diff
@@ -567,17 +577,17 @@ describe('ts-node', function () {
...
     })
 
-    const moduleTestPath = require.resolve('../tests/module')
-
```
cspotcode, line 580: Frankly I don't know how this was passing before this change. It should only resolve after we've registered ts-node. Maybe a race condition, where another test case already installed registered?

src/index.ts
```diff
@@ -498,6 +510,102 @@ export function create (rawOptions: CreateOptions = {}): Register {
   let getOutput: (code: string, fileName: string) => SourceOutput
   let getTypeInfo: (_code: string, _fileName: string, _position: number) => TypeInfo
 
+  const getCanonicalFileName = (ts as unknown as TSInternal).createGetCanonicalFileName(ts.sys.useCaseSensitiveFileNames)
+
+  // In a factory because these are shared across both CompilerHost and LanguageService codepaths
```
cspotcode, line 515: IMHO we should extract more logic into this style of "factory:" reusable bits of functionality that can be shared across our codepaths.

src/index.ts
```diff
@@ -546,21 +655,27 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
 
       const updateMemoryCache = (contents: string, fileName: string) => {
-        // Add to `rootFiles` if not already there
-        // This is necessary to force TS to emit output
-        if (!rootFileNames.has(fileName)) {
+        // Add to `rootFiles` as necessary, either to make TS include a file it has not seen,
+        // or to trigger a re-classification of files from external to internal.
+        if (!rootFileNames.has(fileName) && !isFileKnownToBeInternal(fileName)) {
+          markBucketOfFilenameInternal(fileName)
           rootFileNames.add(fileName)
```
cspotcode, line 679: Mutating rootFiles seems to throw away a lot of internal compiler state.  I don't know 100% how much, but I think avoiding mutating rootFiles is good for performance.

src/index.ts
```diff
@@ -632,13 +747,15 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
+      const sys: _ts.System & _ts.FormatDiagnosticsHost = {
         ...ts.sys,
         ...diagnosticHost,
         readFile: (fileName: string) => {
           const cacheContents = fileContents.get(fileName)
           if (cacheContents !== undefined) return cacheContents
-          return cachedReadFile(fileName)
+          const contents = cachedReadFile(fileName)
+          if (contents) fileContents.set(fileName, contents)
+          return contents
```
cspotcode, line 758: We use `fileContents` elsewhere to know if the compiler thinks a file has "changed."  If the compiler observes a file here, then we need to remember so that our other logic correctly realizes the file has not changed.

src/index.ts
```diff
@@ -687,17 +807,22 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
+          markBucketOfFilenameInternal(fileName)
           rootFileNames.add(fileName)
+          addedToRootFileNames = true
         }
 
         // Update program when file changes.
-        if (sourceFile === undefined || sourceFile.text !== contents) {
+        if (addedToRootFileNames || contentsChanged) {
           builderProgram = ts.createEmitAndSemanticDiagnosticsBuilderProgram(
             Array.from(rootFileNames),
```
cspotcode, line 827: Just being a bit smarter about when a file must be added to `rootFileNames`.  Even if the compiler has already seen a file, it's possible it was marked external.

cspotcode:
@blakeembrey this is finally(!!) done, tested, and ready for a review.  I added my own review comments to hopefully make the changes clearer.

cspotcode:
@blakeembrey Despite the red X, tests are passing.  The only failure is coverage, and I think that's only because compiler docs ask us to implement a function that the compiler never actually calls.

If you think you'll be busy for the next week, and if this sounds ok to you, I was thinking of asking for a code review on the TypeScript Discord.  Compiler experts hang out there, so I think I'd be able to find a reviewer with the right skillset.

src/index.ts
```diff
@@ -515,6 +527,102 @@ export function create (rawOptions: CreateOptions = {}): Register {
   let getOutput: (code: string, fileName: string) => SourceOutput
   let getTypeInfo: (_code: string, _fileName: string, _position: number) => TypeInfo
 
+  const getCanonicalFileName = (ts as unknown as TSInternal).createGetCanonicalFileName(ts.sys.useCaseSensitiveFileNames)
```
Gerrit0, line 530: This seems dangerous, shouldn't case sensitivity be determined based on the OS, or potentially the [forceConsistentCasingInFileNames](https://www.typescriptlang.org/tsconfig#forceConsistentCasingInFileNames) option?

cspotcode: Fair question.  `ts.sys.useCaseSensitiveFileNames` is OS-dependent: true on Posix, false on Windows.  As far as I can tell, `createGetCanonicalFileName` is how the compiler internally converts filenames to cache keys.

cspotcode: As far as I know, `forceConsistentCasingInFileNames` generates additional semantic diagnostics but doesn't affect resolver behavior.  Even when `forceConsistentCasingInFileNames` is turned off, `FOO.ts` and `foo.ts` still resolve to the same file on Windows, right?

Gerrit0: Ah, that makes sense. I was reading `useCaseSensitiveFileNames` as if there was another `useCaseInsensitiveFileNames` option, which it makes sense that there isn't. This should be fine then.

blakeembrey:
> If you think you'll be busy for the next week, and if this sounds ok to you, I was thinking of asking for a code review on the TypeScript Discord.

Absolutely feel free to do this anyway!

src/index.ts
```diff
@@ -1,10 +1,10 @@
-import { relative, basename, extname, resolve, dirname, join } from 'path'
+import { relative, basename, extname, resolve, dirname, join, isAbsolute } from 'path'
 import sourceMapSupport = require('source-map-support')
 import * as ynModule from 'yn'
 import { BaseError } from 'make-error'
 import * as util from 'util'
 import { fileURLToPath } from 'url'
-import * as _ts from 'typescript'
+import type * as _ts from 'typescript'
```
blakeembrey, line 7: 🙏

src/index.ts
```diff
@@ -515,6 +527,102 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
+  // In a factory because these are shared across both CompilerHost and LanguageService codepaths
+  function createResolverFunctions (serviceHost: _ts.ModuleResolutionHost) {
+    const moduleResolutionCache = ts.createModuleResolutionCache(cwd, getCanonicalFileName, config.options)
+    const knownInternalFilenames = new Set<string>()
+    /** "Buckets" (module directories) whose contents should be marked "internal" */
+    const internalBuckets = new Set<string>()
+
+    // Get bucket for a source filename.  Bucket is the containing `./node_modules/*/` directory
+    // For '/project/node_modules/foo/node_modules/bar/lib/index.js' bucket is '/project/node_modules/foo/node_modules/bar/'
+    const moduleBucketRe = /.*\/node_modules\/[^\/]+\//
```
blakeembrey, line 541: What about namespaced `@` node modules?

cspotcode: Ooh, good catch.  I'll add logic for this.  I should also add a test with 2x peer modules, @org/foo and @org/bar, and make sure one bucket doesn't affect the other.

PR #976, opened 2020-03-07
Update command for watching files in mocha

mattcasey:
Mocha seems to ignore the watch rule unless I add a comma after src, as was given in this example: https://github.com/TypeStrong/ts-node/issues/266#issuecomment-583130571. I have tested it on my own projects. This documentation saved me a lot of time!

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ mocha --require ts-node/register --watch-extensions ts,tsx "test/**/*.{ts,tsx}" 
 Mocha 7
 
 ```sh
-mocha --require ts-node/register --extensions ts,tsx --watch --watch-files src 'tests/**/*.{ts,tsx}' [...args]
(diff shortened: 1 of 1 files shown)


mattcasey:
@cspotcode I wanted you to see this since you were asking if anyone had tested the command.

cspotcode:
I have a few questions to narrow down what's going on here.

Why does adding the comma work?  Is this a mocha bug?  Is the comma parsing into an array of strings?  If so, is mocha watching the following globs: `["src", ""]`?  If so, is it possible that passing the empty string is sufficient and we can omit `src` entirely?  For example `--watch-files ''` or `--watch-files=`

mattcasey:
Closing.. argh sorry, i didn't have time to figure out why the original wasn't working for me.

PR #980, opened 2020-03-13
Strip ts-node callstack from reported error

Widdershin:
Hi!

I am a big fan of this library.

For years now I have applied this patch by hand to local copies of this module, since I use small terminal panes and the traceback generally means I have to scroll up to see any errors.

With this patch, the traceback is removed from the reported error.

I have been using this for a few years with no issues, but it could probably use improving if it were to be included.

Figured I would just send this your way in case it appeals to the team of maintainers.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -423,6 +423,13 @@ export function create (rawOptions: CreateOptions = {}): Register {
 
   function reportTSError (configDiagnosticList: _ts.Diagnostic[]) {
     const error = createTSError(configDiagnosticList)
+
+    const stackLines = error.stack.split('\n')
+    const lastLine = stackLines.findIndex(line => line.includes('ts-node'))
+    if (lastLine !== -1) {
+      error.stack = stackLines.slice(0, lastLine).join('\n');
+    }
+
     if (options.logError) {
       // Print error in red color and continue execution.
       console.error('\x1b[31m%s\x1b[0m', error)


src/index.ts
```diff
@@ -423,6 +423,13 @@ export function create (rawOptions: CreateOptions = {}): Register {
 
   function reportTSError (configDiagnosticList: _ts.Diagnostic[]) {
     const error = createTSError(configDiagnosticList)
+
+    const stackLines = error.stack.split('\n')
+    const lastLine = stackLines.findIndex(line => line.includes('ts-node'))
```
blakeembrey: I don't think this is a safe enough check unfortunately.

src/index.ts
```diff
@@ -423,6 +423,13 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
+    if (lastLine !== -1) {
+      error.stack = stackLines.slice(0, lastLine).join('\n')
```
blakeembrey: Can't an error go across files (e.g. have encountered `ts-node` multiple times)?

cspotcode:
What if, rather than potentially dangerous call-stack mutation, we try to flatten the call stack as much as possible for well-known errors?  We would *only* do this for predicted, well-understood errors such as a semantic diagnostics from the TS compiler.

For example, the following call-stack can be shortened from this:
```

/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:413
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
foo.ts:2:1 - error TS2588: Cannot assign to 'a' because it is a constant.

2 a = 456;
  ~

    at createTSError (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:413:12)
    at reportTSError (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:417:19)
    at getOutput (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:522:36)
    at Object.compile (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:727:32)
    at Module.m._compile (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:806:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Object.require.extensions.<computed> [as .ts] (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:809:12)       at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
```

to this:
```

/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:806:43
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
foo.ts:2:1 - error TS2588: Cannot assign to 'a' because it is a constant.

2 a = 456;
  ~

    at Module.m._compile (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:806:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Object.require.extensions.<computed> [as .ts] (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:809:12)       at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
```

We can implement this a few ways, but one is by internally returning errors instead of throwing, then adding logic to `_compile` to `throw` as appropriate.

cspotcode:
@Widdershin Do you have a minute to try out my branch from #981 and let me know what you think?

```
npm install --save TypeStrong/ts-node#ab/shorten-call-stack-for-ts-diagnostics
```

blakeembrey:
My bad, my brain glossed over the fact that this is diagnostic errors only. I feel less bad about that, but also confused - we have a custom inspect implementation that I thought worked and removed the stack altogether for these errors. Maybe something changed in node over time.

cspotcode:
Looks like the custom inspector only takes effect when in a REPL.

Node repl:
```
> require('/home/cspotcode/n/lib/node_modules/ts-node/register')
{}
> require('./foo')
Uncaught:
foo.ts:2:1 - error TS2588: Cannot assign to 'a' because it is a constant.

2 a = 456;
  ~
>
```

ts-node REPL:
```
> require('./foo.ts')
foo.ts:2:1 - error TS2588: Cannot assign to 'a' because it is a constant.

2 a = 456;
  ~

undefined
>
```

`ts-node ./foo.ts`
```

/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:413
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
foo.ts:2:1 - error TS2588: Cannot assign to 'a' because it is a constant.

2 a = 456;
  ~

    at createTSError (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:413:12)
    at reportTSError (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:417:19)
    at getOutput (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:522:36)
    at Object.compile (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:727:32)
    at Module.m._compile (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:806:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:1171:10)
    at Object.require.extensions.<computed> [as .ts] (/home/cspotcode/n/lib/node_modules/ts-node/src/index.ts:809:12)
    at Module.load (internal/modules/cjs/loader.js:1000:32)
    at Function.Module._load (internal/modules/cjs/loader.js:899:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
```

Widdershin:
@cspotcode I tried your branch and it's better but still fills my terminal with tracebacks.

image

That's the full height of the pane where the output is displayed, so I would have to scroll up quite a bit.

Widdershin:
I can appreciate that my edit is a bit of a hack so I like that y'all are thinking through what a well rounded solution might look like.

Just as a bit of end user feedback, I have never needed any of the information from this trace, nor have I missed it when I've removed it.

It's possible that other libraries haven't removed it because it's the default behaviour and editing the stack feels wrong, but this has worked for me.

If there doesn't end up being a solution that removes the stack entirely, I'll likely just maintain a fork so it's no real issue :)

src/index.ts
Widdershin: As it works right now this code drops the entire stack trace, since `ts-node` always has the last frame on the callstack, so the only case where this hack would work out poorly is if "ts-node" was present in the TypeScript error, which I could totally see occurring.

This was more of a "good enough" solution for what I needed, I wouldn't want to pretend this is foolproof.

cspotcode:
@Widdershin Some of these really long stacks are caused by manipulation of `Error.stackTraceLimit`, right?  It defaults to 10, so I don't think you should be seeing stacks that long unless you or a third-party library are opting-in to that behavior.

---

ts-node's `require()` implementation needs to fill the same role as a normal `require()`, with matching behavior.

If you try to `require()` a `.js` file with a syntax error, then node throws an error which is handled in whatever way node handles errors.  It logs them to the console.  The error's stack obeys the value of `Error.stackTraceLimit`.

Similarly, if you try to `require()` a `.ts` file with a syntax error (or a semantic error if you aren't using `transpileOnly`) we need to do the same.  The default formatting should be similar to node's default formatting, and it should similarly obey `Error.stackTraceLimit`.

More aggressive error manipulation can always be implemented by a third-party require hook that is loaded alongside ts-node.  Then you can `ts-node -r my-third-party-error-shortener ./foo.ts`  This interop with third-party solutions is possible because our `require()` implementation matches the behavior of the built-in ones and throws normal-looking errors.

PR #993, opened 2020-04-02
change order of npm installation

jacobdcastro:
After attempting to copy/paste local installs, I hit an error. ts-node requires a peer dependency of typescript. Therefore, I had to install typescript before ts-node for it successfully install. It makes sense to order these appropriately in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,12 +11,12 @@
 
 ```sh
 # Locally in your project.
-npm install -D ts-node
 npm install -D typescript
+npm install -D ts-node
 
 # Or globally with TypeScript.
-npm install -g ts-node
 npm install -g typescript
+npm install -g ts-node
 ```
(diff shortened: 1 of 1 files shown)


cspotcode:
Interesting, I thought yarn was the only one that threw an error when you omit a `peerDependency`. 
 Does npm outright disallow this, or merely log a warning?

jacobdcastro:
It gave me an error outright and didn't install ts-node at all until I installed typecript. I attempted twice with the same result before installing typescript first.

Although, just a minute ago I tried again and worked fine. Possible random glitch earlier.

cspotcode:
image

blakeembrey:
Feel free to merge either way, seems like it wouldn't hurt 👍

PR #994, opened 2020-04-02
Add timestamp to debug logging

sylc:
This PR add timestamp for debug logging.

example output is 
```
[ts-node 2020-04-02T12:35:37.097Z] directoryExists c/ts-node/node_modules/events 13
``` 

Associated with https://github.com/TypeStrong/ts-node/issues/754#issuecomment-606677376

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -38,7 +38,11 @@ function yn (value: string | undefined) {
  * Debugging `ts-node`.
  */
 const shouldDebug = yn(process.env.TS_NODE_DEBUG)
-const debug = shouldDebug ? console.log.bind(console, 'ts-node') : () => undefined
+const timestamp = function(){};
+timestamp.toString = function(){
+  return "[ts-node " + (new Date).toISOString() + "]";
+}
+const debug = shouldDebug ? console.log.bind(console, '%s', timestamp) : () => undefined
 const debugFn = shouldDebug ?
   <T, U> (key: string, fn: (arg: T) => U) => {
     let i = 0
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -38,7 +38,11 @@ function yn (value: string | undefined) {
  * Debugging `ts-node`.
  */
 const shouldDebug = yn(process.env.TS_NODE_DEBUG)
-const debug = shouldDebug ? console.log.bind(console, 'ts-node') : () => undefined
+const timestamp = function(){};
+timestamp.toString = function(){
+  return "[ts-node " + (new Date).toISOString() + "]";
+}
```
cspotcode, line 44: This looks unnecessarily clever.  Let's keep it simple and call the function every time we need a timestamp.  `function timestamp() {return "ts-node ...`

src/index.ts
```diff
@@ -38,7 +38,11 @@ function yn (value: string | undefined) {
...
+const debug = shouldDebug ? console.log.bind(console, '%s', timestamp) : () => undefined
```
cspotcode, line 45: IMO we should simplifying by writing a function.

`function(...args) {console.log(timestamp(), ...args)}`

sylc:
I now actioned your comments.

src/index.ts
```diff
@@ -38,7 +38,9 @@ function yn (value: string | undefined) {
  * Debugging `ts-node`.
  */
 const shouldDebug = yn(process.env.TS_NODE_DEBUG)
-const debug = shouldDebug ? console.log.bind(console, 'ts-node') : () => undefined
+const debug = shouldDebug ?
+  (...args: any) => console.log(`ts-node ${new Date().toISOString()}`, ...args)
```
cspotcode, line 42: Square brackets need to be re-added.  Otherwise looks good.

sylc: done :)

Issue #996, opened 2020-04-04
projectVersion not bumped when rootFileNames changes


cspotcode:
### Expected Behavior

`getProjectVersion()` is incremented when a new file is appended to `getScriptFileNames()` array.

### Actual Behavior

`getProjectVersion()` is not incremented when a file is added to `getScriptFileNames()` during typechecking, before it has been `require()`d.

@blakeembrey it looks like this was changed here:
https://github.com/TypeStrong/ts-node/pull/985/
That pull request means that all files added to the in-memory cache are also added to `getScriptFileNames()` so they need to trigger a bump of `getProjectVersion()`  Otherwise it breaks the TS compiler's expectations, which might cause bugs down the road.

### Steps to reproduce the problem

Omitted a rigorous reproduction until I double-check with you that this change was, indeed, accidental.

If you have a `index.ts` that imports a `foo.ts`, and you do `ts-node ./index.ts`:
typechecking will see the import statement, and `updateMemoryCache` will be called to pull `foo.ts` into the cache, adding it to the `fileVersions` Map.

### Specifications

* ts-node version: 8.8.1
* TypeScript version: 3.8.3
* tsconfig.json, if you're using one: no tsconfig
* node version: 13.8.0
* Operating system and version: tested on Windows & Ubuntu via WSL2

blakeembrey:
Hmm, I believe I purposely reverted this in https://github.com/TypeStrong/ts-node/pull/985/commits/d133b3995c9b33fea5c994f69fa08ec6c94ec89f because I couldn't replicate any issue and wanted to avoid changing the logic you had added 😄 I can make this change again.

blakeembrey:
Actually, I take that back - it looks like I did change this in https://github.com/TypeStrong/ts-node/commit/2e9a2f170f554700470c7086089abb38b8d41adf and I believe it follows your expectation.

cspotcode:
Here's where I think we're missing a `projectVersion++`:
https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L489

That line is effectively adding a new entry to the `getScriptFileNames` array, which as far as I understand, is the set of "root files."  So every time we change that array, we're changing the `Program`, and we need to indicate this by bumping the `getProjectVersion`.  Otherwise TS will reuse an invalid `Program` instance.

[...]

Issue #884, opened 2019-09-11
8.2.0 regression


nazar-pc:
Here is my tsconfig.json:

<details>

```json
{
  "compilerOptions": {
    "module": "umd",
    "target": "esnext",
    "sourceMap": true,
    "forceConsistentCasingInFileNames": true,
    "noImplicitReturns": true,
    "noImplicitAny": true,
    "skipLibCheck": true,
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": false,
    "declaration": true,
    "sourceRoot": "src",
    "baseUrl": "src",
    "outDir": "dist",
    "strict": true,
    "noUnusedParameters": true,
    "noUnusedLocals": true,
    "lib": [
      "es2015",
      "es2017.object",
      "dom"
    ],
    "typeRoots": [
      "node_modules/@types",
      "types"
    ],
    "plugins": [
      {
        "name": "typescript-tslint-plugin"
      }
    ]
  },
  "typedocOptions": {
    "mode": "modules",
    "out": "docs"
  },
  "include": [
    "src"
  ]
}
```

</details>

Code sample:
```typescript
const timeout = setTimeout(() => {}, 0);
if (timeout.unref) {
  timeout.unref();
}
```

With 8.1.1 it works fine, with 8.2.0 and 8.3.0 it fails with:
```
error TS2339: Property 'unref' does not exist on type 'number'.
```

Even though project has `@types/node` installed and `tsc -b` works fine and `tslint` in strict mode doesn't complain either.

blakeembrey:
Seems like https://github.com/TypeStrong/ts-node/commit/9691206def28405d4e53c5a7753f96d15be46e79 then, which itself was an unexpected breaking change in 8.1.0, it can be re-added in 9.0.

blakeembrey:
Can you check by running on 8.0.2?

blakeembrey:
Personally, this looks more like you added “dom” to the list of types than a change in versions. I’m pretty sure if you remove DOM it’ll work.

nazar-pc:
8.0.2 work fine too.

I can't remove DOM because project targets both environments and a lot of other things will fail to compile because of that.

blakeembrey:
Can you create a reproduction? The change I pointed out, which is the only change of code between the version you say broke, didn’t exist on 8.0.2 which is why I asked you to test it. If that works, there must be something else going on.

blakeembrey:
[...]

PR #999, opened 2020-04-07
Make #884 regression test repro correctly; also fix the bug

cspotcode:
Related to #996.  The fix for #996 causes the #884 regression test to start failing.  This PR shows that the underlying issue tested by this regression test is already happening on master.


cspotcode:
Added code changes to fix the failing test.

Issue #995, opened 2020-04-03
Is this really the best way to make a command line script with ts-node?


KasparEtter:
### Desired behavior

I would like to be able to have the following script at `~/projects/ts-hello/src/hello.ts`:

```ts
#!/usr/bin/env ts-node

import os from 'os';

console.log(`Hello, ${os.userInfo().username}!`);
console.log('Your arguments:', process.argv.slice(2));
```

… and have `npm` link the executable with the following configuration at `~/projects/ts-hello/package.json`:

```json
{
    "name": "ts-hello",
    "version": "0.0.0",
    "private": true,
    "bin": {
        "hello": "src/hello.ts"
    },
    "dependencies": {},
    "devDependencies": {
        "@types/node": "^13.9.8"
    }
}
```

For this to work, both `typescript` and `ts-node` have to be [installed globally](https://docs.npmjs.com/cli/install) (see more on this below):

```bash
$ npm install --global typescript ts-node
```

Additionally, we need to enable [ES module interoperability](https://www.typescriptlang.org/v2/tsconfig#esModuleInterop) with the following argument in `~/projects/ts-hello/tsconfig.json`:

```json
{
    "compilerOptions": {
        "esModuleInterop": true
    }
}
```

We can now change to the project directory, install the dependency and [link](https://docs.npmjs.com/cli/link) the package:

```bash
$ cd ~/projects/ts-hello
$ npm install
$ npm link
```

### Actual behavior

Since the motivation for this issue is in part to document the best approach for others (as far as I can tell [a](https://stackoverflow.com/questions/33535879/how-to-run-typescript-files-from-command-line/60486808) [lot](https://stackoverflow.com/questions/37271725/how-to-write-typescript-executable-file) [of](https://stackoverflow.com/questions/55820752/use-an-npm-script-as-bin) [people](https://stackoverflow.com/questions/58346595/error-while-executing-typescript-file-with-ts-node-after-installing-the-project) already struggled with this problem before me), I will elaborate step by step how to arrive at an acceptable solution so that search engines can index all the error messages as well.

First, let's test that the script works without relying on the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)):

```bash
$ cd ~/projects/ts-hello
$ ts-node src/hello.ts first second
Hello, <user>!
Your arguments: [ 'first', 'second' ]
```

The problems start when you move a directory up:

```bash
$ cd ..
$ pwd
~/projects
[...]

PR #1000, opened 2020-04-07
Improve the coverage of the README (#995)

KasparEtter:
Preview available at https://github.com/KasparEtter/ts-node/tree/readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,10 +224,10 @@ Example project structure:
 
 ```text
 <project_root>/
-- tsconfig.json
-- typings/
-  - <module_name>/
-    - index.d.ts
+-- tsconfig.json
+-- typings/
+  -- <module_name>/
+    -- index.d.ts
 ```
 
 Example module declaration file:
--- a/README.md
+++ b/README.md
@@ -23,6 +23,8 @@ npm install -g ts-node
 
 ## Usage
 
+### Shell
+
 ```sh
 # Execute a script as `node` + `tsc`.
 ts-node script.ts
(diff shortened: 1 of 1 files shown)


cspotcode:
Thanks, I made a bunch of tweaks, reverted unrelated formatting changes.

Re the behavior of `--script-mode` when the script is a symlink, we should figure out if that behavior is intentional or a bug.  Better to iron that out first, than to potentially state that a bug is intended via our README.

README.md
```diff
@@ -161,11 +192,27 @@ _Environment variable denoted in parentheses._
...
+* `transformers` `_ts.CustomTransformers | ((p: _ts.Program) => _ts.CustomTransformers)`: An object with transformers or a function that accepts a program and returns an transformers object to pass to TypeScript. Function isn't available with `transpileOnly` flag
+* `readFile`: Custom TypeScript-compatible file reading function
+* `fileExists`: Custom TypeScript-compatible file existence function
+
+### Options via tsconfig.json
+
+Most options can be specified by a `"ts-node"` object in `tsconfig.json` using their programmatic, camelCase names. For example, to enable `--transpile-only`:
+
+```json
+// tsconfig.json
```
KasparEtter, line 206: Strictly speaking, comments are not allowed in JSON. Given the context is given in the line above, we could/should skip this here.

README.md
```diff
@@ -161,11 +192,27 @@ _Environment variable denoted in parentheses._
...
 * `--log-error` Logs TypeScript errors to stderr instead of throwing exceptions (`TS_NODE_LOG_ERROR`, default: `false`)
 
-### Programmatic Only Options
+### Programmatic-only Options
+
+* `transformers` `_ts.CustomTransformers | ((p: _ts.Program) => _ts.CustomTransformers)`: An object with transformers or a function that accepts a program and returns an transformers object to pass to TypeScript. Function isn't available with `transpileOnly` flag
+* `readFile`: Custom TypeScript-compatible file reading function
+* `fileExists`: Custom TypeScript-compatible file existence function
+
+### Options via tsconfig.json
```
KasparEtter, line 201: I intentionally had this without a title in the `CLI and Programmatic Options` section as my understanding is that this only applies to those options (but not the CLI-only or programmatic-only options). Now we are mixing with the titles and sections *what* options are available and *how* to provide them. But given that neither approach is perfect, just do you what you think is best.

README.md
```diff
@@ -122,36 +147,42 @@ Create a new Node.js configuration and add `-r ts-node/register` to "Node parame
...
 ### CLI Options
 
-Supports `--print`, `--eval`, `--require` and `--interactive` similar to the [node.js CLI options](https://nodejs.org/api/cli.html).
+`ts-node` supports `--print` (`-p`), `--eval` (`-e`), `--require` (`-r`) and `--interactive` (`-i`) similar to the [node.js CLI options](https://nodejs.org/api/cli.html).
 
-* `--help` Prints help text
-* `--version` Prints version information
+* `-h, --help` Prints the help text
+* `-v, --version` Prints the version. `-vv` prints node and typescript compiler versions, too
+* `-s, --script-mode` Resolve config relative to the directory of the passed script instead of the current directory. Changes default of `--dir`
```
KasparEtter, line 172: `Resolve` -> `Resolves` for consistency

KasparEtter:
> Re the behavior of --script-mode when the script is a symlink, we should figure out if that behavior is intentional or a bug.

I leave it to you whether you want to create a new issue for this or not. Progress on this would be appreciated, though. 🤓

PR #1008, opened 2020-04-12
Enable webpacking of ts-node

cspotcode:
I was curious if webpacking ts-node would increase startup speed.  I see a 15ms speedup on my Linux box.  Probably would see a bit more on Windows since fs performance is worse.

I don't think this is a high-priority task. I don't even know if it's a good idea. We can close it if anyone wants to clean up the PR queue.

## Tasks before this can be merged:
* [ ] refactor into a single bundle, so that `require('ts-node')` includes all functionality, and all `bin.ts` scripts are one-liners that invoke `index.ts` exports.  Should be done to avoid the current duplication in 2x bin and index bundles.  Or use webpack splitting to keep separate, so `require('ts-node')` doesn't need to load `bin` functionality?
* [ ] fix tests
* [ ] remove `./dist-non-bundled`?  Move deps to devDependencies?  Depends if we still want to publish the non-bundled for debugging.
* [ ] disable minimizing the bundle?  so easier to debug?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@ coverage/
 .DS_Store
 npm-debug.log
 dist/
+dist-non-bundled/
 tsconfig.schema.json
 tsconfig.schemastore-schema.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,12 @@
         }
       }
     },
+    "@types/anymatch": {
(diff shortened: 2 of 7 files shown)


cspotcode:
Closing to clean up the PR queue.

PR #1009, opened 2020-04-12
Fix #1004: resolve config relative to realpath of entrypoint

### LINKED ISSUE #1004: realpath of entry-point script is not resolved
*This issue is a messy brain-dump.  I'll clean it up later and ping people for review so they can read the clean version instead of this mess :D*

Investigate symlink resolution of the entry-point script and how it affects compilation and config file resolution.  E.g. something like `ts-node-script ./node_modules/.bin/this-is-a-symlink`

This is important if a script uses `ts-node-script` in the shebang and is symlinked to the path via something like npm's installation mechanism.  In that case, the symlink is in a totally different directory than the target script.  If we use the `realpath` we will discover the correct tsconfig file.  Otherwise we will search for tsconfig relative to the symlink's location.

### Matching node's behavior

We can use node's `package.json` resolution as a model to follow.  If it is using `realpath` then we should probably do the same.  It has `--preserve-symlinks` and `--preserve-symlinks-main` flags.

> The --preserve-symlinks flag does not apply to the main module, which allows node --preserve-symlinks node_module/.bin/<foo> to work. To apply the same behavior for the main module, also use --preserve-symlinks-main. ([source](https://nodejs.org/api/cli.html#cli_preserve_symlinks))

It sounds like node almost always `realpath`s the entry-point, so we can do the same, even before we've loaded the config file.  From there, we can obey tsconfig's `preserveSymlinks`.

### Questions

* Do we override tsconfig's `preserveSymlinks` with node's `--preserve-symlinks`?  Is node's flag query-able at runtime?
* do we use `fs.realpath()` or `require.resolve` to locate the entry-point?  Latter is guaranteed to match node's behavior but requires us to register .ts and .tsx extensions *before* locating entry-point and config file.  (complexity)

### TODO

* merge other PR that adds `realpath`
* Test coverage for every situation where symlink resolving affects config resolution
  * tests for all other `realpath`-related tickets to make sure we don't break use-cases
* Add `realpath` or `require.resolve` to entry-point codepath *before* config file resolution.

The template below is incomplete; I'll fill in with reproductions when I'm able to come back to this.

Related discussion in #995

### Expected Behavior

If entry-point script is `./foo/index.ts`, which symlinks to `./bar/index.ts`, the latter path should be used for config file resolution, similar to how nodejs will locate the script's `package.json` relative to the realpath instead of the symlink path.

### Actual Behavior

TODO

TODO

### Minimal reproduction

TODO

### Specifications

* ts-node version:
* node version:
* TypeScript version:
```
{}
```

cspotcode:
Fixes #1004.

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -252,20 +252,27 @@ function getCwd (dir?: string, scriptMode?: boolean, scriptPath?: string) {
       throw new TypeError('Script mode cannot be combined with `--dir`')
     }
 
-    // Use node's own resolution behavior to ensure we follow symlinks
-    // This may affect which tsconfig we discover
-    // This happens before we are registered, so tell node's resolver to consider .ts and tsx files
-    // TODO in extremely rare cases, if a foo.js and foo.ts both exist, we may follow the wrong one,
-    // because we are not obeying `--prefer-ts-exts`
-    const hadTsExt = hasOwnProperty(require.extensions, '.ts') // tslint:disable-line
-    const hadTsxExt = hasOwnProperty(require.extensions, '.tsx') // tslint:disable-line
+    // Use node's own resolution behavior to ensure we follow symlinks.
+    // scriptPath may omit file extension or point to a directory with or without package.json.
+    // This happens before we are registered, so we tell node's resolver to consider ts, tsx, and jsx files.
+    // In extremely rare cases, is is technically possible to resolve the wrong directory,
+    // because we do not yet know preferTsExts, jsx, nor allowJs.
+    // See also, justification why this will not happen in real-world situations:
+    // https://github.com/TypeStrong/ts-node/pull/1009#issuecomment-613017081
+    const exts = ['.js', '.jsx', '.ts', '.tsx']
+    const extsTemporarilyInstalled: string[] = []
+    for (const ext of exts) {
+      if (!hasOwnProperty(require.extensions, ext)) { // tslint:disable-line
+        extsTemporarilyInstalled.push(ext)
+        require.extensions[ext] = function() {} // tslint:disable-line
+      }
+    }
     try {
-      if(!hadTsExt) require.extensions['.ts'] = function() {}  // tslint:disable-line
-      if(!hadTsxExt) require.extensions['.tsx'] = function() {} // tslint:disable-line
       return dirname(require.resolve(scriptPath))
     } finally {
-      if(!hadTsExt) delete require.extensions['.ts'] // tslint:disable-line
-      if(!hadTsxExt) delete require.extensions['.tsx'] // tslint:disable-line
+      for (const ext of extsTemporarilyInstalled) {
+        delete require.extensions[ext] // tslint:disable-line
+      }
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/bin.ts
```diff
@@ -251,7 +252,21 @@ function getCwd (dir?: string, scriptMode?: boolean, scriptPath?: string) {
       throw new TypeError('Script mode cannot be combined with `--dir`')
     }
 
-    return dirname(scriptPath)
```
blakeembrey, line 254: I think I’d be more comfortable noting that we only support symlinked directories instead of doing this, but up to you. If we only support directories it would become easy. Alternatively we can implement the resolver in a loop ourselves instead of messing with require.

blakeembrey: We might not even have TSX, JSX support extensions added either, so it’s another difference here. You could expose the function that creates a list of extensions, then just loop over that list and resolve like node.js does since we don’t need the extra require logic.

cspotcode: Do we need to support loading a directory containing an `index.*` file, or a directory containing a `package.json` with a `main` field pointing to another file?  E.g. `require('./directory-containing-package-json')`?

The chicken-and-egg problem with file extensions is a real bummer.  I'll try to come up with some realistic examples of this causing problems in real life.

The only reason I was hoping to support symlinked scripts is so they can be put on the `PATH`.  `npm` and `yarn` do this a lot, and I can see it coming up in monorepo situations. (lerna, yarn2)

cspotcode:
Here is my analysis for if `--script-mode` ambiguous config resolution will be a problem in real-world scenarios.

### Things we don't know that make it ambiguous

* whether or not .tsx or .jsx are enabled
* whether .ts(x) takes precedence over .js(x)

When resolution might match 2x different files, we may follow the wrong one, which may lead us to discover the wrong config.

### Ways ambiguity can happens

- 2x matched files are peers of each other, e.g. ./somedir/foo.jsx and ./somedir/foo.ts, *and* symlink to different target directories
- file is peer with directory of the same name, e.g. `/somedir/foo.tsx` and `/somedir/foo/package.json`

### Contrived examples

- `foo.jsx` and `foo.ts` peers in same directory, symlinked to *different* target directories.  `preferTsExts` and `allowJs` determine which one is resolved.
  - This seems extremely rare.  Sure, peers might exist, but they either wouldn't be symlinks, or they'd point to the same target directory
- `foo.jsx` and `foo/` in same location.  Do we look at `foo.jsx` or `foo/index.*`, or whatever `foo/package.json` says? (`foo.jsx` takes precedence, but jsx might not be enabled)

### Real-world examples

None

I can't think of a real-world situation where the contrived examples will occur.

### Mitigation strategies

Based on the above, I don't think ambiguity is a real problem.  But if we want to avoid issues, here are a few strategies:

- Do not support symlinked files, only symlinked directories
  - Breaks the `npm link` use-case
  - May break yarn 2 and lerna?

- Detect and throw an error whenever ambiguity affected config resolution
  - Throw when directory we resolve initially for config discovery is *different* than directory resolved to run the script (in all other cases, ambiguity has no real effect)
  - When anyone hits ambiguity in the real world, we will get a ping on our issue tracker
  - Only do this test when `TS_NODE_DEBUG` is enabled, to avoid impacting performance?

- Do nothing; allow ambiguity
  - Affected users can specify full path to avoid ambiguity

PR #1018, opened 2020-04-20
Report diagnostics for all files added to compilation, even if they are not require()d directly

cspotcode:


src/index.ts
```diff
@@ -530,22 +531,31 @@ export function create (rawOptions: CreateOptions = {}): Register {
...
-        // Get the relevant diagnostics - this is 3x faster than `getPreEmitDiagnostics`.
-        const diagnostics = service.getSemanticDiagnostics(fileName)
-          .concat(service.getSyntacticDiagnostics(fileName))
+        // Emit diagnostics for all newly-encountered files.
+        // This matches node, where require()ing one file that require()s many others will throw
+        // when *any* file has an error.
+        const diagnostics = []
+        for (const sourceFile of program!.getSourceFiles()) {
+          const { fileName } = sourceFile
+          if (!diagnosticsReportedFor.has(fileName)) {
```
blakeembrey, line 547: Nit: Should we invalidate diagnostics reported when the file version changes? Maybe we can store the version alongside it for this?

cspotcode: Yeah, good point.  A correct solution may be even more complicated than that.

We only update a file's contents when it has been deleted from `require.cache` *and* then is re-required.  But technically, should we invalidate our cache as soon as `require.cache` is mutated, even before a file is re-required?

What if adding `foo.ts` to compilation causes a diagnostic to appear in `bar.ts` due to declaration merging or `///` references?  What if `bar` was already required?

Maybe we can simplify.  What if we throw when we see *any* diagnostics *anywhere* in the program?  We can assume the diagnostics were triggered by the `require()` call, since all previous `require()` calls succeeded.  When an error happens, we *undo* the changes we made to the program so that the problematic file is removed from the program.  (remove it from `rootFiles` array)  This means if a user tries to `require()` the same file again, they get the same errors, but if they try to `require()` something else, they *don't* get the same errors.

```
try {
    require('./ts-file-causes-many-diagnostics.ts');
} catch(e) {}
require('./ts-file-does-not-cause-diagnostics.ts');
```

I expect behavior in most idiomatic TS codebases will be the same as `tsc && node`:  
*any* type error *anywhere* in the codebase will immediately throw, before any code executes.  Because in an idiomatic codebase, the compiler will immediately be forced to load the entire codebase via following `import` statements.

lastthyme (approved):

PR #1023, opened 2020-05-02
Make shebang example local installation compatible

mvasin:
`#!/usr/bin/env ts-node-script` shebang assumes you have ts-node installed globally, `#!/usr/bin/env npx ts-node-script` assumes you have npm version 5.2 or higher. I prefer the second assumption a bit more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ ts-node-transpile-only scripts.ts
 ### Shebang
 
 ```typescript
-#!/usr/bin/env ts-node-script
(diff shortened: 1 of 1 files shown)


cspotcode:
Doesn't work on Linux, unfortunately.  You can test in docker to confirm,
and check our issue tracker for other tickets where we've explained this.
It's a Linux kernel limitation.

mvasin:
Thanks for clarification! Didn't expect to find that in the issue tracker, my bad.

PR #1026, opened 2020-05-03
Fix package.json exports for `ts-node/register`

addaleax:
Commit f6cd5d4d696e broke `require('ts-node/register')` because
`exports:` provides an *exclusive* list of possible entry points,
and makes importing through other entry points impossible.

In particular, this breaks node-tap with TypeScript support.

Refs: https://medium.com/@forbeslindesay/is-promise-post-mortem-cab807f18dcc

@cspotcode Would be cool to see a release very very soon in order to un-break CI for lots of people :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,16 @@
   "main": "dist/index.js",
   "exports": {
     ".": "./dist/index.js",
-    "./esm": "./esm.mjs"
+    "./esm": "./esm.mjs",
+    "./register": "./register/index.js",
+    "./register/index": "./register/index.js",
+    "./register/index.js": "./register/index.js",
+    "./register/files": "./register/files.js",
+    "./register/files.js": "./register/files.js",
+    "./register/transpile-only": "./register/transpile-only.js",
+    "./register/transpile-only.js": "./register/transpile-only.js",
+    "./register/type-check": "./register/type-check.js",
+    "./register/type-check.js": "./register/type-check.js"
   },
   "types": "dist/index.d.ts",
   "bin": {


addaleax:
Also, if you’re interested https://github.com/nodejs/node/pull/33074/ contains proposed wording for Node’s ESM docs that would warn about this :)

package.json
```diff
@@ -5,7 +5,8 @@
   "main": "dist/index.js",
   "exports": {
     ".": "./dist/index.js",
-    "./esm": "./esm.mjs"
+    "./esm": "./esm.mjs",
+    "./register": "./register/index.js"
```
blakeembrey, line 9: @cspotcode We’ll also need to include the other `ts-node/register` options, not sure how often they’re used.

cspotcode: On it!  Will be pushing out a change in a few minutes.

addaleax: @blakeembrey Thanks, done!

cspotcode:
8.10.1 is out, should be fixed via #1027.

I plan to revisit this PR when I have time to learn more about the potential pitfalls of package.json "exports"

@addaleax thanks again for the quick bug report and PR!

cspotcode:
In the future, we might tweak our tests to be more E2E-style, so we `npm pack` and then `npm install`  into a temp project directory.  This allows testing the `ts-node` bin entrypoint and `require('ts-node/*')`.

addaleax:
@cspotcode Just fyi, it looks like there is actually an Easy Way Around This™ that just needs to be documented: https://twitter.com/MylesBorins/status/1257058330457133058

cspotcode:
@addaleax thanks, yeah, that seems to work.  So I guess each key in the `"exports"` dictionary is more of a prefix, and anything starting with that prefix will match?

Looks like the proper exports declaration for us is:
```
  "exports": {
    "./": "./",
    "./esm": "./esm.mjs"
  }
```

addaleax:
@cspotcode I’m not an expert here either, but that’s how I would read https://nodejs.org/api/esm.html#esm_subpath_exports, yes. :+1:

PR #1027, opened 2020-05-03
revert addition of package.json "exports" because it is a breaking ch…

cspotcode:
…ange and requires more thought

Fixes #1026.

I would still like to merge #1026 eventually, but for an immediate fix, this seems the safest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,10 +3,6 @@
   "version": "8.10.0",
   "description": "TypeScript execution environment and REPL for node.js, with source map support",
   "main": "dist/index.js",
-  "exports": {
-    ".": "./dist/index.js",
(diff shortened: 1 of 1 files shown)


cspotcode:
@addaleax I appreciate the very quick bug report and PR!  The medium post you linked mentioned issues with importing `./package.json` and I think I need more time to read about the implications.  But then I'll definitely come back to #1026 and get it merged.

addaleax:
@cspotcode #1026 doesn’t really apply as a patch anymore, so I’ve deleted the branch, but yeah, if you add an `"exports":` key, making sure that all possible entry points are listed should help a lot :)

PR #1028, opened 2020-05-04
Re-add "exports" declaration to package.json in backwards-compatible way

cspotcode:
See #1026


src/index.spec.ts
```diff
@@ -4,10 +4,12 @@ import { join } from 'path'
 import semver = require('semver')
 import ts = require('typescript')
 import proxyquire = require('proxyquire')
-import { register, create, VERSION } from './index'
+import * as tsNodeTypes from './index'
```
cspotcode, line 7: Type-only import, since we want to require() the locally-installed ts-node later.

src/index.spec.ts
```diff
@@ -35,6 +44,25 @@ describe('ts-node', function () {
...
+    testsDirRequire.resolve('ts-node/dist/bin-script')
+    testsDirRequire.resolve('ts-node/dist/bin-script-deprecated')
+    testsDirRequire.resolve('ts-node/dist/esm')
+    testsDirRequire.resolve('ts-node/register')
+    testsDirRequire.resolve('ts-node/register/index')
+    testsDirRequire.resolve('ts-node/register/files')
+    testsDirRequire.resolve('ts-node/register/transpile-only')
+    testsDirRequire.resolve('ts-node/register/type-check')
+    testsDirRequire.resolve('ts-node/tsconfig.schema.json')
+    testsDirRequire.resolve('ts-node/tsconfig.schemastore-schema.json')
```
cspotcode, line 64: Cannot load them, because they do things like install require hooks.  But resolving should be enough to know that package.json "exports" is still compatible with legacy require() calls.

src/index.spec.ts
```diff
@@ -523,11 +551,14 @@ describe('ts-node', function () {
...
-        jsx: 'preserve'
-      }
+    let registered: tsNodeTypes.Register
+    before(() => {
+      registered = register({
+        project: PROJECT,
+        compilerOptions: {
+          jsx: 'preserve'
+        }
+      })
```
cspotcode, line 561: Since ts-node is now being loaded inside a `before()` callback, we must also register ts-node inside a `before()` callback.

src/index.spec.ts
```diff
@@ -637,10 +668,11 @@ describe('ts-node', function () {
     })
 
     describe('JSX preserve', () => {
-      let old = require.extensions['.tsx'] // tslint:disable-line
+      let old: (m: Module, filename: string) => any
```
cspotcode, line 671: Must be moved inside a `before()` callback.

src/index.spec.ts
blakeembrey: Should we remove some of these in the next major version?

cspotcode: Good point, I've already merged other breaking changes to master, so our next release will be major already. 

Do you want me to tweak the `"exports"` object to only include a few?

Off the top of my head, here's a whitelist of the ones we should keep, what do you think @blakeembrey ?

```
ts-node

// only reliably way to ask node for the root path of a dependency is Path.resolve(require.resolve('ts-node/package'), '..')
ts-node/package

// All bin entrypoints for people who need to augment our CLI: `node -r otherstuff ./node_modules/ts-node/dist/bin`
ts-node/dist/bin
ts-node/dist/bin-transpile
ts-node/dist/bin-script

// Must be require()able, obviously
ts-node/register
ts-node/register/files
ts-node/register/transpile-only
ts-node/register/type-check

// Maybe this is useful?  I dunno; I'd hate to break someone
ts-node/tsconfig.schema
```

blakeembrey: SGTM. I think we could safely omit `tsconfig.schema` but add it back if requested, but happy to keep it there for anyone using it as a manually referenced schema.

blakeembrey: I'm assuming the reason we ship with `tsconfig.schema` at all is because of `$ref` or `$schema` in `tsconfig.json`? Maybe we could just omit it and have people reference a static version hosted outside the package going forward?

cspotcode: It's so you can add `"$schema": "./node_modules/ts-node/tsconfig.schema.json"` to your tsconfig file and *know* you are using the correct schema for the version of ts-node you are using.  For example, if you're using an old or experimental build that changes which ts-node  flags you are allowed to embed in a tsconfig, you'll want to refer to the bundled schema.  Even then, you're probably referencing it via filename, not `require()`.

For everyone else, their editor is using the schemastore schema by default, which already includes our stuff.

I'll omit it.

src/index.spec.ts
blakeembrey: `import type` possibly?

package.json
```diff
@@ -3,6 +3,23 @@
...
+  "exports": {
+    ".": "./dist/index.js",
+    "./package": "./package.json",
+    "./package.json": "./package.json",
+    "./dist/bin": "./dist/bin.js",
+    "./dist/bin.js": "./dist/bin.js",
+    "./dist/bin-transpile": "./dist/bin-transpile.js",
+    "./dist/bin-transpile.js": "./dist/bin-transpile.js",
+    "./dist/bin-script": "./dist/bin-script.js",
+    "./dist/bin-script.js": "./dist/bin-script.js",
```
blakeembrey, line 15: Should we provide these as exports or only allow `node node_modules/ts-node/dist/bin` style? Do we want both `.js` and without?

cspotcode: I wasn't sure how people are using `ts-node` in the wild, and I couldn't think of any compelling benefits that outweighed the possibility of breaking something.

PR #1031, opened 2020-05-09
Implement ERR_REQUIRE_ESM when a file is `require()`d which should be loaded as ESM

cspotcode:
Node's built-in `require.extensions['.js']` will detect if a file should be loaded as ESM instead of CommonJS, according to the file's associated `package.json`.  If it should be loaded as ESM, node throws an error.  This change makes our `require.extensions` throw the same error.

Mocha, for example, is relying on this error to detect when files are ESM and when they're CommonJS.
https://github.com/mochajs/mocha/blob/master/lib/esm-utils.js#L10-L22

This change detects when we're on a version of node that supports ESM, so it knows if it needs to (possibly) throw this error.  If so, it runs the same code as node to find the relevant `package.json` and see if the file should be treated as ESM.  If so, it throws an error that matches node's built-in error.  I did this by copy-pasting the relevant bits of node's source code.

Things I don't really love about this:
* ESM is imposing changes on our CommonJS hooks
* copy-pasting more code out of node's source
* feature-detection method.  Not sure I am using the best approach.
  * Alternative: KISS and check if node version is >=12 (even faster)
  * Alternative: we can bundle a subdirectory like the following, then attempt to `require()` the `.js` file and see if we get the error. (slower but is proper "feature detection")
```
./esm-detection
  ./package.json -> {"type": "module"}
  ./index.js -> empty file
```

Looking at how a few node versions behave:

node 12 always parses `import()` syntax, doesn't support native ESM without the `--experimental-modules`, but *always* throws the error if you try to `require()` a file that should be loaded as ESM.

node 10 & 11 always parses `import()` syntax, only loads ESM with `--experimental-modules`, and never honors the `package.json` `"type"` field.


cspotcode:
This is technically a breaking change.  Where previously we would attempt to execute `.ts` files as CJS even if the engine would otherwise honor `package.json` `"type": "module"`, now we will throw an error.

I can implement this change conditionally for now, so it's only turned on if our ESM loader is used.

Node 12's behavior is weirder than I thought depending on the minor version.  Based on the behavior, I think checking `process.versions.node` >=12 is best.  Other forms of feature detection will log warnings or give false negatives.

node 12.15
require()s ESM package executes as CJS *and* logs warning about not supporting ESM (we want to avoid the warning)

node 12.16 throws an `ERR_REQUIRE_ESM`
yet `package.json` "exports" are not implemented

dist-raw/node-cjs-loader-utils.js
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+function createErrRequireEsm(filename, parentPath, packageJsonPath) {
+  // Attempt to create an error object that matches node's native error close enough
+  const code = 'ERR_REQUIRE_ESM'
+  const err = new Error(getMessage(filename, parentPath, packageJsonPath))
+  err.name = `Error [${ code }]`
+  err.stack
+  Object.defineProperty(err, 'name', {
```
blakeembrey, line 88: This is pretty funky, might want to add a comment on why you do this since the first look makes it seem this line would erase `err.name = ` above.

dist-raw/node-cjs-loader-utils.js
```diff
@@ -0,0 +1,118 @@
+// copied from https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/loader.js
```
blakeembrey, line 1: Is this whole file copied or which parts should be reviewed? Curious on things like we read JSON ourselves instead of relying on the `require` cache.

cspotcode: Good point, this is actually coming from several files in node's source.  I added comments to each item in this file with a permalink to the span of node's source code on github.

dist-raw/node-cjs-loader-utils.js
cspotcode: Done

cspotcode:
@blakeembrey 

Thanks for the review.  I added comments explaining what's going on in `dist-raw/node-cjs-loader-utils.js`.

Re using the `require` cache to read package.json files:
I don't have a strong opinion either way.  Admittedly I was a tiny bit annoyed at having to duplicate more of node's source-code, so I wanted to copy-paste with as few modifications as possible, to make sure our stuff behaves as close as possible to node.

Would you prefer I tweak that code to use `require()`?  On second thought, seems like that would work perfectly fine.

PR #1032, opened 2020-05-10
Run tests e2e against an `npm pack && npm install`ed ts-node

cspotcode:
This is necessary to test that our entrypoints are working correctly with package.json "exports".

This change runs `npm pack` to create a tarball, then `npm install`s it into `tests/node_modules/`

The tests `require('ts-node/register')` and exec `./node_modules/.bin/ts-node` so it matches how a consumer invokes `ts-node`.


src/index.spec.ts
```diff
@@ -5,16 +5,36 @@ import semver = require('semver')
...
+  const tempDir = mkdtempSync(join(TEST_DIR, 'tmp'))
+  await execP(`npm pack --ignore-scripts "${ROOT_DIR}"`, { cwd: tempDir })
+  const tarballPath = join(tempDir, readdirSync(tempDir)[0])
+  writeFileSync(TARBALL_PATH, readFileSync(tarballPath))
+  unlinkSync(tarballPath)
+  rmdirSync(tempDir)
+  await execP(`npm install`, { cwd: TEST_DIR })
+  const packageLockPath = join(TEST_DIR, 'package-lock.json')
+  existsSync(packageLockPath) && unlinkSync(packageLockPath)
+})
```
blakeembrey, line 34: I wonder if these a way to do this outside of the test suite or possibly a separate set of tests? Either way this looks like a great start at coverage.

cspotcode: I can definitely move the `npm pack` into our `npm build`.  That would be better since it would `npm pack` using the same node and npm versions as we use to `publish`.

Are you thinking you want 2x sets of tests, one unit tests, the other doing these e2e-style tests against an `npm install`?

I thought about that but I wasn't sure how useful it'd be since the tests mostly look e2e-style already.  Your call though, happy to split things out.

blakeembrey: Let's keep it how it is then, I don't have strong opinions either way but it's great to know you thought about this.

cspotcode:
@blakeembrey This is ready for (one more) review.  I moved the `npm pack` into a build step instead of a test step.  Installation still happens at test time, in case the goal is to e2e test against an older version of node & npm.

Here are the changes since your last review to make it a little easier.
https://github.com/TypeStrong/ts-node/compare/2869699..b56dfde

PR #1034, opened 2020-05-10
Push coverage from GH Actions; remove TravisCI

cspotcode:
@blakeembrey Are there environment variables that need to be passed to coveralls?  I can't seem to see where they would be configured on TravisCI.

~~Also fixes failing lint step after upgrading typescript compiler version.
node 6 tests uncovered this since they do not obey package-lock.json~~ EDIT moved this to a separate PR #1035 

Do we need to preserve notifications behavior?
```
notifications:
  email:
    on_success: never
    on_failure: change
```


.github/workflows/continuous-integration.yml
```diff
@@ -40,7 +40,10 @@ jobs:
...
       # lint, build, test
       - run: npm install
+      - run: npm run lint
       - run: npm run build
       - run: npm rm tslint
-      - run: npm install ${{ matrix.typescript }}
-      - run: npm run test-cov
+      - run: npm install ${{ matrix.typescript }} --force
+      - run: set +e ; npm run test-cov ; echo $? > tests-exit-code
+      - run: npm install -g coveralls@2 && cat ./coverage/lcov.info | coveralls
```
blakeembrey, line 48: You can use `if: ${{ always() }}` instead: https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always.

blakeembrey:
> @blakeembrey Are there environment variables that need to be passed to coveralls?

I can't recall ever having to do anything... But I found this: https://github.com/nickmerwin/node-coveralls#usage. We should probably just use this: https://github.com/coverallsapp/github-action.

.github/workflows/continuous-integration.yml
cspotcode: Thanks, I was having so much trouble navigating the github actions docs.

blakeembrey:
I believe you can also grab `github-token` from `${{ github.token }}` for the Coveralls GitHub Action.

.github/workflows/continuous-integration.yml
blakeembrey: It's been pretty challenging, but I've been using it in some private monorepos lately so got a bunch of first-hand experience (which seems to be the only way to learn).

blakeembrey:
> Do we need to preserve notifications behavior?

It's not urgent, but if we can it's probably nice - I mostly disabled this across all repos due to the number of notifications I'd otherwise receive.

blakeembrey:
@cspotcode We should also consider creating a "nightly" GitHub Action next that tests the latest TypeScript build. We have it currently set up on Travis CI, though I've mostly not kept up with the failures.

cspotcode:
I'll look into a nightly test.

For notifications, I think that's configured on a user-by-user basis.  At least I can't find anything that lets me configure in the workflow yml.
image
https://github.com/settings/notifications

cspotcode:
@blakeembrey I think I have a nightly test set up correctly.

Do you know why it's saying coverage dropped so much?  Also, how do you typically use the coverage results?  Since it's for the emitted .js, I can't use the UI to visualize coverage.

cspotcode:
Nevermind my question about dropping coverage; it must have been caused by the intentially failing test.  Now coverage says it's good.

cspotcode:
@blakeembrey This is ready for a final rubber stamp.

blakeembrey:
> Also, how do you typically use the coverage results?

In this repo, I'm not sure since a lot of the tests run using the CLI - I've never actually double checked that we're measuring it correctly (pretty sure I wasn't). In general, I view it locally if it's well under what I expect. We could move it to properly measure coverage on TypeScript though, things have improved a lot since I originally set up `ts-node`.

PR #1036, opened 2020-05-11
[BREAKING] Bump minimum supported node version to 10

cspotcode:
Gonna make breaking changes as PRs labelled as such so I can merge them all when we make the next major release.


merceyz:
Should probably update the build target as well https://github.com/TypeStrong/ts-node/blob/196aafb002069d585e87e13891197db14c68d72f/tsconfig.json#L4
You can check https://node.green/ to see which target is suitable

PR #1040, opened 2020-05-12
Fix coveralls reporting so that only aggregate coverage fails the build; not individual pushes

cspotcode:


cspotcode:
@blakeembrey do you have any insight as to why the coverage reporting is behaving this way?  `coverage/coveralls` reports that coverage is good, but the individual `Coveralls - run-*` report that coverage decreased by zero. (??)  I'm not sure if there is a configuration option within coveralls that I don't have access to which might affect this.

blakeembrey:
> do you have any insight as to why the coverage reporting is behaving this way?

I don't, maybe a bug in the action? It seems like the webpage correctly shows "coverage didn't change" but maybe the action is working differently on zero.

PR #1043, opened 2020-05-16
Fix code coverage so it maps to TS sources

cspotcode:
Upgraded from istanbul to nyc, so this is blocked by dropping support for node 6.

The captured coverage is of the `npm install`ed instance of ts-node in ./tests/node_modules.  So I wrote a small remapper script to post-process the coverage report, rewriting the paths.


cspotcode:
Coveralls can visualize the coverage now.

image
image

PR #1054, opened 2020-05-23
fix: pass transformers object to ts.transpileModule method (Fixes #1051)

### LINKED ISSUE #1051: Transformers with transpileOnly
I have read the readme file and it does mention that the transformers are not supported with the `transpileOnly` mode.

I do see that the source code is using [ts.transpileModule](https://github.com/TypeStrong/ts-node/blob/master/src/index.ts#L491) which does support custom transformers. So my question is mainly to understand the reasoning behind not allowing transformers, when the underlying API of typescript allows it?

thetutlage:
As per the README, only the `transformers factory function` is not allowed during the `transpileOnly` module. However, an object can still be passed.

In this PR https://github.com/TypeStrong/ts-node/pull/879/files#diff-f41e9d04a45c83f3b6f6e630f10117feR383, the code already guards against the factory function, but doesn't pass the transformers to the `transpileModule` method.

Related issue https://github.com/TypeStrong/ts-node/issues/1051

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -488,10 +488,15 @@ export function create (rawOptions: CreateOptions = {}): Register {
   let getTypeInfo: (_code: string, _fileName: string, _position: number) => TypeInfo
 
   const getOutputTranspileOnly = (code: string, fileName: string, overrideCompilerOptions?: Partial<_ts.CompilerOptions>): SourceOutput => {
+    if (typeof transformers === 'function') {
+      throw new TypeError('Transformers function is unavailable in "--transpile-only"')
+    }
+
     const result = ts.transpileModule(code, {
       fileName,
       compilerOptions: overrideCompilerOptions ? { ...config.options, ...overrideCompilerOptions } : config.options,
-      reportDiagnostics: true
+      reportDiagnostics: true,
+      transformers: transformers
     })
 
     const diagnosticList = filterDiagnostics(result.diagnostics || [], ignoreDiagnostics)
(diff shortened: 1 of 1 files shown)


cspotcode:
@thetutlage Thanks for the speedy PR!

What is the reason for moving the assertion into `getOutputTranspileOnly`?  If I'm reading the change correctly, this change delays the `TypeError` so it's no longer thrown by `register()` or `create()` but is instead thrown by `require()` calls.

thetutlage:
The main reason for moving the `typeof transformers === 'function'` check inside the `getOutputTranspileOnly` method is to narrow down the `transformers` type.

Inside `getOutputTranspileOnly` the `transformers` variable is typed as follows

Screen Shot 2020-05-24 at 11 33 36 AM

Now, in order to narrow it down, we have two options

- Move the `if` condition inside this method (like I did)
- Manually type cast `transformers` to `Exclude` the factory function

What you think?

cspotcode:
Using a type assertion makes more sense.  We need to keep the assertion
where it was because that's the more intuitive place for a `TypeError` to
be thrown.  It makes the API better.

thetutlage:
If I get it right, you are saying to move the `typeof` check to its original place (where it was earlier) and instead go with 2nd option, ie: `Manually type cast transformers to Exclude the factory function`

Right?

cspotcode:
Correct.

thetutlage:
Does it look fine now?

cspotcode:
No reason.  Yeah, we should move it back.  I'll do that here and merge.

cspotcode:
Nevermind, I'm going to do it as a separate PR, cuz it'll be clearer that way.

thetutlage:
@cspotcode Any idea when this will be released?

cspotcode:
No set date; maybe this weekend?  Depends when someone has time; work is
keeping me busy this month.

In the meantime, you can install off of master (npm can install straight
from git) or downgrade to an older version.

PR #1082, opened 2020-07-23
docs: update IntelliJ (and WebStorm) desc

bluelovers:


cspotcode:
@bluelovers thanks, this is your plugin?  Can you tell us a bit more about it?

When people ask us questions about this plugin, where should we send them with questions?

bluelovers:
https://github.com/bluelovers/idea-run-typescript

image

image

image

Issue #1101, opened 2020-08-10
A new ESM loader equivalent of 'ts-node/register/transpile-only'


concision:
@cspotcode 

Continuing the discussion from [#1007](https://github.com/TypeStrong/ts-node/issues/1007#issuecomment-671155266).

<br>

CommonJS has the file shortcut `ts-node/register/transpile-only` for installing the TypeScript loader without type checking. An ESM loader which has the equivalent functionality should be created and exposed to the end user. 

https://github.com/TypeStrong/ts-node/blob/ce7c323557c15bf878bbc4dd5b2c0aac219fca4b/esm.mjs#L7

The code from [esm.mjs](https://github.com/TypeStrong/ts-node/blob/master/esm.mjs#L7) can be duplicated in a new file, but with the `transpileOnly` flag set:

```ts
export const {resolve, getFormat, transformSource} = esm.registerAndCreateEsmHooks({transpileOnly: true})
```

<br>

What should the loader filename look like? There is `ts-node/esm.mjs`, should it be `ts-node/esm/transpile-only.mjs`?

cspotcode:
`ts-node/esm/transpile-only.mjs` makes sense to me.  A user does `node --loader ts-node/esm/transpile-only ./foo.ts`

concision:
Resolved with #1102

PR #1102, opened 2020-08-10
Added a transpile-only ESM loader (#1101)

concision:
This resolves the feature request #1101: A new ESM loader equivalent of 'ts-node/register/transpile-only'

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -822,7 +822,7 @@ describe('ts-node', function () {
 
             expect(err.message).to.contain('Unable to compile TypeScript')
             expect(err.message).to.match(new RegExp('TS2345: Argument of type \'(?:number|1101)\' is not assignable to parameter of type \'string\'\\.'))
-            expect(err.message).to.match(new RegExp('TS2322: Type \'"(?:hello world|string)"\' is not assignable to type \'number\'\\.'))
+            expect(err.message).to.match(new RegExp('TS2322: Type \'(?:"hello world"|string)\' is not assignable to type \'number\'\\.'))
             expect(stdout).to.equal('')
 
             return done()
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -814,5 +814,30 @@ describe('ts-node', function () {
         return done()
       })
     })
+
+    describe('support a transpile only mode', () => {
+      it('should execute successfully with type errors and transpile-only enabled', (done) => {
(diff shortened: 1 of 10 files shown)


cspotcode:
Looks good but it'll need tests before being merged, and our package.json will need to be updated.

You can probably copy-paste an existing ESM test, but add a deliberate type error to make sure it gets ignored.

We also have tests that each entrypoint can be resolved, which will validate the changes to package.json.

concision:
I have added `esm/transpile-only.mjs` to the `package.json` exports and corresponding resolution tests.

I also added two tests to verify the loader should behave as expected:
- verify that the transpile-only ESM loader skips type checks in the presence of type errors
- verify that the transpile-only ESM loader throws a type error in the presence of type errors

Are there any other needed changes I need to make?

tests/esm-transpile-only/index.ts
```diff
@@ -0,0 +1,12 @@
+import { foo } from './foo.js'
+import { bar } from './bar.js'
+import { baz } from './baz.js'
+import { biff } from './biff.js'
```
cspotcode, line 4: Do you think we can remove the foo, bar, baz, and biff files?  Since `esm-transpile-only` is focused on verifying the transpile-only loader, so the focus is on the semantic errors.

src/index.spec.ts
```diff
@@ -802,6 +804,31 @@ describe('ts-node', function () {
           return done()
         })
       })
+
```
cspotcode, line 807: Nitpick: this suite name might imply that the ESM loader only supports transpile-only via this entrypoint, though it can also load this flag from environment variables or tsconfig.json.

Maybe name the 2x test cases to match some of our other specs' names, and remove the `describe()` wrapper?
```
it('should support transpile only mode via dedicated loader entrypoint', ...
it('should throw errors', ...
```

tests/esm-transpile-only/index.ts
concision: Sounds good; removed them.

src/index.spec.ts
concision: Is this acceptable?
https://github.com/TypeStrong/ts-node/blob/525b7308fc45355fa46f6306c36895e6de19c2fd/src/index.spec.ts#L808-L830

cspotcode: Yeah, looks good

cspotcode:
@concision looks great, thanks.  It's merged.  I need a review of #970 before making our next release, but hopefully we can get that done this weekend.  It'll be 9.0.0

PR #1103, opened 2020-08-10
Fix #1098: `ignored()` bug: ESM loader will try to compile .cjs, .mjs…

### LINKED ISSUE #1098: `ignored()` bug: ESM loader will try to compile .cjs, .mjs, and other unexpected file extensions
ESM loader uses our current `ignored()` implementation to determine if it should transform a file.  This only excludes the .js and .jsx file extensions as needed.  This worked fine with our CommonJS loader, because we knew we were only registered for the .ts, .tsx, and maybe .js and .jsx extensions.

The ESM loader is different, so it's asking `ignore()` about .mjs files.  (maybe other extensions too?)

We need to update the `ignored()` implementation to include rather than exclude file extensions so that we only transform the same files that `tsc` can transform.

https://github.com/TypeStrong/ts-node/blob/ce7c323557c15bf878bbc4dd5b2c0aac219fca4b/src/index.ts#L837-L845

concision:
Fixes #1098

### DIFF
--- a/src/esm.ts
+++ b/src/esm.ts
@@ -72,7 +72,7 @@ export function registerAndCreateEsmHooks (opts?: RegisterOptions) {
 
     // If file has .ts, .tsx, or .jsx extension, then ask node how it would treat this file if it were .js
     const ext = extname(nativePath)
-    if (ext === '.ts' || ext === '.tsx' || ext === '.jsx') {
+    if (ext !== '.js' && !tsNodeInstance.ignored(nativePath)) {
       return defer(formatUrl(pathToFileURL(nativePath + '.js')))
     }
 
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -728,8 +728,35 @@ describe('ts-node', function () {
     it('should create generic compiler instances', () => {
       const service = create({ compilerOptions: { target: 'es5' }, skipProject: true })
       const output = service.compile('const x = 10', 'test.ts')
-
       expect(output).to.contain('var x = 10;')
+    })
+  })
+
+  describe('issue #1098', () => {
+    function testIgnored (ignored: tsNodeTypes.Register['ignored'], allowed: string[], disallowed: string[]) {
+      for (const ext of allowed) {
+        expect(ignored(join(__dirname, `index${ext}`))).equal(false, `should accept ${ext} files`)
+      }
+      for (const ext of disallowed) {
+        expect(ignored(join(__dirname, `index${ext}`))).equal(true, `should ignore ${ext} files`)
+      }
+    }
+
+    it('correctly filters file extensions from the compiler when allowJs=false and jsx=false', () => {
+      const { ignored } = create({ compilerOptions: { }, skipProject: true })
+      testIgnored(ignored, ['.ts'], ['.js', '.tsx', '.jsx', '.mjs', '.cjs', '.xyz', ''])
+    })
+    it('correctly filters file extensions from the compiler when allowJs=true and jsx=false', () => {
+      const { ignored } = create({ compilerOptions: { allowJs: true }, skipProject: true })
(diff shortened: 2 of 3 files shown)


cspotcode:
Looks good; can we add a test for this?  We might be able to add to one of our existing ESM test cases.  We need to make sure that the test will fail if `ts-node` attempts to compile a `.cjs` or an `.mjs` file.

concision:
I have added a few white-box tests for `ignored()` to check that the correct extensions are included/excluded, based on `allowJs` and `jsx`. Notably, all combinations of `allowJs` and `jsx` should ignore other extensions such as `.cjs`, `.mjs`, `.xyz`, and even extension-less files. I am not sure if you had anything more elaborate in mind, but I think this covers tests for the scope of this issue.

<hr>

I did have a question regarding mixed capitalization extensions (e.g. `.tS`, `.Ts`, `.TS`). The TypeScript compiler seems to accept these mixed capitalization extensions, but ts-node does not. Should we support sending these unconventional extensions to the TypeScript compiler (i.e. when `ignored("filename.TS") === false`)?

This could be accomplished with a `.toLowerCase()` call on `ext` here (and in conjunction with the change in the next section):
https://github.com/TypeStrong/ts-node/blob/de289f679e50df94dadcbaca00539eaff09e4e17/src/index.ts#L837-L846

<hr>

I also think the ESM loader `getFormat()` function should be amended to use the ignore function for consistency:
https://github.com/TypeStrong/ts-node/blob/77f568ded01cae1e37bc36a95e9b8e1f6ca35a78/src/esm.ts#L73-L77
Instead:
```ts
    // If file is not ignored, then ask node how it would treat this file if it were .js
    if (!tsNodeInstance.ignored(nativePath)) {
      return defer(formatUrl(pathToFileURL(nativePath + '.js')))
    }
```
Although, I am not sure if there are any ramifications to this. Let me know what you think.

cspotcode:
Tests look great.

Should we add the `.d.ts` extension to the tests?  Thinking out loud, if node's module resolution points to a .d.ts file -- either in CommonJS or with our ESM hooks -- then something has gone wrong, but it's probably not our fault.  In that case, we do not want to ignore that file, because surely node cannot handle it.  If we handle the .d.ts file, then we'll be able to throw a meaningful error message.

Agreed about amending the `getFormat` hook; good catch.  Should we combine `ignored()` with the existing `ext ===` checks?  There's no need to do `nativePath + '.js'` if the file extension is already `.js`.

Regarding mixed capitalization extensions: I just tested, and our require hooks don't support mixed capitalization on Linux.  Seems like this has been a very long-standing limitation and we haven't received any bug reports about it.  I feel confident leaving it out of this pull request.  If you want, feel free to file a separate issue to investigate.  But I don't think it's a high priority.

concision:
Currently `ignored("xyz.d.ts") === false`, so `.d.ts` files will be blindly accepted and piped to the TypeScript compiler. I was curious what happens if I try to coerce the loader to run a `.d.ts` directly. Looks like we have been beaten by a few months already, as this was the error that was raised:
https://github.com/TypeStrong/ts-node/blob/77f568ded01cae1e37bc36a95e9b8e1f6ca35a78/src/index.ts#L629-L637
I think it is fine as is, this error will be automatically raised if _somehow_ a user manages to get a `.d.ts` file to be interpreted as a module. Good to go on this front, I think.

<br>

I have made the `getFormat` change to the following. I think this should be sufficient.
https://github.com/TypeStrong/ts-node/blob/afdf9649ad7aa799decec78a4cc01c2b62d941b1/src/esm.ts#L73-L77

<br>

I am mildly surprised that no issue has ever been raised about mixed capitalized extensions, but it definitely is not a priority or really that important. A nitpick, I suppose. Thanks for looking into it.

cspotcode:
Looks great.  I added .d.ts extensions to the tests to avoid any surprises there, and tests are still passing so I'll go ahead and merge this.

Thanks again @concision !

PR #1114, opened 2020-08-13
WIP REPL tab completion

cspotcode:
Took a quick stab at completions in the REPL.

Issues:

first time you get tab completions, the input line freezes while the language service spins up.  We'd probably want to force it to spin up before the first prompt is shown

TS tab completions sometimes replace already-entered text.  For example, they may propose rewriting `.f` into `?.foo`  Node's built-in completion doesn't support this.  Is it worth using a third-party library?  Does one exist with the features we need?

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -2,7 +2,7 @@
 
 import { join, resolve, dirname } from 'path'
 import { start, Recoverable } from 'repl'
-import { inspect } from 'util'
+import { inspect, callbackify } from 'util'
 import Module = require('module')
 import arg = require('arg')
 import { diffLines } from 'diff'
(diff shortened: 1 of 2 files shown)


blakeembrey:
This sounds really great! Is the main blocker right now the second issue? I assume the first one is workaround-able or we can put this behind a flag until we resolve it.

cspotcode:
Yeah, second issue is the blocker.  I don't remember off the top of my head, but I don't think there's any way for us to register an event listener when tab-completion "uses" a completion item.  If there was, we could maybe rewrite the input to replace `.` with `?.`

cspotcode:
I did some more research here.  Rewriting the line of input can be done like this:

```
replService.stdin.on('keypress', keypress => {
    if(keypress === '\t') {
      const completion = '?.bar';
      const completeOn = '.';
      const moveCursorToEnd = repl.cursor > repl.line.length - completeOn.length;
      repl.line = repl.line.slice(0, -completeOn.length) + completion;
      if(moveCursorToEnd) {
        repl.cursor = repl.line.length;
      }
      repl._refreshLine();
    }
  });
```

Implementation must:

- only do this for a TTY
- take care to update repl.cursor only if it is in or right of the affected part of the line
- Trigger on a known tab keypress, NOT trigger in `completer()` which is called by node even when the user does not want completions
- async completions?  Wait for completions to finish before applying?

Issue #983, opened 2020-03-16
Extending ts-node 


watzon:
### Desired Behavior

Allow complete control over the creation of a new REPL in code, including the ability to inject a new custom context.

### Is this request related to a problem?

There are many instances where one may want to extend the behavior of ts-node by creating their own REPL that wraps it and injects a custom context. One such case is creating a REPL for a framework such as AdonisJS, which is what I'm trying to do. The problem currently is, as far as I can tell, there is no way to inject a custom context outside of using the `--require` flag, but that just requires a file.

It would be nice to, in my own code, be able to do something like this:

```typescript
import { REPL, Context } from 'ts-node'

const myContext = new Context({
    myGlobal: require('./some-file')
})

const repl =  new REPL(myContext)
repl.run()
```

### Alternatives you've considered

Forking and modifying ts-node or a similar project to fit my needs.

cspotcode:
Seems reasonable to me.  This would mean refactoring `startRepl` to be a public API.  The guts of it would move from `bin.ts` to `index.ts`, probably wrapped us as a `createRepl` function.
 https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L327-L330

```
import { createRepl, register } from 'ts-node';
const service = register(... options ...);
const repl = createRepl({
    service,
    context: myContext
});
```

Alternatively, we expose *only* the eval implementation via a `createReplEval` function.
https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L350-L353

That way you fully control construction of the REPL, simply passing it a ts-node `eval` implementation.

Not something I have bandwidth to implement, but I'd personally accept a PR as long as it didn't impose additional burden on maintainers.  Seems like it wouldn't since it's not adding new features, just a tiny bit more public API surface.

watzon:
Glad you're open to the idea. If I get some free time I'll open a PR.

MarcManiez:
I would also have use for this in order to reproduce something like a rails console for the ts repo I currently work in. I'm right where y'all are in terms of bandwidth, so who knows whether I'll have the time to build this ¯\_(ツ)_/¯. At any rate, I wanted to say @watzon, you're not alone 🙂

PR #1121, opened 2020-09-03
Make ts-node eval public for node REPL consumption

MarcManiez:
Fixes https://github.com/TypeStrong/ts-node/issues/983

I took the second approach suggested by @cspotcode in https://github.com/TypeStrong/ts-node/issues/983:

> Alternatively, we expose _only_ the eval implementation via a `createReplEval` function.
https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L350-L353
> 
> That way you fully control construction of the REPL, simply passing it a ts-node `eval` implementation.

This necessitates moving quite a few things around. I've done my best to respect the conventions I could discern, but please let me know if I've missed something.


cspotcode:
@MarcManiez thanks for tackling this.  I hope to make a proper review today.  If for some reason I forget by Monday, feel free to ping me about it.

I appreciate the effort to respect existing conventions even when you had to move so much around.

Our existing API has a `create()` function that creates the ts-node service, then returns an object exposing everything about the service.  Should we take the same approach for the REPL?  Something like this?

```
import {create, createReplService} from 'ts-node';
const tsNodeService = create(/* options here */);
const tsNodeReplService = createReplService(tsNodeService);
const {eval} = tsNodeReplService;
```

Also, I don't know how @blakeembrey feels about this, but I'm in favor of moving the REPL functionality to a new file: `repl.ts`. 
 When concision sent us a few PRs recently, they mentioned that it was a bit tough to jump into the code with almost everything in `index.ts`.  If we can, splitting things out might help new contributors.

MarcManiez:
Thanks for taking a peep @cspotcode :)

I took a stab at what you suggested with `createReplService`: 9ab7eca. Did you see that including any properties other than `eval`?

I also moved the REPL code to its own file like you suggested. Happy to revert it if it's not to everyone's liking. From my newcomer's perspective, I agree with concision in that splitting the code into more files could help understand the code's structure a little better.

cspotcode:
Very cool, I'll take a look later.

Are you comfortable with me pushing changes to your branch when I review?
Sometimes there are small nitpicks I can fix, or I can propose changes that
we can always revert if there's disagreement.  I figure it's easier than
asking you to do all the work.

MarcManiez:
Absolutely no worries with that, thanks for checking in!

NickSeagull:
Hello! Any progress on this? I was about to make a fork in order to allow exactly this, and would like to use this feature in my current project. Would like to help on what it would be needed :)

JDvorak:
Looking forward to this! Thank you @MarcManiez

cspotcode:
No blockers, no update.  I've had a busy week with other priorities.

In the meantime, you can install ts-node directly off of a git branch using
npm's git support.  So you can start installing and using this feature
today.  And when we are able to publish and merge it (in the next couple
weeks I expect) you can switch from the git dependency back to the npm
package.

cspotcode:
I am *finally* getting around to reviewing this.  I'm not quite done yet.  Because of the refactoring, which I think is a good thing, it's taking me a tad longer to verify that I understand all the changes and that nothing has accidentally broken when merging upstream changes from master.

MarcManiez:
That's exciting, thank you for handling the merge 🙏 

I see codecov/patch is off the mark. If you have advice on how to best address that, I'm happy to help!

cspotcode:
I'm not sure why codecov/patch is off.  Hopefully it ... fixes itself by the time we merge?

I reordered the declarations in `repl.ts` to match their ordering from `bin.ts`, just because it make it easier for me to review.

I also wanted to see if I could reduce coupling between repl.ts and bin.ts.  So I expanded the API of `ReplService` and used the new API surface in `bin.ts`.  I realize you may have wanted to do this yourself and tried to keep your changes more minimal, which I definitely appreciate.

I'm not 100% done -- there are a couple TODOs in the code and I need to add tests for the new APIs -- but if you feel like taking a look, I welcome any review feedback.

cspotcode:
I think this is good enough to merge, pending a review.  Added a test and a `.md` file with some very brief docs.

I don't love the naming on `EvalAwarePartialHost`.

Some of the `@internal`s could be removed by moving those functions within the body of `createReplService`, but I'm ok being a bit lazy and deferring that work to a future PR unless anyone disagrees.

development-docs/repl-api.md
```diff
@@ -0,0 +1,27 @@
...
+- Create ts-node service using EvalState-aware `readFile` and `fileExists` implementations from REPL
+- Pass service to repl service
+- Either:
+  - call REPL method start() to start a REPL
+  - create your own node repl but pass it REPL service's eval() function
+
+
+```
+const tsNodeReplService = tsNode.createReplService()
+const {readFile, fileExists} = repl.getStateAwareHostFunctions()
```
Gerrit0, line 13: This looks off - should it be `tsNodeReplService.getStateAwareHostFunctions()`?

MarcManiez: Looks like `getStateAwareHostFunctions` was left over from the previous commit which renamed `getStateAwareHostFunctions` to `createEvalAwarePartialHost`. So maybe this was the intent?

```suggestion
const {readFile, fileExists} = tsNode.createEvalAwarePartialHost()
```

src/index.spec.ts
```diff
@@ -370,6 +372,34 @@ describe('ts-node', function () {
...
+    it('REPL can be created via API', async () => {
+      const stdin = new PassThrough()
+      const stdout = new PassThrough()
+      const stderr = new PassThrough()
+      const replService = createReplService({
+        stdin,
+        stdout,
+        stderr
+      })
+      const service = create(replService.evalStateAwareHostFunctions)
```
MarcManiez: As a newcomer, I experienced some confusion seeing the `replService` juxtaposed with the `service`. I noticed that the `Register` type applies to the service and that naming discrepancy was also a head-scratcher for me. With some digging, I kind of see what they represent.

Nothing immediately applicable, but just sharing some thoughts, as I recall y'all considering what might make the repo easier on newbies!

cspotcode: Yeah, I don't like the name of `Register`.  IMO it should be `Service` so that if you do `import {Service} from 'ts-node';`  We can rename it to `Service` and export by the legacy `Register` name for backwards-compatibility.

cspotcode:
Anybody know why the tests are failing on Windows?  Maybe related to #1126

src/index.spec.ts
cspotcode: Another question:
Given that our `create()` function returns a `Register`, which I'd like to rename to `Service`, does it make sense to rename `createReplService()` to `createRepl()` for consistency?  `create(): Service`, `createRepl(): ReplService`?

MarcManiez: Yes, I think that would make a lot more sense! I might even rename the `ReplService` type to `TsRepl`, or `TsNodeRepl`.

cspotcode: I'm thinking it's best to avoid putting `TsNode` into the names of exports, to match TypeScript and node's APIs.  For example:

* `import {Duplex} from 'streams';` (it's called Duplex, not DuplexStream)
* `import {LanguageService} from 'typescript';` It's named LanguageService instead of TsLanguageService, and lots of usages I've seen do a wildcard import to add the `ts` prefix: `import * as ts from 'typescript'; ts.LanguageService`

src/index.spec.ts
```diff
@@ -385,17 +385,15 @@ describe('ts-node', function () {
...
       replService.setService(service)
       replService.start()
-      stdin.write(
-        'const a = 123\n' +
-        '.type a\n'
-      )
+      stdin.write('\nconst a = 123\n.type a\n')
       stdin.end()
-      await promisify(setTimeout)(100)
+      await promisify(setTimeout)(1e3)
```
MarcManiez, line 390: 🤯

PR #1122, opened 2020-09-10
fix: Alias --es-module-specifier-resolution to --experimental-specifier-resolution for Node 12 compatibility

nguyensomniac:
Context here: https://github.com/TypeStrong/ts-node/issues/1007#issuecomment-689860997

The `--es-module-specifier-resolution` flag was introduced [in Node 12](https://medium.com/@nodejs/announcing-a-new-experimental-modules-1be8d2d6c2ff). With the release of Node 13, the flag was renamed to `--experimental-specifier-resolution`, with `--es-module-specifier-resolution` [acting as an alias to the new name](https://github.com/nodejs/node/issues/30520#issuecomment-588288220).

This fix aliases the `--es-module-specifier-resolution` flag in older versions of node, so that module resolution acts as expected. I also changed what seems to be a dead branch in the `esm-usage-example` directory.

I tested this with node 12, 13, and 14 and all seems to be working as expected.

### DIFF
--- a/dist-raw/node-options.js
+++ b/dist-raw/node-options.js
@@ -25,7 +25,8 @@ function parseArgv(argv) {
     '--preserve-symlinks': Boolean,
     '--preserve-symlinks-main': Boolean,
     '--input-type': String,
-    '--experimental-specifier-resolution': String
+    '--experimental-specifier-resolution': String,
+    '--es-module-specifier-resolution': '--experimental-specifier-resolution',
   }, {
     argv,
     permissive: true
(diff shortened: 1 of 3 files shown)


cspotcode:
@nguyensomniac thanks, seems like a simple change, and thanks for adding a test.

Node's docs do not mention the older flag name: `--es-module-specifier-resolution`
https://nodejs.org/docs/latest-v12.x/api/esm.html#esm_resolution_algorithm

Was the flag renamed during node 12's lifetime?

ExE-Boss:
> Was the flag renamed during node 12's lifetime?

Yes, it was.

Probably even long‑ish before **v12** became **LTS**.

nguyensomniac:
Ah, you're correct. Running `node --help` looks like it was renamed after 12.15

cspotcode:
Thanks!  I added a comment to remind me when this flag was changed.

This will be published to npm after I merge #1121; until then, you can install directly from git.  I always forget the syntax, but I think it's `npm install github:TypeStrong/ts-node#master`

PR #1149, opened 2020-11-09
Remove coveralls reporting

cspotcode:
Before merging, I need to double-check with @blakeembrey to be sure no one else is using the coveralls reporting.


blakeembrey (approved):
Seems good, do you know why everything failed in the tests though?

cspotcode:
Good catch, seems like I accidentally left a couple lines in the GH Actions workflow that were trying to run an npm script which gets removed by this PR.  Removes those lines, I'll double-check that it works.

PR #1153, opened 2020-11-12
Add nested package.json resolution to esm resolver

mlegenhausen:
This PR adds an additional resolution strategy when resolving the `index` file if instead an `package.json` file is used to redirect to a `main` or `module` entry.

This is needed when you want to import e. g. `fp-ts` which defines for each module a `lib` (commonjs) and `es6` (esm) file and uses a `package.json` that only contains a `main` and `module` definition.

### DIFF
--- a/raw/node-esm-resolve-implementation.js
+++ b/raw/node-esm-resolve-implementation.js
@@ -231,7 +231,12 @@ function resolveExtensions(search) {
 }
 
 function resolveIndex(search) {
-  return resolveExtensions(new URL('index', search));
+  const index = resolveExtensions(new URL('index', search));
+  if (index !== undefined) return index;
+
+  const pkg = resolveExtensions(new URL('package', search));
+  const config = getPackageConfig(pkg, search);
+  return packageMainResolve(pkg, config, search);
 }
 
 function finalizeResolution(resolved, base) {


cspotcode:
Thanks, I have a few quick questions, and I think this will need some changes to actually work.

Our `node-esm-resolve-implementation.js` is copy-pasted out of node's source code, then modified as necessary.  The README here explains why.  Just to be sure, did you take a look at that directory?  It looks like you've modified `raw/` but left `dist/` unchanged.  `raw/` contains copies of the unmodified node source code to make it easy to identify our changes via diff, but it's not actually used by `ts-node`.

We need to match node's resolution behavior, with .ts resolution added, but otherwise not deviate.  Do these changes keep our behaviour consistent with node 15?

mlegenhausen:
Ah ok, thought it was a ts-node related issue. With that hint I was able to find an open issue in the nodejs issue tracker that describes exactly the problem: https://github.com/nodejs/node/issues/32103#issuecomment-596873109

I think it is the best to solve it there and then migrate the change.

cspotcode:
Ok excellent, thanks for sharing your research and a link that that bug.
I'll keep an eye on it.

PR #1158, opened 2020-11-22
Rename `Register` to `Service`; re-export as `Register` for backwards…

cspotcode:
Based on conversation here: https://github.com/TypeStrong/ts-node/pull/1121#discussion_r528259816

`create()` will create a `Service`.  `register()` will create and register it into the node runtime.  But the object itself is not really a `Register` (noun), it's a `Service` which handles compilation.

Other names could be `Compiler`, `TsNodeCompiler`, or `Compiler`?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1031,15 +1031,15 @@ function registerExtensions (
  */
 function registerExtension (
   ext: string,
-  register: Service,
+  service: Service,
   originalHandler: (m: NodeModule, filename: string) => any
 ) {
   const old = require.extensions[ext] || originalHandler // tslint:disable-line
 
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -1024,7 +1031,7 @@ function registerExtensions (
  */
 function registerExtension (
   ext: string,
-  register: Register,
+  register: Service,
```
blakeembrey, line 1034: Should we rename this variable too?

blakeembrey (approved):
Awesome, thanks!

src/index.ts
cspotcode: Yeah, good catch.

Issue #779, opened 2019-02-11
Swc support for faster transpilation


cspotcode:
I just wanted to drop this on the radar in case you hadn't seen it.  `swc` is a transpiler written in Rust that's an order of magnitude faster than those written in JS.
https://swc-project.github.io

On the one hand it seems out of scope for ts-node.

On the other hand, people turn to ts-node when they want to run TS code without pre-compilation.  It handles all the details: sourcemaps work, your tsconfig is respected, etc.  If ts-node had a `--useSwc` mode that made it an order of magnitude faster (with caveats) I think people would find that very useful.

I imagine that the hypothetical `--useSwc` flag is invalid unless `--transpileOnly` is also specified.  Or `--useSwc` implies `--transpileOnly`.  swc can't perform typechecking.

blakeembrey:
That does look useful. I also see https://github.com/swc-project/swc/issues/126. I wonder if it makes sense to have something more generic than `ts-node` for this sort of thing. E.g. `node -r swc/register`. With the way things are moving, it might be preferable to use a light-weight transpiler at runtime and compile time, and only use `tsc` for type checking with `noEmit` enabled.

cspotcode:
I'm very skeptical that swc will ever be able to perform typechecking.
Just feels way too complex an undertaking.

I agree that a very lightweight compiler for dev and runtime is awesome.
tsc --noEmit can be moved to the test suite.  Shameless plug: I threw
together tsc-to the other day.  It's like tsc but skips typechecking.

I think the missing piece for TS developers is: automatically generating
swc configuration from our tsconfig, e.g. grabbing the correct react pragma
and approximating the "target" flag.  Whether that lives within ts-node or
swc, I'm not sure.  But it'd be really cool if a TS dev didn't have to
think much about it.

I don't think swc's compilation flags map one-to-one with tsconfig.
Decorators won't work the same.  Swc offers more granular control than TS's
"target"

cspotcode:
I did a bit more poking around.  swc isn't ready for primetime; I filed a few issues: https://github.com/swc-project/swc/issues/created_by/cspotcode

Hopefully it will be soon!

whitecolor:
@blakeembrey 
In TS 3.4 they also added [incremental flag](https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/#faster-subsequent-builds-with-the---incremental-flag) and option, to speed up sequential builds, no sure if it affects ts-node buids?

[...]

PR #1160, opened 2020-11-22
swc transpiler and new --transpiler option to use third-party transpilers

cspotcode:
Implements #779

Usage information is at the top of this description.  Below are design questions and development to-dos.

## Usage

```jsonc
// tsconfig.json
{
  "ts-node": {
    "transpiler": "ts-node/transpilers/swc-experimental",
    "transpileOnly": true
  }
}
```

```bash
# Can substitute @swc/core to be even faster if you're able to use native modules
npm install --save-dev typescript @swc/wasm github:TypeStrong/ts-node#ab/swc-compiler

# should be very fast via swc
ts-node ./index.ts
```

---

## Implementation notes, design questions

~~This is incomplete.  To be a proper drop-in compiler I think it will need to implement additional parts of the `TsCommon` interface.  It can pull these from `require('typescript')`~~ <--this much has been implemented.

But ideally our `compiler` option supports composable compilers, so you can specify `"compiler": ["ntypescript", "ts-node/compilers/swc"]` and the swc wrapper will defer to `ntypescript` for things it does not implement itself.  This has not been implemented yet.

Still, it might be cool for users to specify `"transpileOnly": true, "compiler": "ts-node/compilers/swc"` and get a performance boost.

## Other compilers: (note they are only worth considering if they're noticeably faster than `ts.transpileModule`)
- https://github.com/alangpierce/sucrase
- https://github.com/evanw/esbuild
  - https://esbuild.github.io/api/#transform-api
  - https://github.com/evanw/esbuild/issues/590

## Configuration with options: bike-shedding

How should we allow passing configuration options to the transpiler?

```
"transpiler": "ts-node/transpilers/swc-experimental",
"transpilerOptions": {
  "swc": "@swc/wasm"
}
```

```
"transpiler": {
  "name": "ts-node/transpilers/swc-experimental",
  "swc": "@swc/wasm"
}
```

```
"transpiler": ["name", { "swc": "@swc/wasm"}]
```

## TODO
- [x] use peerDependenciesMeta
https://yarnpkg.com/configuration/manifest/#peerDependenciesMeta
https://blog.npmjs.org/post/186983646370/npm-cli-roadmap-summer-2019
- [x] expose as factory which accepts options, so it can be customized (use `@swc/wasm` vs `@swc/core`)
- [x] update ts-node logic to invoke factory when detected
- [x] warn when `transpileOnly` turned off, since you won't get the speed benefit of `swc`?
- [x] ~~warn when `createTypescriptCompiler` is misspelled by detecting when compiler instance doesn't have well-known compiler APIs such as `transpileModule` or `createProgram`~~ Instead, I've switched to a proper transpiler API that does not need to mimic TS compiler API
- [ ] investigate weird WASM shutdown slowness: when using @swc/wasm, there is a delay in node process shutdown once code has finished running.  Does not happen with @swc/core
- [ ] Compare against https://github.com/Brooooooklyn/swc-node/blob/master/packages/register/register.ts#L9; see if there are any lessons to be learned or additional work for us to achieve feature parity
- [x] remove TODOs from comments


src/compilers/swc.ts
```diff
@@ -0,0 +1,95 @@
...
+targetMapping.set(/* ts.ScriptTarget.ES3 */ 0, 'es3')
+targetMapping.set(/* ts.ScriptTarget.ES5 */ 1, 'es5')
+targetMapping.set(/* ts.ScriptTarget.ES2015 */ 2, 'es2015')
+targetMapping.set(/* ts.ScriptTarget.ES2016 */ 3, 'es2016')
+targetMapping.set(/* ts.ScriptTarget.ES2017 */ 4, 'es2017')
+targetMapping.set(/* ts.ScriptTarget.ES2018 */ 5, 'es2018')
+targetMapping.set(/* ts.ScriptTarget.ES2019 */ 6, 'es2019')
+targetMapping.set(/* ts.ScriptTarget.ES2020 */ 7, 'es2019')
+targetMapping.set(/* ts.ScriptTarget.ESNext */ 99, 'es2019')
+targetMapping.set(/* ts.ScriptTarget.Latest */ 99, 'es2019')
```
Gerrit0, line 95: 99 is set twice - intentional?

Gerrit0 (commented):
From someone fairly unfamiliar with ts-node's architecture, looks good! As a user, I would definitely want a warning for ts-node failing to use the specified compiler.

cspotcode:
Thinking about this, I see some problems:

* With the current design, I'm not sure there's a clean way to warn the user when they use swc without transpileOnly, but I agree we should have such a warning.

* I was hoping to avoid adding complexity to `ts-node`'s core by doing this via the existing `--compiler` option , but then I ended up changing the compiler loader anyway to support factories.

* It's unfortunate that every call to `transpileModule` needs to re-create the swc options.

Given the above problems, perhaps it is better to introduce a proper `--transformer` option, similar to `--compiler`.  When this option is used, it provides the `--transpileModule` implementation and throws an error when used without `--transpileOnly`.  This also allows us to enforce a very clean `create()` factory API, so compiler options can be parsed and established only once, and then all subsequent calls simply invoke the transformer.

`--transformer`?  `--transpiler`?  `--transpileModuleProvider`?  `--transpileOnlyCompiler`?  Not sure what to call it.

klesun:
It shortens `ts-node` compilation delay about 3-6 times, thanks so much for this PR!

QuentinFarizon:
When using yarn one should use `yarn add --dev https://github.com/TypeStrong/ts-node#main typescript @swc/core @swc/helpers`

Thanks for the 3x launch time improvement !

PR #1165, opened 2020-11-26
Fix: Keep search in resolveReplacementExtensions

frandiox:
Related to https://github.com/TypeStrong/ts-node/issues/1007#issuecomment-733894228

The search params were lost after making a partial copy to change the file extension.

Even though the `dist-raw` is largely a copy of Node's implementation, I believe this specific line is custom in `ts-node`.

@cspotcode Thanks for all the pointers you wrote in the comment!

### DIFF
--- a/dist-raw/node-esm-resolve-implementation.js
+++ b/dist-raw/node-esm-resolve-implementation.js
@@ -307,7 +307,8 @@ function resolveReplacementExtensions(search) {
     const pathnameWithoutExtension = search.pathname.slice(0, search.pathname.length - 3);
     for (let i = 0; i < replacementExtensions.length; i++) {
       const extension = replacementExtensions[i];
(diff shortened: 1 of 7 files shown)


cspotcode:
Thanks for sending this!  It's a holiday for me so it'll be at least a few days before I review this.  When I do, I'll double check that the modified line is a ts-node modification.  Should be easy with a diffing tool.

Also, ideally we'd have a test for this, too.  I think we can add your original reproduction example to our existing ESM tests.  Import the same file twice with different params and have it log a message each time.

frandiox:
@cspotcode I've added a test for this. If you revert my first commit, it will fail.
Enjoy your holidays! 🎉

cspotcode:
Looks good, thanks again.  I'm gonna merge this, then take this as an excuse to update our copy-pasted code to the latest from node 15 in a followup PR.  Then I'll publish to npm.

PR #1179, opened 2020-12-16
Clarify README for tsx/jsx endings.

### LINKED ISSUE #1178: SyntaxError with tsx file ending.
### Expected Behavior

According to the readme, ts-node registers itself for file endings `.tsx?`, i.e. `.ts` and `.tsx`. Therefore the behavior should be the same for both file endings.

### Actual Behavior

`SyntaxError: Cannot use import statement outside a module` with.tsx file

```sh
# create a test.ts file
$ echo "import * as path from 'path'" >> test.ts
$ echo "console.log(Object.keys(path))" >> test.ts

# works as expected
$ ts-node test.ts
[
  'resolve',    'normalize',
  'isAbsolute', 'join',
  'relative',   'toNamespacedPath',
  'dirname',    'basename',
  'extname',    'format',
  'parse',      'sep',
  'delimiter',  'win32',
  'posix',      '_makeLong'
]

# copy to test.tsx
$ cp test.ts test.tsx

# try running it with ts-node
$ ts-node test.tsx
(node:19628) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
C:\Users\aval\tsdev\chroma.ts\ts-node-error\test.tsx:1
import * as path from 'path'
^^^^^^

SyntaxError: Cannot use import statement outside a module
...

# -O does nothing
$ ts-node -O "{\"module\":\"commonjs\"}" test.tsx
(node:1784) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
C:\Users\aval\tsdev\chroma.ts\ts-node-error\test.tsx:1
import * as path from 'path'
^^^^^^

SyntaxError: Cannot use import statement outside a module
...


```

### Minimal reproduction

see above

### Specifications

* ts-node v9.1.1
* node v14.15.2
* compiler v4.1.3

* tsconfig.json, if you're using one: none, see above
* 
image

* If Windows, are you using WSL or WSL2?: no

NaridaL:
Fix #1178

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,9 @@ Create a new Node.js configuration and add `-r ts-node/register` to "Node parame
 
 ## How It Works
 
-**TypeScript Node** works by registering the TypeScript compiler for `.ts` extensions (and `.js` when `allowJs == true`, and `.tsx`/`.jsx` when [`jsx`](https://www.typescriptlang.org/docs/handbook/jsx.html) is set). When node.js has an extension registered (via `require.extensions`), it will use the extension internally for module resolution. When an extension is unknown to node.js, it handles the file as `.js` (JavaScript). By default, **TypeScript Node** avoids compiling files in `/node_modules/` for three reasons:
+**TypeScript Node** works by registering the TypeScript compiler for `.ts`, `.tsx`, `.js`, and `.jsx` extensions.
+`.js` and `.jsx` are only registered when allowJs is true.  `.tsx` and `.jsx` are only registered when [jsx](https://www.typescriptlang.org/docs/handbook/jsx.html) is enabled.
+When node.js has an extension registered (via `require.extensions`), it will use the extension internally for module resolution. When an extension is unknown to node.js, it handles the file as `.js` (JavaScript). By default, **TypeScript Node** avoids compiling files in `/node_modules/` for three reasons:
 
 1. Modules should always be published in a format node.js can consume
 2. Transpiling the entire dependency tree will make your project slower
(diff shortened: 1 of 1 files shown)


NaridaL:
Fixes #1178

NaridaL:
This isn't entirely unambiguous, as you could read it as both tsx and jsx are enabled if you set `jsx`, but I didn't really see a way of improving that short of introducing a table.

cspotcode:
Good point; how do you feel about this wording?  Too verbose?  Still ambiguous?

```markdown
**TypeScript Node** works by registering the TypeScript compiler for `.ts`, `.tsx`, `.js`, and `.jsx` extensions.  `.js` and `.jsx` are only registered when allowJs is true.  `.tsx` and `.jsx` are only registered when jsx is enabled.  When node.js has an extension registered (via `require.extensions`), it will use the extension internally for module resolution. When an extension is unknown to node.js, it handles the file as `.js` (JavaScript). By default, **TypeScript Node** avoids compiling files in `/node_modules/` for three reasons:
```

NaridaL:
No, that's better 👍

NaridaL:
I changed it. I left the link in to make it obvious that jsx refers to the compiler option in this context.

cspotcode:
Thanks, I took your lead and also added the best link I could find for allowJs.  If it looks good to you, I'll go ahead and merge.

NaridaL:
Well we both tried to quote jsx at the same time. I guess you were faster 😋 

LGTG

cspotcode:
Perfect, thank you!

Issue #903, opened 2019-11-07
TypeScript 3.7 Support


lukeshiru:
I have TypeScript 3.7 installed on a project, and I run tests with ava and ts-node. The issue I found is that compilation works as expected, but tests fail because ts-node doesn't recognize the syntax for nullish coalescence and optional chaining.

So if I have:
```typescript
foobar?.foo?.bar ?? "default value"
```

I get errors all over the place. Isn't `ts-node` using the local typescript package to do compilation?

keesvanlierop:
It does, but it is set to 3.6.4 still. Needs to be updated and then good to go I think?

blakeembrey:
TypeScript 3.7 is already supported, you would need to update your local installation.

keesvanlierop:
It's not supported though. As OP mentions, you'll see errors on optional chaining and nullish coalescing expressinons when running ts-node.

lukeshiru:
One of the errors I'm getting:
```
Uncaught exception in src/tryCatch/tryCatch.test.ts

  ./node_modules/ts-node/src/index.ts:493

  ./src/tryCatch/tryCatch.ts:16
          returnedValue = catchCallback?.(error);
                                        ^

  SyntaxError: Unexpected token '.'

  Module.m._compile (node_modules/ts-node/src/index.ts:493:23)
  module.exports (node_modules/default-require-extensions/js.js:7:9)
  require.extensions.<computed> (node_modules/ts-node/src/index.ts:496:12)
```
Tried a clean install with `npm`, still getting errors.

keesvanlierop:
I noticed the fix is to change `"target": "esnext"` to `"target": "es2018"`. No change required here.

blakeembrey:
By that error it is clearly workin, as you noted it’s failing because you’re compiling an output that node.js doesn’t understand yet.

lukeshiru:
Ohhhh yup. It seems to be a target issue. Thanks @keesvanlierop 😄

iwasaki-kenta:
Is there any way to have nullish coalescing work on target `esnext`? The use case is that bigint's are only supported should `esnext` be targeted, rather than `es2018`.

lukeshiru:
@iwasaki-kenta my solution was to set target to es2018 and then: 
```json
"lib": [
      "es2018",
      "esnext.bigint"
    ]
```
So I have 2018 stuff, but I still support BigInt. It will only work expressed as `BigInt("1")` not like `1n` 😞

[...]

PR #1201, opened 2021-01-28
docs(README.md): add troubleshooting tip for syntax errors

jedwards1211:
as discussed in #903.  I had to search for awhile to find that issue and figure out what was going on, and there were some duplicate issues, so I think it would help people to explain this directly in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,9 +221,18 @@ Most options can be specified by a `"ts-node"` object in `tsconfig.json` using t
 
 Our bundled [JSON schema](https://unpkg.com/browse/ts-node@8.8.2/tsconfig.schema.json) lists all compatible options.
 
-## SyntaxError
+## Help! I'm getting a `SyntaxError` in `ts-node`, but my code compiles with `tsc`!
 
-Any error that is not a `TSError` is from node.js (e.g. `SyntaxError`), and cannot be fixed by TypeScript or `ts-node`. These are runtime issues with your code.
+This happens when your `compileOptions.target` is too high, so TypeScript leaves some fancy new syntax as-is, instead of transpiling it into a form your version of Node understands.
+
+For example, if you're using the `?.` optional chaining operator with `"target": "esnext"` or `"target": "es2020"`, `tsc` won't transpile it, 
+and Node 12, which doesn't support `?.`, will throw a `SyntaxError`.  If you lower `"target"` to `"es2019"`, `tsc` will transpile the `?.` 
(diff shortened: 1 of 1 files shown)


cspotcode:
Nice, thanks for writing this up.

We already have this section explaining `SyntaxError`: https://github.com/TypeStrong/ts-node#syntaxerror
Should your explanation be merged into it?

Should we also explain how to identify the difference between a TypeScript syntax error and a node syntax error?  They have different names, so once you know where to look, it's easy and reliable to tell them apart.
I think this may be addressed by merging your additions into the existing `SyntaxError` section.  Perhaps renaming it will also help, something like `SyntaxError vs TSError`?

jedwards1211:
I don't think `SyntaxError vs TSError` would be obvious enough, I had completely failed to notice the `SyntaxError` section because the name of the errors was the last thing on my mind when I stumbled on this issue.  I was thinking "my code compiles with `tsc`, so why does it not work with `ts-node`?"  I thought `ts-node` was somehow invoking an old version of `tsc` or using the wrong compile options.  So I think some version of "my code compiles with `tsc`, so why does it not work with `ts-node`?" is more likely for people to find.

And yeah I think it would be good to merge in what I wrote because I'm not sure the existing explanation there would have been enough to make the light go off in my head.  "cannot be fixed by TypeScript or ts-node. These are runtime issues with your code" sounds like it's saying the mistake is in your code itself, not the transpilation settings.

jedwards1211:
If nothing else adding `Help!` to the title would really help, e.g. `Help! I'm getting a SyntaxError!`, because the `Help!` catches your eye immediately when you're looking for troubleshooting info.

jedwards1211:
Okay I merged my explanation into the syntax error section, see what you think

cspotcode:
Thanks, I haven't forgotten about this.  I'm also working on creating a documentation website in #1207

I may roll these changes into a dedicated website page about errors and how to interpret them.  Not sure yet; it's all a work-in-progress.
https://typestrong.org/ts-node/docs/interpreting-errors

cspotcode:
I ended up adding this to the new "Troubleshooting" page: https://typestrong.org/ts-node/docs/troubleshooting#unsupported-javascript-syntax

I'm going to close this to keep the issue tracker tidy, but please feel free to suggest edits or improvements.  It's been a while since I've looked at this so I may have neglected something obvious, even if we already discussed it.  Thanks again for sending this PR.

Issue #1099, opened 2020-08-10
Docusaurus site to break up the README?


cspotcode:
@blakeembrey how do you feel about having a Docusaurus site for ts-node, letting us break up the README into multiple pages?  Motivation is to improve the appeal and discoverability of the information in our README, since it's a bit long.

I've used Docusaurus at work, and it's the simplest way I know to render markdown into an attractive docs site.  You create `docs/*.md` files and each is rendered as a page.  It'll be easy to setup on Github Actions.  It could live at https://TypeStrong.github.io/ts-node

On the other hand, I don't like that the npmjs.com README will be different than the documentation site.  We'd either have to make the readme empty, or somehow duplicate high-profile documentation between the docs site and the README.  Or some smart copy-pasting script to keep them in sync?  None of those ideas sound good.

Alternatively, maybe we add a table of contents to the README with http://npm.im/markdown-toc?

cspotcode:
Initial implementation is merged; will keep iterating on the pages till it's ready to be "official."

PR #1207, opened 2021-02-01
Docusaurus-powered documentation website

cspotcode:
https://typestrong.org/ts-node

Implements #1099 

See also: https://github.com/TypeStrong/discussions/issues/6

What if we merge this now so that the website gets published automatically, then iteratively improve the docs over time?

In other words, the website will be live but incomplete, and we won't link to it until it's ready.


cspotcode:
The sidebar navigation, table of contents, and "prettiness" of the site may encourage people to read the documentation, more easily find the answers they need.  Plus I think it frees us up to create dedicated pages with tsconfig samples, examples, stuff like that.

But for npm we still want to have a README.  So maybe there's some easy, hacky way to build up the README from the various docs pages, essentially copy-pasting them together via a build script.  Then we can write docs for the docsite and still get a meaningful README published to npm.

cspotcode:
Added typedoc API docs since it was so easy to do.  May prove helpful to link people to an HTML rendition of our programmatic options.

Added a "under construction" message to the first page.

Marked as "ready for review" because I think this can safely be merged today, then we can improve the pages over time until we deem the site ready to recommend to new users.

cspotcode:
Merged to master, and I'll track future website & docs work in follow-up issues and PRs.

PR #1227, opened 2021-02-16
Refactor tests to stop using callbacks

cspotcode:


src/index.spec.ts
```diff
@@ -15,17 +15,23 @@ import Module = require('module')
...
+type TestExecReturn = { stdout: string, stderr: string, err: null | ExecException }
+function exec (cmd: string, opts: ExecOptions = {}): Promise<TestExecReturn> & { child: ChildProcess } {
+  let childProcess!: ChildProcess
+  return Object.assign(
+    new Promise<TestExecReturn>((resolve, reject) => {
+      childProcess = childProcessExec(cmd, {
+        cwd: TEST_DIR,
+        ...opts
+      }, (error, stdout, stderr) => {
+        resolve({ err: error, stdout, stderr })
```
blakeembrey, line 27: Nit: It's probably more ergonomic to use as `err ? reject(err) : resolve({ stdout, stderr })`.

cspotcode: Yeah, that's fair.  There are a few tests that assert that an error *does* happen, and I wasn't sure how it would look to catch the error and then assert against it.
```
let err; try {await exec()} catch(e) {err = e} assert(err)
```

blakeembrey: Makes sense. I mostly use jest now and it has a `.rejects` you can use, forgot what's included with mocha.

cspotcode: I want to switch ts-node to a test runner that supports parallel tests.  I'm thinking of switching to ava, and I'm worried that jest installs too much magic into the runtime environment to enable testing ts-node, since we install into `require.extensions` and install source-map-support.  What do you think?

PR #1231, opened 2021-02-21
Remove tslint, add prettier, add semicolons

### LINKED ISSUE #1211: Switch to prettier; drop tslint
I don't want to spend much effort on this, but with tslint being super-deprecated, it'd be nice to drop it.

IMO this should happen between feature work, so wait till after all in-progress PRs are merged to avoid merge conflicts.

Not even sure how much value we'll get from eslint if we have prettier.  I don't get the sense that contributors to this project will make the type of mistakes that eslint catches.

#1169, #1170

cspotcode:
Fixes #1211 

As best I can tell, using tabs is the better choice for accessibility, so it's not an opinion or style choice.

Using semicolons is prettier's default.

Editorconfig will tell github to render tabs as 2 spaces, so code in the GH UI will still look the same.

Disabled tabs in dist-raw since that stuff should match whatever nodejs/node is doing.

## TODO

* [ ] remove old `// tslint:disable` comments

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
     "tsconfig.schemastore-schema.json"
   ],
   "scripts": {
-    "lint": "tslint \"src/**/*.ts\" --project tsconfig.json",
-    "lint-fix": "tslint \"src/**/*.ts\" --project tsconfig.json --fix",
+    "lint": "prettier --check .",
+    "lint-fix": "prettier --write .",
     "clean": "rimraf dist && rimraf tsconfig.schema.json && rimraf tsconfig.schemastore-schema.json && rimraf tests/ts-node-packed.tgz",
     "build": "npm run build-nopack && npm run build-pack",
     "build-nopack": "npm run clean && npm run build-tsc && npm run build-configSchema",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -128,5 +127,11 @@
     "make-error": "^1.1.1",
     "source-map-support": "^0.5.17",
     "yn": "3.1.1"
+  },
+  "prettier": {
+    "singleQuote": true,
+    "trailingComma": "all",
+    "printWidth": 120,
+		"useTabs": true
```
blakeembrey, line 135: Why not just use prettier with defaults completely?

cspotcode: I'm not overly attached to any choices; happy to change them.

tabs seem like the better choice for accessibility by the visually impaired.  At least as far as I can read online (https://github.com/prettier/prettier/issues/7475, https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/)

singleQuote is to match existing style but it's not a big deal

No preference on printWidth.  I had assumed without evidence that the default would cause a lot of lines to be wrapped.  I tested and very few were actually affected

cspotcode:
Rolled back formatting changes; pulled in latest changes from master; renamed PR title to match.

PR #1292, opened 2021-04-07
Adding "extends": support to `'ts-node'` loading from `tsconfig.json`

seivan:
So that one `tsconfig` extending another will merge the `'ts-node'` options specified in both.

`readTsConfigExtendForTsNodeRecursively(cwd: string, ts, rawConfig)` will read as far as it can and merge with previous config. 
Each `extend` has lower priority than previous, so your original  `"ts-node"` values will not be overwritten by the next `extend` and so forth. 

Based on comment https://github.com/TypeStrong/ts-node/issues/1007#issuecomment-812923417

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -41,6 +41,36 @@ function fixConfig(ts: TSCommon, config: _ts.ParsedCommandLine) {
   return config;
 }
 
+function readTsConfigExtendForTsNodeRecursively(
+  cwd: string,
+  ts: TSCommon,
+  rawConfig: { [key: string]: any } = {}
+): TsConfigOptions {
+  const configKey = 'ts-node';
+  const recurse = (
+    extending: { [key: string]: any } = {},
+    original: { [key: string]: any } = {}
+  ) => {
+    return readTsConfigExtendForTsNodeRecursively(cwd, ts, {
+      [configKey]: { ...extending[configKey], ...original[configKey] },
+    });
+  };
+
+  const { extends: extendFile, ...config } = rawConfig;
+
+  const result =
+    extendFile === undefined
+      ? filterRecognizedTsConfigTsNodeOptions(config[configKey])
+      : [extendFile]
+          .filter((x) => x && typeof x === 'string')
+          .map((x) => x as string)
+          .map((extended) => resolve(cwd, extended))
+          .map((pathString) => ts.readConfigFile(pathString, ts.sys.readFile))
+          .map((readFile) => readFile.config)
+          .map((extendedConfig) => recurse(extendedConfig, config))[0];
+
+  return result ?? {};
+}
 /**
  * Load TypeScript configuration. Returns the parsed TypeScript config and
  * any `ts-node` options specified in the config file.
@@ -105,11 +135,11 @@ export function readConfig(
   }
 
   // Fix ts-node options that come from tsconfig.json
-  const tsNodeOptionsFromTsconfig: TsConfigOptions = Object.assign(
-    {},
-    filterRecognizedTsConfigTsNodeOptions(config['ts-node'])
+  const tsNodeOptionsFromTsconfig: TsConfigOptions = readTsConfigExtendForTsNodeRecursively(
+    cwd,
+    ts,
+    config
   );
-
   // Remove resolution of "files".
   const files =
     rawApiOptions.files ?? tsNodeOptionsFromTsconfig.files ?? DEFAULTS.files;
@@ -179,8 +209,8 @@ export function readConfig(
  */
 function filterRecognizedTsConfigTsNodeOptions(
   jsonObject: any
-): TsConfigOptions {
-  if (jsonObject == null) return jsonObject;
+): TsConfigOptions | undefined {
+  if (jsonObject == null) return undefined;
   const {
     compiler,
     compilerHost,
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -1,4 +1,4 @@
-import { resolve, dirname } from 'path';
+import { resolve, dirname, isAbsolute } from 'path';
 import type * as _ts from 'typescript';
 import { CreateOptions, DEFAULTS, TSCommon, TsConfigOptions } from './index';
 import { getDefaultTsconfigJsonForNodeVersion } from './tsconfigs';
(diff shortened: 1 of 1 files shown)


cspotcode:
Linking to related tickets: #4, #921

src/configuration.ts
```diff
@@ -41,6 +41,36 @@ function fixConfig(ts: TSCommon, config: _ts.ParsedCommandLine) {
...
+      [configKey]: { ...extending[configKey], ...original[configKey] },
+    });
+  };
+
+  const { extends: extendFile, ...config } = rawConfig;
+
+  const result =
+    extendFile === undefined
+      ? filterRecognizedTsConfigTsNodeOptions(config[configKey])
+      : [extendFile]
```
cspotcode, line 64: Nit: get rid of the `filter().map()` chaining.  We prefer code like `const`, `if()`, `for(const a of b)`, etc.

src/configuration.ts
```diff
@@ -41,6 +41,36 @@ function fixConfig(ts: TSCommon, config: _ts.ParsedCommandLine) {
...
+          .filter((x) => x && typeof x === 'string')
+          .map((x) => x as string)
+          .map((extended) => resolve(cwd, extended))
```
cspotcode, line 67: Does this work when ./bar/tsconfig.json refers to ./bar/tsconfig2.json by relative path? Because the path is not relative to cwd?  Does it work for `"extends": "@tsconfig/node14/tsconfig.json"`?  (https://github.com/tsconfig/bases)

cspotcode (changes requested):
Made an initial review.  This will also need tests.

src/configuration.ts
seivan: @cspotcode It's not for everyone 🐼. Once you have a look over the logic itself, I'll swap it out.

src/configuration.ts
seivan: Regarding relative paths... 
Not sure if I want to say yes, I do use this myself and I my cwd wasn't relative, but then again I am doing other stuff that might impact it. 
Good call, would do best to extract `dirname/basename` from current config in order to make relative paths work - but I gotta check first. 
There is another approach and that is to let `ts` itself try to find the file, there is a function that for that last I checked. 


Regarding `@tsconfig/node14/tsconfig.json`
Yeah no chance that would work. 
I am not familiar with that, is that something you distribute or TS? 
If it's the latter, then I'd reckon there wouldn't be a `ts-node` key there?

src/configuration.ts
```diff
@@ -41,6 +41,36 @@ function fixConfig(ts: TSCommon, config: _ts.ParsedCommandLine) {
   return config;
 }
 
+function readTsConfigExtendForTsNodeRecursively(
+  cwd: string,
+  ts: TSCommon,
+  rawConfig: { [key: string]: any } = {}
```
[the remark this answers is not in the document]
seivan: Those aren't exactly the same thing unless you mean `Record<string, any>`? 
The reason for that is to extract out the extends file using spread and more importantly get type checks on objects. 
Using `any` makes it real tricky to work with nil coalescing and chain operators, as in the type checker won't complain if you aren't offering a default value.

src/configuration.ts
cspotcode: We'll need tests for this, and we'll need to support typescript's `"extends"` resolution behavior.  @tsconfig/node14 was just an example; our users can choose to `"extends"` from anywhere that typescript supports, and we'll need to match that behavior.  Ideally, we figure out how TypeScript is doing this logic and do the exact same, using as many TypeScript APIs as possible to avoid duplication.

https://www.typescriptlang.org/tsconfig#extends

seivan:
@dandv

> regarding extending tsconfig, might what you pointed out be the reason why transpileOnly didn't work for me if placed in the parent tsconfig.json?

Correctomundo. You can see that by running `tsc -p tsconfig.json --showConfig` on your "final" `tsconfig`. It should show you what your final config looks like once TS gets its hand on it. 

What I am trying to do here is somewhat of a hack really, it's just recursively following the extends path and collects `ts-node` key value pairs until it can merge them together.

src/configuration.ts
seivan: Right, if TS exposes that, then I guess best bet is to use it, but to reimplement it is overkill. 
In general, I think the simplest approach is just following the links.

In regards to your previous points about relative paths, this is an easy fix: 
```ts
.map((extended) => path.isAbsolute() ?  [path] : [currentConfigPath, extendPath]
.map((extended) => resolve(currentConfigPath, extended))
```

However I don't care much for `extends: "@tsconfig/node14/tsconfig.json"`, on the premise that I just don't approve of this path they're going down on.

cspotcode:
I looked at TypeScript's source code.

`parseJsonConfigFileContent`, which ts-node calls, is responsible for discovering, parsing, and merging extended tsconfigs.

It internally makes this chain of function calls:
- `parseJsonConfigFileContentWorker`
  - calls `parseConfig`
    - calls `parseOwnConfigOfJson`
      - calls `getExtendsConfigPath`:
        -  resolves the "extends" string into a path to the extended config file.
    - calls `getExtendedConfig` passing the path to the resolved extended config
      - calls `readJsonConfigFile`
      - calls `parseConfig` (recursing)
    - merges options from the extended and extending configs

We need to mimic the resolution logic of `getExtendsConfigPath`, but unfortunately it's not exported.  However, it's pretty short, so we could copy-paste it.  It calls out to `nodeModuleNameResolver`, which *is* exported.  (Technically calling an internal signature, but that's ok; we call a few internals as long as they're exposed)

`parseJsonConfigFileContent` also accepts a cache, a JS `Map`, to store parsed config files.  We can use it to avoid double-parsing each tsconfig file.

seivan:
@cspotcode Sounds a bit like what I mentioned earlier
Looking at it https://github.com/microsoft/TypeScript/blob/38da7c600c83e7b31193a62495239a0fe478cb67/src/compiler/moduleNameResolver.ts#L956-L957

It seems like a simplified version will just do the job fine - unless you want to go ahead and use all that. 

```ts
.map(path => isAbsolute(path) ? [path] : [currentConfigPath, path])
.map(paths => resolve(paths)
```

src/configuration.ts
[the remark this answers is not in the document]
seivan: Switched to Record

src/configuration.ts
seivan: Using relative paths and not CWD now.

seivan:
@cspotcode 

> parseJsonConfigFileContent also accepts a cache, a JS Map, to store parsed config files. We can use it to avoid double-parsing each tsconfig file.


Is this a cache that they offer while traversing the extend or something you had in mind to store? 
There is no double parsing when following the extend for `ts-node` so I am assuming you mean the using cached content of initial parse?

cspotcode:
> Is this a cache that they offer while traversing the extend or something you had in mind to store?

I'm looking at the `extendedConfigCache` argument to `parseConfig` and `getExtendedConfig`.

src/configuration.ts
cspotcode: > However I don't care much for extends: "@tsconfig/node14/tsconfig.json", on the premise that I just don't approve of this path they're going down on.

Unfortunately tsc already does this so we need to match.  Otherwise it gets confusing and we need to document how our behavior deviates, and I'll be the one answering the issues that get files.

I certainly understand why you might choose not to use this feature on your projects, but for `ts-node`, we need to match TSC's behavior or else we're making things confusing.

cspotcode:
@seivan do you plan to come back to this, or should I close it?

seivan:
@cspotcode I actually think your original idea of supporting a single env variable with all the options was better. I did this for my fork and thought it was enough but as I understand, you want the exact same inheritance support as `TS` has with its extend keyword. I am not sure I want to go down that rabbit hole for just this.

cspotcode:
@seivan Understood, no worries.

PR #1297, opened 2021-04-13
Website updates for v10 release

cspotcode:
Goal is to move all documentation into the new website format.

Closes #1296 
Closes #1228 
Closes #1221

## TODOs

* [ ] Refer to CommonJS / ESM consistently across all docs; link to common page explaining the difference
  * CommonJS
  * Native ECMAScript modules
  * ESM
  * native ESM
* [ ] Refer to ts-node consistently; naming is inconsistent throughout the readme
  * TypeScript Node
  * TS Node
  * ts-node
  * `ts-node` (in backticks)
* [x] write Usage page
* [ ] remove "under construction" banner
* [x] transform links in README.md to point to headers correctly
* [ ] Revisit "Troubleshooting errors" page
* [x] Revisit "Imports: CommonJS vs native modules"
  * name
    * CommonJS vs native ECMAScript modules
  * add "best of both worlds" explanation to CommonJS: can use modern syntax while preserving compatibility with the majority of node tooling that has not necessarily updated to native ESM
* [ ] Revisit "shebang"
  * move to "advanced?"
  * deduplicate with "usage"
  * explain shebang gotchas
  * consider keeping as discussion thread?
* [ ] add "API" page to "Advanced" section
  * list cool things you can do with the API
    * create a custom REPL
    * simpler TS compiler wrapper
    * detect if ts-node is already installed via `process.`
* [x] Add "Make it fast" page describing transpileOnly and `swc` compiler

PR #1298, opened 2021-04-19
Update AVA recipe

novemberborn:
* Fix spelling
* Add AVA 4 config

### DIFF
--- a/website/docs/recipes/ava.md
+++ b/website/docs/recipes/ava.md
@@ -1,8 +1,8 @@
 ---
-title: Ava
+title: AVA
 ---
 
-Assuming you are configuring Ava via your `package.json`, add one of the following configurations.
+Assuming you are configuring AVA via your `package.json`, add one of the following configurations.
 
 ## If you are downleveling to CommonJS (recommended)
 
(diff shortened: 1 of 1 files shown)


cspotcode:
Thanks for this.  Is AVA 4 released yet?  If so, would you mind putting the AVA 4 config above the AVA 3 config? I figure it's better for people to see the most recommended, up-to-date configs at the top of the page, and need to scroll down for older, out-of-date stuff.  What do you think?

novemberborn:
`npm i ava@latest` still gives you AVA 3, so I think that should remain at the top.

cspotcode:
Is AVA 4 unstable? is it recommended for new users?  I want to understand
in case I get tickets about it.

novemberborn:
It's a prerelease, alpha at the moment, though that's more a reflection on planned breaking changes than an assessment of stability.

Perhaps add "(available as a prerelease)"?

cspotcode:
I took me a while to get back to this, but I ended up rolling it into #1297
https://typestrong.org/ts-node/docs/recipes/ava

The prerelease stuff is commented out but we can update the docs when it's out of prerelease.  I just want to keep things simple on my end.

novemberborn:
Great!

PR #1351, opened 2021-05-28
docs(recipe): how to use `cross-env` or `dotenv`

reslear:


cspotcode:
This will need another polish pass before we can add it to our docs.

Some thoughts:

- should we recommend using `"require": ["cross-env/config"]` option in `tsconfig.json` so that additional flags do not need to be added to `package.json` scripts?  Similar to https://typestrong.org/ts-node/docs/paths
- I see some typos; will need a spellcheck
- formatting should be cleaned up
- Would it be better with an example `.env` file?
- Add titles to fenced code blocks, for example: <code>```json title="tsconfig.json"</code>
- needs to be added to `sidebars.js` (see https://github.com/TypeStrong/ts-node/blob/main/CONTRIBUTING.md#documentation)

If you'd rather not do these cleanup tasks, then you might want to share this in our discussion forum instead.  It would go well in the "Recipes" category.  https://github.com/TypeStrong/ts-node/discussions

cspotcode:
Moved to #1612

PR #1364, opened 2021-06-06
Disk cache prototype

### LINKED ISSUE #908: Disk caching in transpile-only via new project "typescript-cached-transpile"
I'm sharing this here in case anyone else is having speed issues even with transpile-only.

I wrote a new npm module, [typescript-cached-transpile](https://www.npmjs.com/package/typescript-cached-transpile), which monkey-patches typescript's `transpileModule` function to use a disk cache.

Here's the relevant code: https://github.com/cspotcode/personal-monorepo/blob/master/packages/typescript-cached-transpile/src/create.ts#L21-L88

It's very conservative.  It doesn't look at filesystem atimes or mtimes.  The cache key is computed from sha1 sums of both the configuration (typescript version, typescript-cached-transpile version, and all compiler flags) and the filename and source text.  It doesn't work with transformers, and it doesn't cache when the compiler returns diagnostics, because I don't cache the diagnostics. (theoretically we could cache them as JSON pretty easily)

I tested on my team's codebase, where `ts-node --transpile-only` was adding a whopping 13 second(!) to startup.  Even with the overhead of computing sha1 sums, caching reduced the overhead to about a second.

I understand the reasons for not introducing an incomplete caching solution to ts-node.  However, in my case, I need things to be fast, which means I already *need* transpile-only mode, so it makes perfect sense to use a caching solution that requires it.

If you use this, and it works well or you find bugs, please let me know.

### LINKED ISSUE #951: Restore cache functionality (default to off)
#701 removed caching and I find performance of using ts-node [unbelievably slow](https://github.com/TypeStrong/ts-node/issues/754) in any version beyond v7.0.1. The commit which removes caching behavior is in response to #672 (Cache doesn't take modified dependencies into account). I use ts-node in combination with `TS_NODE_FILES=true` and have never hit this problem nor was I able to reproduce the issue.

It seems that the maintainer is willing to accept a PR to add caching behavior in #920. Before I submit a PR, I'd like to confirm if it will be acceptable to restore the previous behavior, but have caching disabled by default and update the README to state that this is discouraged and link th #672 as to the reason why?

cspotcode:
Disk caching prototype.  Created in the SWC integration, but should be generalized to support `transpileModule`, too.  swc is so fast that caching doesn't really help at all, so this would only be useful for `transpileModule`.

Closes #908
Closes #951

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -4,7 +4,7 @@ import { join, resolve, dirname, parse as parsePath } from 'path';
 import { inspect } from 'util';
 import Module = require('module');
 import arg = require('arg');
-import { parse, createRequire } from './util';
+import { parse, createRequire, hasOwnProperty } from './util';
 import { EVAL_FILENAME, EvalState, createRepl, ReplService } from './repl';
 import { VERSION, TSError, register } from './index';
 import type { TSInternal } from './ts-compiler-types';
@@ -371,11 +371,6 @@ function evalAndExit(
   if (isPrinted) {
     console.log(typeof result === 'string' ? result : inspect(result));
   }
-}
-
-/** Safe `hasOwnProperty` */
-function hasOwnProperty(object: any, property: string): boolean {
-  return Object.prototype.hasOwnProperty.call(object, property);
 }
 
 if (require.main === module) {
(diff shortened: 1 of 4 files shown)


CMCDragonkai:
Is there any progress on this?

cspotcode:
No need, superceded by the swc option.

kkimdev:
This would be helpful for us as our project might need to use tsc for type reflection. (Deepkit RPC requirement)

PR #1376, opened 2021-06-20
Test coverage: moduleType overrides during ts-node loader usage

jayaddison:
Extends #1371.

This isn't working as expected currently: when running the `test.esm.js` module directly (from the base directory: `TS_NODE_PROJECT=tests/module-types/tsconfig.json node --trace-warnings --experimental-specifier-resolution=node --loader ts-node/esm tests/module-types/test.esm.js`), [one of the assertions fails](https://github.com/jayaddison/ts-node/blob/c9c443c57a6c636d9400ba5983b2f14a347a169c/tests/module-types/test.esm.js#L7) because both `import`ed modules have a value of `function` for `requireType`.

Without the `--experimental-specifier-resolution=node` flag, a `ERR_UNSUPPORTED_DIR_IMPORT` error is raised by the attempts to import modules from directory paths.


cspotcode:
Thanks, what needs to happen next?  Does `tests/module-types/tsconfig.json` need to be modified to add the necessary rule for `esm-exception`?

> Without the --experimental-specifier-resolution=node flag, a ERR_UNSUPPORTED_DIR_IMPORT error is raised by the attempts to import modules from directory paths.

Is this necessary to get code coverage?

jayaddison:
> Does `tests/module-types/tsconfig.json` need to be modified to add the necessary rule for `esm-exception`?

I don't think so - [that `tsconfig.json` should already set an ESM override](https://github.com/TypeStrong/ts-node/blob/c9c443c57a6c636d9400ba5983b2f14a347a169c/tests/module-types/tsconfig.json#L7) for files in subdirectores, if I understand it correctly?

(just in case, I did try adding an explicit `esm-exceptions` subdirectory configuration there; no dice, both `requireType` values are still returned as `function`)

> > Without the --experimental-specifier-resolution=node flag, a ERR_UNSUPPORTED_DIR_IMPORT error is raised by the attempts to import modules from directory paths.
> 
> Is this necessary to get code coverage?

Not for code coverage, no, but to get the test module to run without raising that error it does seem to be required.

cspotcode:
Oh, I see the confusion.  The logic shares elements of both TS `"include"` globbing and .gitignore / .gitattributes.

Globs match the behavior of TS `"include"`, so `src/cjs-subdir` matches everything in that directory.  It is an "implicit glob."

Patterns are matched from top to bottom, and anything at the bottom overrides anything at the top.  Same as `.gitignore` or `.gitattributes` where anything at the bottom overrides anything at the top.

https://github.com/TypeStrong/ts-node/blob/c9c443c57a6c636d9400ba5983b2f14a347a169c/tests/module-types/tsconfig.json#L6-L8

> Not for code coverage, no, but to get the test module to run without raising that error it does seem to be required.

What about importing the files by name instead of importing the directory?

jayaddison:
Thank you - I've applied some rearrangements here, and am now able to run `TS_NODE_PROJECT=tests/module-types/tsconfig.json node --trace-warnings --experimental-specifier-resolution=node --loader ts-node/esm tests/module-types/test.esm.js` and get a passing result.

That said I'll want to reconfirm that the test coverage is really proving what it claims to, and also ensure that it works when run as part of the standard test suite.

jayaddison:
(to try to explain what I meant about proving the test coverage: we should get confidence that the overrides are really being used, to make sure that passing test results aren't simply a result of default behaviour)

cspotcode:
> to make sure that passing test results aren't simply a result of default behaviour

Agreed.  Does it make sense to copy-paste `tests/module-types` into `tests/module-types-2`, and make it into an opposite of the first project?  That is to say, one of them has `package.json` `type: "module"` and uses overrides to force CommonJS, and the other project has `package.json` without `type` and uses overrides to force ESM.

Can we remove `--experimental-specifier-resolution=node`?  That would make the tests a bit simpler to understand.

jayaddison:
> > to make sure that passing test results aren't simply a result of default behaviour
> 
> Agreed. Does it make sense to copy-paste `tests/module-types` into `tests/module-types-2`, and make it into an opposite of the first project? That is to say, one of them has `package.json` `type: "module"` and uses overrides to force CommonJS, and the other project has `package.json` without `type` and uses overrides to force ESM.

Sounds good; I've been messing around a bit further here as you'll see, but the tests do not pass yet.
 
> Can we remove `--experimental-specifier-resolution=node`? That would make the tests a bit simpler to understand.

I'll try to, yep.  Whenever I've tried that previously it has led to the `ERR_UNSUPPORTED_DIR_IMPORT` errors, but I'll try to learn more about that and remove the additional option if possible.

cspotcode:
Test are failing on 14.13.0 and 12.15.  I believe we test on these versions specifically because they do *not* throw an error when attempting to `require()` ESM; instead they allow it to be executed as CJS.  We'll need to figure out what the correct behavior is in those situations.  I assume we should either tweak `ts-node` to throw the error, respecting the moduleTypes overrides, or disable these unit tests on those versions, because users can reasonably be expected to upgrade to the latest node 12 or 14 to leverage newer `ts-node` features.

jayaddison:
Thanks @cspotcode - I guess that kind of require of ESM-as-CJS-by-default might only work by chance, if at all?

Looking at some of the code comments (particularly [this one](https://github.com/TypeStrong/ts-node/blob/5643ad64cf39ee0dfa2a9323e8d1dd9f400e5884/src/index.ts#L41-L42)), would it be fair to say that `ts-node` should try to do what `node` itself does?

(in this case currently, that's to throw an `ERR_UNKNOWN_FILE_EXTENSION` type error exception under Node 12 due to the `.ts` extension..)

jayaddison:
> (in this case currently, that's to throw an `ERR_UNKNOWN_FILE_EXTENSION` type error exception under Node 12 due to the `.ts` extension..)

...I've since guessed that that's probably not a sensible error to reproduce, because that was encountered when running the `test.esm.js` code without any preprocessing / transpilation (if that's the correct term?) of the imported `.ts` files into JavaScript.

cspotcode:
The goal is to match node's behavior.  The behavior matching is as follows:
- `.ts`, `.tsx`, `.js`, `.jsx` files: mimic node's handling of `.js` files, as if the file had the `.js` extension
- `.ts`, `.tsx`, `.jsx` that match a `moduleTypes` override:
  - if the override is CJS: behave as if the file had the `.cjs` extension.  Whatever node would do for a `.cjs` file, we do the same
  - if the override is ESM: behave as if the file had the `.mjs` extension.  Whatever node would do for a `.mjs` file, we do the same

We must behave the same way that node would, which means:
- if node would throw an error when attempting to `require` the file, we should throw the same error
- if node would execute the file as CJS, we should do the same
- if node would execute the file as ESM, we should do the same

In order to correctly match node's behavior with the `ERR_UNKNOWN_FILE_EXTENSION` error, which has changed over time, we check node's version number:
https://github.com/TypeStrong/ts-node/blob/main/src/index.ts#L33-L38
(we also version switch in [one other place](https://github.com/TypeStrong/ts-node/blob/main/dist-raw/node-esm-resolve-implementation.js#L10-L26), but that is for a different reason)

cspotcode:
I just realized your comment was about `ERR_UNKNOWN_FILE_EXTENSION`, not about `ERR_REQUIRE_ESM`.  My mistake.  I've edited my last comment to say `ERR_REQUIRE_ESM` since that is what I was referring to.

The goal regarding `ERR_UNKNOWN_FILE_EXTENSION` is to teach node what to do for `.jsx`, `.ts`, and `.tsx` extensions so that it doesn't have to throw that error when it encounters those files.  I agree with you that it is not a sensible error to reproduce, because we want to teach node to avoid that error.

jayaddison:
No problem at all; we're on the same page, I think - and it's better to overcommunicate in these kinds of conversation.

If I understand correctly then we'll end up with a test matrix of six possible outcomes when `moduleTypes` overrides are in place:

- Two outcomes like the ones that pass already for recent Node versions (1x CJS, 1x ESM)
- Two outcomes for Node ~= 14.13 (1x CJS, 1x ESM)
- Two outcomes for Node ~= 12.15 (1x CJS, 1x ESM)

It's possible these collapse down to four outcomes, but it'll be a little while before I have time to confirm behaviour across both Node 12.15 and Node 14.13.

I also get the feeling that the test coverage here might end up being a base level of coverage rather than exhaustive; in other words, it's not going to reliably test all the differences that could occur for varying CJS and EJS file content.  Still, making sure that each line of code is exercised seems like a worthwhile improvement.

Feel free to continue on further with the branch if you will have more time and/or want to progress this soon.  I should have some time outside work hours next week to look at this again.

jayaddison:
I've added 'expected failure' cases for the moduleType override test under Node 12.15, 14.13 - but also have to admit I haven't yet gone upstream to Node to figure out if those are really the ranges affected (i.e. why the failures on those versions as of 7b6daa3 occurred for those unit test runs).

cspotcode:
That's fine, I feel confident those node versions are in the test matrix
for this reason.  There may be some comments in the GH actions workflow
file; I'm not sure.  FWIW `n` and `volta` make it fast and easy to switch
node versions.  I use Volta for all this testing.

Is there anything else required here?  Tests don't need to be perfect as
long as we think they are proving that the feature works.

jayaddison:
> Is there anything else required here?

I don't _think_ so?  Although I do feel a bit out of my depth here.

cspotcode:
@jayaddison ok, thanks, I'll give this a proper review as soon as I have time, and hopefully get this merged.

jayaddison:
Thanks in advance @cspotcode, and take your time; glad to make any further adjustments to make sure the code's acceptable.

PR #1388, opened 2021-07-07
fix: prevent implicit 'use strict' output

### LINKED ISSUE #1387: REPL outputs `'use strict'` after first empty line of input
### Expected Behavior

If you type nothing into the first line of REPL input, you should not see `'use strict'`.  The TS compiler emits `'use strict'` but that's not what the user entered, so they shouldn't see it.

### Actual Behavior

Hitting enter in the REPL with no input on the first line shows `use strict`

ejose19:
Fixes #1387


cspotcode:
Thanks!

PR #1395, opened 2021-07-12
chore: normalize npm version in ci jobs

ejose19:
Having 2 different versions of `npm` could lead to issues causing the jobs to randomly fail.

Related:
https://github.com/actions/setup-node/issues/268
https://github.com/npm/cli/issues/417


cspotcode:
I ended up making a smaller tweak to keep the npm caches separated.

PR #1396, opened 2021-07-13
Set of extra ignored diagnostics for REPL mode

### LINKED ISSUE #469: Throwing error makes all future code unreachable
```
> throw new Error('Hi GH')
[eval].ts:2
throw new Error('Hi GH');
^

Error: Hi GH
    at [eval].ts:2:7
    at ContextifyScript.Script.runInContext (vm.js:59:29)
    at ContextifyScript.Script.runInNewContext (vm.js:65:15)
    at _eval (C:\Users\CorbinCrutchley\AppData\Roaming\npm\node_modules\ts-node\dist\_bin.js:181:29)
    at REPLServer.replEval (C:\Users\CorbinCrutchley\AppData\Roaming\npm\node_modules\ts-node\dist\_bin.js:220:18)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:116:13)
    at REPLServer.emit (events.js:211:7)
> let a = 5
Thrown: ⨯ Unable to compile TypeScript
[eval].ts (1,1): Unreachable code detected. (7027)
```

This behavior is not present in the `node` REPL.

### LINKED ISSUE #729: Can't redefine a function
(Similar but possibly different / easier than https://github.com/TypeStrong/ts-node/issues/472)

A popular workflow in the Lisp/Clojure community is "REPL-driven development", by which you develop things at the REPL first, repeatedly (re)defining functions at a granular level, until each function is deemed correct.

Errors like `[eval].ts(17,10): error TS2393: Duplicate function implementation.` impair this workflow, and make `ts-node` less useful for such a crowd.

Is there any way to disable this behavior?

Thanks - Victor

### LINKED ISSUE #850: noUnusedLocals causes problems working in repl
When ts-node is launched in console mode the `noUnusedLocals` option causes problems and makes the console hard to use:

```
$ ts-node
> import {ZnsDomain} from './lib/models'
[eval].ts:1:1 - error TS6133: 'ZnsDomain' is declared but its value is never read.

1 import {ZnsDomain} from './lib/models'
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

undefined
```

Can we ignore this compiler option when running in console mode or find any other workaround?

### LINKED ISSUE #1120: Set of extra ignored diagnostics for REPL mode
In various tickets, users point out TS diagnostics that are annoying in the REPL.  I wonder if we should have a set of diagnostics to be ignored only in the REPL, only in the `eval.ts` file.  It could be enabled with a `--loose-repl` flag, or we could add it as an `--experimental-loose-repl` flag and make it default in a future major version.

The idea is to put links to those tickets here as I find them, sorta like an Epic for improving the REPL diagnostics experience.

Tickets:
#729 (though a proper solution to this is somehow removing the old function declarations which is probably way too complex)
#850
#469

cspotcode:
Closes #1120 
Closes #729
Closes #850
Closes #469

Refactors diagnostic-ignoring logic to support per-file filters.

REPL can add a filter which applies *only* to the REPL's virtual file and removes diagnostics which are annoying in interactive mode.  For example, "unreachable code" and "duplicate function declaration"

Also refactors some test macros

### DIFF
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -205,7 +205,7 @@ export function main(
       repl: createRepl({
         state,
         composeWithEvalAwarePartialHost: evalAwarePartialHost,
-        ignoreExtraDiagnostics: false,
+        ignoreDiagnosticsThatAreAnnoyingInInteractiveRepl: false,
       }),
     };
     ({ evalAwarePartialHost } = evalStuff.repl);
@@ -221,7 +221,7 @@ export function main(
       repl: createRepl({
         state,
         composeWithEvalAwarePartialHost: evalAwarePartialHost,
-        ignoreExtraDiagnostics: false,
+        ignoreDiagnosticsThatAreAnnoyingInInteractiveRepl: false,
       }),
     };
     ({ evalAwarePartialHost } = stdinStuff.repl);
--- a/src/bin.ts
+++ b/src/bin.ts
@@ -205,6 +205,7 @@ export function main(
       repl: createRepl({
         state,
         composeWithEvalAwarePartialHost: evalAwarePartialHost,
+        ignoreExtraDiagnostics: false,
       }),
     };
     ({ evalAwarePartialHost } = evalStuff.repl);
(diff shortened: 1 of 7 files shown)


cspotcode:
I gave this some more thought; I feel good enabling by default in a minor release.

Issue #1126, opened 2020-09-24
`.type` command is crippling the REPL


thron7:
### Expected Behavior

    $ npx ts-node
    > let a = 3
    undefined
    > 3 + 4
    7
    > .type a
    let a: number
    > 3 + 4
    7

### Actual Behavior

    $ npx ts-node
    > let a = 3
    undefined
    > 3 + 4
    7
    > .type a
    let a: number
    > 3 + 4
    3

### Steps to reproduce the problem

- start the REPL
- define a symbol
- `.type` the symbol
- do a simple operation like `3 + 4`

### Specifications

* ts-node version: v9.0.0
* node version: v12.18.3
* TypeScript version: v4.0.3
* tsconfig.json, if you're using one:
```
{}
```
* Operating system and version: Windows 10 v1607
* If Windows, are you using WSL or WSL2?: neither

cspotcode:
Looks like this is only an issue on Windows, so it might have to do with newlines.

We only execute things added in the diff; not sure if the diff is showing "changed" instead of "added" on Windows due to newlines.
https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L335

Or maybe there's a cache poisoning issue in our `service`, since `.type` uses our REPL's `undo()` mechanism.
https://github.com/TypeStrong/ts-node/blob/master/src/bin.ts#L319

The compiled output gets stuck after a .type check, where the compiled output includes and stops at the typechecked identifier, even though the identifier has been removed from input and replaced with subsequent user input.
![image](https://user-images.githubusercontent.com/376504/94152417-870c0100-fe49-11ea-874f-cdb4660f72bf.png)

cspotcode:
I did some more digging into this.  It seems like it might be a bug inside TypeScript itself.

After we call `getQuickInfoAtPosition`, the service seems to get stuck for the file in question, returning the same emit output even when we update the script's source text to include new commands.

Additionally, it happens on Windows only.

ejose19:
I'd suggest renaming the issue to reflect the actual problem, as this is not a performance issue.

PR #1401, opened 2021-07-18
feat(repl): allow using "type" command on types/interfaces

ejose19:
This makes "type" command return a more accurate description than the currently `any` for types/interfaces. It's worth noting that until https://github.com/microsoft/TypeScript/issues/25894 is fixed, interfaces/enums will be shown as just "$kind $identifier".

Based on https://stackoverflow.com/a/55777714.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -348,6 +348,7 @@ export interface TsConfigOptions
 export interface TypeInfo {
   name: string;
   comment: string;
+  kind?: _ts.ScriptElementKind;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


ejose19:
Regarding the `undefined as unknown as ${identifier}`, it's required due https://github.com/TypeStrong/ts-node/issues/1126 (https://github.com/TypeStrong/ts-node/issues/1126#issuecomment-721392645), so it's done this way to avoid having errors in subsequent commands.

ejose19:
Had to split the type tests into 2 due to the current linked issue, as `getQuickInfoAtPosition` is using an invalid position after the first `type` call.

Also, fixed a test issue with ESM, since on 16.5.0 it no longer displays the error code in stdout (https://github.com/nodejs/node/commit/33cad271c583536befefaa9d320228c59c122ad3)

PR #1420, opened 2021-08-05
fix: allow to specify pretty errors in TTY

### LINKED ISSUE #1418: pretty option is ignored in terminal
### Expected Behavior

When I specify `pretty: false` I expect the errors to respect that.
This is particularly useful if I am developing a tool that uses ts-node and I need to handle 3rd party potentially faulty code.

### Actual Behavior

Because of this line, when I use a terminal as the main output for my tool, `pretty` is ignored: 
https://github.com/TypeStrong/ts-node/blob/e8a4d76ad6e9772296e31b871560ae6bd78a1165/src/index.ts#L628-L630

Run ts-node on a faulty codebase with `--pretty false` option.
Errors are still styled with terminal only decoration.

### Minimal reproduction

https://github.com/TypeStrong/ts-node-repros/pull/16

### Specifications

ts-node v8.3.0
node v14.17.0
compiler v4.3.5

tsconfig.json
```
{
  "compilerOptions":{
    "pretty": false
  }
}
```
* Operating system and version: MacOS - Big Sur

elevatebart:
closes #1418


cspotcode:
Looks good, did you test this?  I don't think it needs automated tests but it would be nice to do some manual testing:

```
ts-node --pretty ./file-with-errors.ts > out.txt // <-- not a TTY but should still have pretty errors
ts-node ./file-with-errors.ts // <-- in a directory where pretty: false is specified in tsconfig.json
```

cspotcode:
Looks like this needs `npm run lint-fix` to fix linter failures.

PR #1448, opened 2021-09-06
Remove misleading text about passing arguments to env

chee:
gnu's `env` has a `-S` argument, which splits the string into command and arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,12 +165,9 @@ ts-node-cwd script.ts
 console.log("Hello, world!")
 ```
 
-Passing CLI arguments via shebang is allowed on Mac but not Linux.  For example, the following will fail on Linux:
+Passing CLI arguments via shebang requires passing the `-S` argument to `env` when using GNU Coreutils.
 
-    #!/usr/bin/env ts-node --files
-    // This shebang is not portable.  It only works on Mac
-
-Instead, specify all `ts-node` options in your `tsconfig.json`.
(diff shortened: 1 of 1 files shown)


cspotcode:
Last I checked it's not available on all distros.  The relevant discussion might be in the issue tracker or in "discussions" somewhere; I'm not sure.

chee:
@cspotcode yeah, it's only been in GNU Coreutils since july 2018, so distros that have a 3+ year old version of gnu's `env` will not support it. the `-S` option is also needed for FreeBSD releases 6.0 and higher. `-S` is unsupported by OpenBSD, NetBSD and busybox.

cspotcode:
Ok thanks, is there some way we can either include that info in the docs, or link to a discussion

cspotcode:
Ok thanks for the detailed response.  Is there some way we can include that info in the docs, or link to a GH discussion thread with the details?  Could be something like "requires -S which is only on newer versions of env, see here (link to GH discussion) for details"

What do you think?

cspotcode (changes requested):
Before we can merge, this will need the changes we talked about here:

> Ok thanks for the detailed response. Is there some way we can include that info in the docs, or link to a GH discussion thread with the details? Could be something like "requires -S which is only on newer versions of env, see here (link to GH discussion) for details"

cspotcode:
Making appropriate changes as part of #1545, but I'm linking to your compatibility notes in the docs.

PR #1457, opened 2021-09-12
Update esm loader hooks API

### LINKED ISSUE #1372: Update ESM hook for consolidated node API
nodejs/node#37468 consolidates the loader hooks to only 2x functions.  If/when it lands, we'll need to export implementations of those functions.  We should be able to implement them largely on top of what we already have.  The `default*` functions we get from node will be different, though.

Updated node docs are here: https://github.com/nodejs/node/blob/master/doc/api/esm.md#loaders

jonaskello:
This is an attempt to fix #1372.

Currently I just added a load hook that throws but it does not get called when using node `v17.0.0-nightly20210912df22736d80`. Not sure if this is because I am doing something wrong or if the changes are not fully in nightly. This version of node does warn about the old hooks so not sure what is going on yet. Will try again with tomorrows nightly of node.

### DIFF
--- a/src/esm.ts
+++ b/src/esm.ts
@@ -27,7 +27,8 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
     preferTsExts: tsNodeInstance.options.preferTsExts,
   });
 
-  return { resolve, getFormat, transformSource };
+  // return { resolve, getFormat, transformSource };
+  return { resolve, load };
 
   function isFileUrlOrNodeStyleSpecifier(parsed: UrlWithStringQuery) {
     // We only understand file:// URLs, but in node, the specifier can be a node-style `./foo` or `foo`
@@ -64,11 +65,23 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
 
     // pathname is the path to be resolved
 
-    return nodeResolveImplementation.defaultResolve(
+    console.log('got here');
+
+    const x = nodeResolveImplementation.defaultResolve(
       specifier,
       context,
       defaultResolve
     );
+    console.log('x', x);
+    return x;
+  }
+
+  async function load(
+    url: string,
+    context: {},
+    defaultLoad: typeof load
+  ): Promise<{ format: Format }> {
+    throw new Error('load');
   }
 
   type Format = 'builtin' | 'commonjs' | 'dynamic' | 'json' | 'module' | 'wasm';
--- a/package.json
+++ b/package.json
@@ -65,7 +65,8 @@
     "test": "npm run build && npm run lint && npm run test-cov --",
     "test-local": "npm run lint && npm run build-tsc && npm run build-pack && npm run test-spec --",
     "coverage-report": "nyc report --reporter=lcov",
-    "prepare": "npm run clean && npm run build-nopack"
+    "prepare": "npm run clean && npm run build-nopack",
+    "esm-usage-example": "npm run build-tsc && cd esm-usage-example && node --experimental-specifier-resolution node --loader ../esm.mjs ./index"
   },
   "engines": {
     "node": ">=10.0.0"


cspotcode:
Thanks for the PR.  I'm swamped with work, so I'll do my best to review in a timely manner, but it might drag a bit.  However, anyone who wants to use this PR *before* it is merged and published can install directly from git.  npm's documentation explains how to install dependencies directly from a git branch instead npmjs.com.  Our CI also uploads a package tarball as an artifact, which anyone can download and install locally.

Re the warnings from node about deprecated hooks: I think we can detect the node version and conditionally export *only* the hooks expected by that version of node.  We can export `undefined` for the others.  We do version detection in a few places in the codebase; here is one: https://github.com/TypeStrong/ts-node/blob/aaf60523ac0f77dc52b3c729f1f179a85dcac2c0/src/index.ts#L43-L62

src/esm.ts
```diff
@@ -28,7 +28,8 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
     preferTsExts: tsNodeInstance.options.preferTsExts,
   });
 
-  return { resolve, getFormat, transformSource };
+  // return { resolve, getFormat, transformSource };
+  return { resolve, load };
```
cspotcode, line 32: We will need to detect the version number of node and export hooks appropriate to that version, essentially supporting both the old and new loader APIs at the same time.  While it's tempting to say that we break backwards-compatibility, this is really inconvenient in practice.  There are many reasons a user might be on node 14 -- it is the LTS version -- and I don't want to bar those users from access to newer ts-node features.

src/esm.ts
```diff
@@ -65,11 +66,23 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
...
+    console.log('x', x);
+    return x;
+  }
+
+  async function load(
+    url: string,
+    context: {},
+    defaultLoad: typeof load
+  ): Promise<{ format: Format }> {
+    throw new Error('load');
```
cspotcode, line 85: With the new loader hook consolidation, either the `resolve` or the `load` hook can fill the role of `getFormat`.  I haven't had time to dig in; do you know if it's better for `resolve` or `load` to perform `getFormat` duties?  I assume it should be combined with `resolve`, since there is nothing in `load` that affects the format.

src/esm.ts
jonaskello: Yes, I am actually using v14 myself with ts-node/esm so I agree backwards compat is necessary. I will look into you example above.

src/esm.ts
jonaskello: I'm not sure yet but I'll look deeper into how to merge the hooks. Just need to get load to be called at all first (node's current nightly is not calling into it for some reason).

jonaskello: Ah, forgot the re-export at the root of the repo (esm.mjs). Now load is getting called :-).

jonaskello: According to docs load() should return an object with format and source. So I think `resolve` can be left untouched and `load` should just call into the existing `getFormat` and `transformSource`. The main problem is the default that is being to deferred to is different as it is for `load`. So I need to find some way to provide a default for the existing getFormat` and `transformSource`.

src/esm.ts
jonaskello: Since `resolve` seem unchanged and load is new, I think we can just return an object with all hooks and that would make it backwards compatible. I'll give that a try.

cspotcode:
Should we add node's nightly builds to our test matrix?  We already test against nightly typescript.

src/esm.ts
cspotcode: Another thing to get right:

What happens when an ESM module tries to import a CommonJS module?  And what if that CommonJS module is `.ts` and needs to be transpiled?  I assume, although I'm not sure, that `defaultLoad` will call out to our `require.extensions` hook if you pass it a URL pointing to a TS file and `format: 'commonjs'`

cspotcode: I did some experimenting.  Here is what I found:

If `defaultLoad` does not receive a `format`, it will internally call the old `defaultGetFormat`.  If the computed format is `commonjs`, it will *not* attempt to load anything.  Instead it will return `{format: 'commonjs', source: null}`.  This tells node to use the CommonJS codepath, which respects our `require.extensions` hooks.

The problem is: if the format is determined to be `'module'`, it will try to load it.  If the file doesn't exist it will error.  Today, we intentionally call `defaultGetFormat` with files that do not exist.  If we want to classify `foo.ts`, we call `defaultGetFormat` with `foo.js`.

The situation is no better if *both* `foo.js` and `foo.ts` exist.  Because then it's loading the wrong file.  And there are other formats: json, wasm, etc.

cspotcode: Should we begin shipping our own copy of `defaultGetFormat`?

I believe that this is the correct file, at the correct version number matching what we already have:

https://github.com/nodejs/node/blob/v15.3.0/lib/internal/modules/esm/get_format.js

Matches the version from which we pulled https://github.com/TypeStrong/ts-node/blob/main/dist-raw/node-esm-resolve-implementation.js (See comments at top of file)

cspotcode:
Linking to actions/setup-node#80 re testing against node nightlies.  If they don't have a good solution, we can roll our own I guess.

src/esm.ts
jonaskello: I added the file you suggested and patched its imports and it seems to work.

src/esm.ts
```diff
@@ -72,6 +78,35 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
     );
   }
 
+  async function load(
+    url: string,
+    context: {},
+    defaultLoad: typeof load
+  ): Promise<{ format: Format; source: string | Buffer }> {
+    const { format } = await getFormat(url, context, defaultGetFormat);
```
cspotcode, line 86: What if `context.format` has already been set, perhaps because we have been composed with another loader?  Does `defaultGetFormat` respect this?

If I remember correctly, `defaultLoad` will respect an incoming `format` hint and will not overwrite it, so I think we should match that behavior.  Please correct me if I'm wrong.

src/esm.ts
```diff
@@ -72,6 +78,35 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
...
+
+    const { source: rawSource } = await defaultLoad(
+      url,
+      { format },
+      defaultLoad
+    );
+
+    const defaultTransformSource: typeof transformSource = async (
```
cspotcode, line 94: It took me a second to understand what's going on here; might benefit from a couple comments.

`defaultTransformSource` is an identity function?  Is that idea that any transformations from *outside* of ts-node, if they happen at all, are going to come from `defaultLoad`?

src/esm.ts
```diff
@@ -28,7 +29,12 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
     preferTsExts: tsNodeInstance.options.preferTsExts,
   });
 
-  return { resolve, getFormat, transformSource };
+  // The hooks API changed in node version X so we need to check for backwards compatibility
+  // TODO: When the new API is released, change to the correct node version here
+  const newHooksAPI = parseInt(process.versions.node.split('.')[0], 10) >= 17;
+  return newHooksAPI
+    ? { resolve, load }
+    : { resolve, getFormat, transformSource };
```
cspotcode, line 37: Somewhere in this file we should add a couple comments explaining our approach, since it might throw off a new reader.  We are implementing the *new* hooks API on top of our implementation of the *old* hooks API, which relies on copy-pasted code from the *old* hooks implementation in node.

cspotcode (changes requested):

src/esm.ts
jonaskello: Yes, it is a identity function because the old `defaultTransformSource` inside node was an identity function. 
https://raw.githubusercontent.com/nodejs/node/v15.3.0/lib/internal/modules/esm/transform_source.js
The idea was just to be able to re-use the existing `transformSource` and it required the old `defaultTransformSource` to defer to. Perhaps we should not use the existing functions for the old hooks in `load`?

jonaskello: I added some comments in an attempt to clarify.

src/esm.ts
jonaskello: I'm now checking the version to avoid the deprecation warning.

src/esm.ts
```diff
@@ -28,7 +29,12 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
     preferTsExts: tsNodeInstance.options.preferTsExts,
   });
 
-  return { resolve, getFormat, transformSource };
+  // The hooks API changed in node version X so we need to check for backwards compatibility
+  // TODO: When the new API is released, change to the correct node version here
+  const newHooksAPI = parseInt(process.versions.node.split('.')[0], 10) >= 17;
```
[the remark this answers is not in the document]
jonaskello: Changed to use `versionGte` (needed to export it).

src/esm.ts
jonaskello: Added some comments in `load` but perhaps more are needed?

src/esm.ts
jonaskello: You mean we should check if context already have a `format` and if so skip the call to `getFormat`?

jonaskello: I added a check for using existing format hint on the context, let me know if this is what you expected.

src/esm.ts
cspotcode: Thanks, I made some nitpicky changes, mostly I added a big block comment to the top of the file explaining today's implementation and how we will likely change it in the future.  I think (I hope?) this should address any confusion.

src/index.ts
```diff
@@ -47,7 +47,7 @@ export type {
 const engineSupportsPackageTypeField =
   parseInt(process.versions.node.split('.')[0], 10) >= 12;
 
-function versionGte(version: string, requirement: string) {
+export function versionGte(version: string, requirement: string) {
```
cspotcode, line 51: Marked as `@internal`

cspotcode:
I've added node nightly to the test matrix which is causing some failures.  It also looks like stack traces are not showing correct line numbers for the errors.  I don't know if this is a ts-node bug, a node nightly bug, or if those stack traces have never been correct.

jonaskello:
According to the [docs](https://github.com/nodejs/node/blob/master/doc/api/esm.md#loadurl-context-defaultload) it seems `source` return value for formats `builtin` and `commonjs` should be undefined. I added a check for this and I think it might fix the tests.

jonaskello:
Well, it seemed to fix all the tests but one :-).

cspotcode:
That's great; do you know what's up with the remaining failing test?

jonaskello:
Thanks, I'll try to get some time to look into the last test failure.

FYI, I'm also working towards a more complete solution for the resolve hook in a new package called [ts-esm-resolve](https://github.com/jonaskello/ts-esm-resolve). It is still in an early stage but it would of course be interesting to know if something like this package would be considered useful for ts-node?

cspotcode:
The fix was a one-line change; tests are passing now. https://github.com/TypeStrong/ts-node/pull/1457/commits/63c06182f29b4c40367ff92dac048a7a83388c78

> it would of course be interesting to know if something like this package would be considered useful for ts-node?

Possibly, yeah.  We want to do both CJS and ESM resolution so that projects "just work," so if we support project references for ESM it should also work with CJS.  And since we already resolve the user's configuration, we'd want it to accept a pre-parsed config or set of multiple project-referenced configs, share FS cache, stuff like that.  We'll also need to be able to hook into format determination to override as dictated by https://typestrong.org/ts-node/docs/module-type-overrides

Glancing at the readme, looks like it ticks at least a few of those boxes.

cspotcode:
Does esbuild's sync API still have that per-call process spawning overhead?  That's why I implemented swc support for ts-node, but I want to add esbuild as well once that process-spawning limitation goes away.

jonaskello:
> The fix was a one-line change; tests are passing now. 

Nice :-)

>  We want to do both CJS and ESM resolution...

The initial goal for ts-esm-resolve is for ESM resolution. Not sure how much work it would be to add CJS resolution. A lot of the logic is of course similar so it might be rather easy.

>  ...accept a pre-parsed config or set of multiple project-referenced configs

Yes, currently it is accepting a path to a tsconfig file and resolving it to an internal `Map` of all referenced tsconfigs. Might be better to make that map a direct input param instead so pre-parsed config could be passed in.

> ...share FS cache

That should be possible since all FS calls are done through callback functions passed to the resolve function. I currently use this to mock the filesystem in the test-cases but it should also be possible to use it for cached FS calls.

jonaskello:
> ...hook into format determination

AFAIK the resolve process does not need to know the format, that is only needed later when executing the module?

cspotcode:
Noting that here I was concerned about some sourcemaps, but I can't remember why or if they're fixed or covered by tests.
https://github.com/TypeStrong/ts-node/pull/1457#issuecomment-922379846

This is still an experimental API surface so I feel good merging now.

PR #1468, opened 2021-09-19
Switch from prettier to dprint

cspotcode:
It's faster and formatted output is very similar.

Configuration in dprint.json is not based on any opinions; the goal is to minimize the diff that will be applied when running dprint for the first time.  Those options make dprint's output similar to what we already get with prettier.


cspotcode:
Closing in favor of #1722

PR #1469, opened 2021-09-20
Create SECURITY.md

### LINKED ISSUE #1464: Trying to get in touch regarding a security issue
Hey there!

I'd like to report a security issue but cannot find contact instructions on your repository.

If not a hassle, might you kindly add a `SECURITY.md` file with an email, or another contact method? GitHub [recommends](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) this best practice to ensure security issues are responsibly disclosed, and it would serve as a simple instruction for security researchers in the future.

Thank you for your consideration, and I look forward to hearing from you!

(cc @huntr-helper)

zidingz:
To verify that the repository authorises cspotcode@gmail.com as its security contact. Closes #1464

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,5 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+Please report security issues to `cspotcode@gmail.com`


cspotcode:
@blakeembrey is this ok with you?  Is there a shared TypeStrong email we should use instead?

cspotcode:
@zidingz I have "admin" access to this repository, so I have the ability to merge this PR without any approval, but I still want to double-check if there is interest in using a shared team email address as our security contact.

blakeembrey:
@cspotcode This is good with me until we find another solution. Do you know of any free shared email inbox solutions?

cspotcode:
I don't.  I think Gmail has some sort of option where you can share access to an inbox?  But the 2fa would still need to be owned by one person.

Or is it sufficient to say that messages should be addressed to 2x addresses, and rely on reply-all to continue the conversation.  Back when I worked on projects which required all email to be encrypted, this is what we did, because there's no way to encrypt messages to a mailing list.  We don't have that requirement here, but still, using reply-all is probably more common than it might seem.

cspotcode:
@zidingz I have merged this PR.  We may amend the security contact soon, but for now, `cspotcode@gmail.com` is correct.

PR #1480, opened 2021-10-01
Fix REPL bug with previous line carryover

### LINKED ISSUE #1363: Bug with binary operators in REPL
### Expected Behavior

It should produce a syntax error.

### Actual Behavior

It seems to combine what you typed with the expression you entered previously.

```
$ npx ts-node
> 100 + 10
110
> * 4
140
```

### Minimal reproduction

See above.

### Specifications

* ts-node version: v10.0.0
* node version: v12.16.1
* TypeScript version: v4.3.2
* tsconfig.json, if you're using one: problem still occurs when none present
* Operating system and version: Ubuntu 18.04.5 LTS
* If Windows, are you using WSL or WSL2?: N/A

TheUnlocked:
Fixes #1363.

Multiline inputs aren't an issue since multiline inputs are passed into `evalCodeInternal` in their entirety.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -208,23 +208,13 @@ export function createRepl(options: CreateReplOptions = {}) {
     context: Context;
   }) {
     const { code, enableTopLevelAwait, context } = options;
-    const result = appendCompileAndEvalInput({
+    return appendCompileAndEvalInput({
       service: service!,
       state,
       input: code,
       enableTopLevelAwait,
       context,
     });
-    // A semicolon is added to make sure that the code doesn't interact with the next line,
-    // for example to prevent `2\n+ 2` from producing 4.
-    // This is safe since the output will not change since we can only get here with successful inputs,
-    // and adding a semicolon to the end of a successful input won't ever change the output.
-    // We check to make sure the previous line ends in a newline just in case, even though it should
-    // always be the case.
-    if (state.input.charAt(state.input.length - 1) === '\n') {
-      state.input = `${state.input.slice(0, -1)};\n`;
-    }
-    return result;
   }
 
   function nodeEval(
@@ -520,6 +510,12 @@ function appendCompileAndEvalInput(options: {
     undo();
   } else {
     state.output = output;
+
+    // Insert a semicolon to make sure that the code doesn't interact with the next line,
+    // for example to prevent `2\n+ 2` from producing 4.
+    // This is safe since the output will not change since we can only get here with successful inputs,
+    // and adding a semicolon to the end of a successful input won't ever change the output.
+    state.input = `${state.input.slice(0, -1)};\n`;
   }
 
   let commands: Array<{ mustAwait?: true; execCommand: () => any }> = [];
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -217,14 +217,13 @@ export function createRepl(options: CreateReplOptions = {}) {
     });
     // A semicolon is added to make sure that the code doesn't interact with the next line,
     // for example to prevent `2\n+ 2` from producing 4.
-    // We don't care about the result so we just toss it.
-    appendCompileAndEvalInput({
-      service: service!,
-      state,
-      input: ';\n',
-      enableTopLevelAwait,
-      context,
-    });
+    // This is safe since the output will not change since we can only get here with successful inputs,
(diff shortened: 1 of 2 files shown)


src/repl.ts
```diff
@@ -208,13 +208,23 @@ export function createRepl(options: CreateReplOptions = {}) {
...
       enableTopLevelAwait,
       context,
     });
+    // A semicolon is added to make sure that the code doesn't interact with the next line,
+    // for example to prevent `2\n+ 2` from producing 4.
+    // This is safe since the output will not change since we can only get here with successful inputs,
+    // and adding a semicolon to the end of a successful input won't ever change the output.
+    // We check to make sure the previous line ends in a newline just in case, even though it should
+    // always be the case.
+    if (/\n$/.test(state.input)) {
```
cspotcode, line 224: A detailed explanation within the code!  Love it; thanks for this.

cspotcode:
Thanks for sending this.

I haven't looked at this code in a bit, but I noticed coverage of one line being removed by this change.

https://app.codecov.io/gh/TypeStrong/ts-node/compare/1480/changes#D1L605

Is it possible that the line in question is *trying* to do the same thing as this PR, but not doing it well enough?  Again, I haven't looked at this code in a bit, so I don't know off-hand.

cspotcode:
Also, we should add a test for this, since none of our tests currently are catching the bug being fixed here.

I recently made an effort to make our tests more approachable by splitting them into several files.  Not perfect, still a work-in-progress.  But REPL tests are here:
https://github.com/TypeStrong/ts-node/blob/main/src/test/repl/repl.spec.ts

TheUnlocked:
Looks like that lost line of coverage was from the code attempting to fix this same sort of issue, but it seems like the motivation was probably slightly different at the time and so it had an overly restrictive condition. Since the new fix appears to subsume whatever that line did, I've removed that code.

cspotcode:
Ok sounds good, thanks.  My only remaining concern:

The old code was inside `appendToEvalState`, which is called in 3 different places.  Your code is only used in 1, and I'm wondering if those 2 other places also need this semicolon insertion.

For example, I think the `.type` command might benefit from this as well.

If what I'm saying makes sense, can we move your code to where the old code was?

TheUnlocked:
Moving the code to `appendToEvalState` causes regressions in other tests, particularly surrounding things like
```
[
    1,
    2,
    3
]
```
Looking more closely at the code, it seems like it was trying to do this fix for things like `()` and `[]` (to prevent function invocation and indexing) and actually covered subtraction and division, but it didn't capture addition and multiplication for some reason (possibly because `* x` isn't valid syntax, and maybe whoever wrote it just forgot that `+ x` is).

Adding `+` and `*` to the regular expression without any of the added code fixes the issue (sort of, there was also a bug with that code where the control flow was slightly wrong, but it works when I fix that), but it also means that if any new operators are added to Javascript in the future, they would need to be explicitly opted in as well. Automatically inserting a semicolon at the end of every completed input results in plenty of not-strictly-necessary semicolons at the end of inputs, but those don't hurt anything, and it's less prone to failure later on.

The other two situations that use `appendToEvalState` are:
* `.type`, where it could previously insert a semicolon on the previous line before checking the type. Since a semicolon will always be inserted on the previous line in the new system anyways, and since all input changes from a `.type` command are reverted after it finishes running (so there's no need to add a semicolon after it), having semicolon insertion for that function call is unnecessary with the new code.
* As a shortcut to get an undo function to reset the REPL state to the beginning. Inserting a semicolon there is unnecessary (even if there is data in the `state`, the `start` function adds some code itself which seems to avoid the issue... I couldn't figure out any way to break it even with manually setting the state before starting the REPL).

There was an issue where `evalCode` didn't call `evalCodeInternal` and so wouldn't hit the fix, so I moved the fix into `appendCompileAndEvalInput` itself, and it seems to be working everywhere now as far as I can tell.

cspotcode:
Thanks for the details analysis.  Sounds good to me.

The only thing I changed:
I want to be paranoid about avoiding unnecessary semicolons, so I tweaked the logic so it only adds a semicolon if one isn't already there.  Seems double-semicolons compile as an extra blank line which can affect stack trace line numbers.  Probably not an issue for anyone, but I was able to make a test case for it, and it works.

PR #1483, opened 2021-10-01
fix: Filter empty strings out of list options

### LINKED ISSUE #1301: Empty TS_NODE_IGNORE defaults to `node_modules/`
### Expected Behavior

Running `TS_NODE_IGNORE= webpack` should make ts-node (internally used by webpack) take into account references to TypeScript files contained in `node_modules/`.

### Actual Behavior

It behaves the same as if `TS_NODE_IGNORE` was not defined at all. When setting `TS_NODE_IGNORE` to a dummy value, e.g. `TS_NODE_IGNORE=/non-existent-directory webpack`, things work as expected.

Clone https://github.com/inad9300/webpack-typescript-loader/tree/ts-node-ignore, run `npm i` and `npm start`. Compare this with `npm run really-start`.

### Specifications

* ts-node version: 9.1.1
* node version: 14.16.0
* TypeScript version: 4.2.4
```
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "moduleResolution": "node"
  }
}
```
* Operating system and version: latest macOS

ValeriaVG:
Filters out empty strings from `split` list env variables, allowing `TS_NODE_IGNORE=  <...>` to be treated as an empty array.

Closes #1301

### DIFF
--- a/src/test/index.spec.ts
+++ b/src/test/index.spec.ts
@@ -602,6 +602,21 @@ test.suite('ts-node', (test) => {
         ]);
       });
 
+      test('should ignore empty strings in the array options', async () => {
+        const { err, stdout } = await exec(
+          `${BIN_EXEC} tsconfig-options/log-options1.js`,
+          {
+            env: {
+              ...process.env,
+              TS_NODE_IGNORE: '',
+            },
+          }
+        );
+        expect(err).toBe(null);
+        const { options } = JSON.parse(stdout);
+        expect(options.ignore).toEqual([]);
+      });
+
       test('should have flags override / merge with `tsconfig.json`', async () => {
         const { err, stdout } = await exec(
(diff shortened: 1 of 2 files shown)


cspotcode:
Thanks, where does this `split` function get used?  Is it *only* used to split environment variables?  Just want to be sure we don't break anything else.

Any chance you can tweak the JSDoc comment a bit to describe what it does, or at least explain that it's meant for env var parsing?  Looks like it does some whitespace removal too.  It might be nice to further clarify its intended usage, now that it's doing even more than a string split.

ValeriaVG:
@cspotcode Yes, it's imported only in `src/index.ts` and used only here:
```ts
 ignore: split(env.TS_NODE_IGNORE),
 // ...
 ignoreDiagnostics: split(env.TS_NODE_IGNORE_DIAGNOSTICS),
```

cspotcode (approved):
Approved.  Pretty sure the test failures are some unrelated test flakiness that I'll address somewhere else.

cspotcode:
@ValeriaVG thanks again!

PR #1491, opened 2021-10-07
Add traceResolution support

### LINKED ISSUE #1128: support for traceResolution
### Desired Behavior

Would be great if `ts-node` would support tsconfig's `traceResolution` and provide resolution repport

Kind of: I cannot trace and debug the mapping and resolution problems

TheUnlocked:
Resolves #1128.

Something to note is that in this implementation, I've introduced some "warm-up" code which runs the compiler on empty input when you start the REPL so that it will spit out all of the trace messages before it prompts (otherwise it would show all of the trace messages only after you submit the first prompt which would feel very weird). Currently the warm-up code only runs if `traceResolution` is enabled since otherwise it's not necessary, but a secondary advantage it has is that while it takes slightly longer to start up the `ts-node` REPL when it runs, the first time you submit a prompt it runs immediately instead of waiting a bit for the compiler to warm up (since it's already compiled code before). I'm wondering if it would be a good idea to always run the warm-up code even when `traceResolution` is not enabled so that the REPL feels a bit snappier once it finishes loading.


cspotcode:
> I'm wondering if it would be a good idea to always run the warm-up code even when traceResolution is not enabled so that the REPL feels a bit snappier once it finishes loading.

Yeah I agree.  It will help with #1114, where the first time you used tab-completion it would be slow because it was initializing the typechecker.

Can we extract that "warmup" change into a separate pull request?

TheUnlocked:
Sure. I've opened #1498.

PR #1498, opened 2021-10-10
Invoke TS compiler before starting REPL prompt

TheUnlocked:
Following up discussion from #1491

Gets the typescript compiler warmed up by invoking it before the REPL prompt appears. Slightly increases time to prompt but improves first tab completion/input performance.

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -329,6 +329,11 @@ export function createRepl(options: CreateReplOptions = {}) {
       }
     }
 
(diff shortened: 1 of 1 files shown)


cspotcode:
Thanks again for these.  Are you doing Hacktoberfest?  I ask because I may be distracted this month adding support for TS4.5's big new features, but I also want to make sure anyone doing Hacktoberfest gets credit for their contributions.  So if you're doing that, I'll be sure to meet whatever deadlines are necessary in reviewing these PRs.

Issue #1510, opened 2021-10-12
Do our tests need to detect new "module" options?


cspotcode:
TS 4.5 introduces --module es2022

Is there anywhere in the code where we check module configs?  Or hardcode module enum values?  If so, we should add tests that break whenever TS adds new module options we have not accounted for.  That way our nightly tests will warn us when we may need code changes.

---

We need to prove that swc `ModuleKind` mapping includes relevant values, and that swc properly classifies modules in `node12` and `nodenext` mode.

PR #1521, opened 2021-10-18
add Target ES2021 mapping in transpilers/swc

safareli:
I was getting error: `jsc.target should be es5 or upper to use getter / setter` 
and after investigation I've isolated issue and here is the "fix"

### DIFF
--- a/src/test/transpilers.spec.ts
+++ b/src/test/transpilers.spec.ts
@@ -3,24 +3,25 @@
 // Should consolidate them here.
 
 import { context } from './testlib';
-import {
-  contextTsNodeUnderTest,
-  testsDirRequire,
-} from './helpers';
+import { contextTsNodeUnderTest, testsDirRequire } from './helpers';
 import * as expect from 'expect';
 
 const test = context(contextTsNodeUnderTest);
 
 test.suite('swc', (test) => {
   test('verify that TS->SWC target mappings suppport all possible values from both TS and SWC', async (t) => {
-    const swcTranspiler = testsDirRequire('ts-node/transpilers/swc-experimental') as typeof import('../transpilers/swc');
+    const swcTranspiler = testsDirRequire(
+      'ts-node/transpilers/swc-experimental'
+    ) as typeof import('../transpilers/swc');
 
     // Detect when mapping is missing any ts.ScriptTargets
     const ts = testsDirRequire('typescript') as typeof import('typescript');
-    for(const key of Object.keys(ts.ScriptTarget)) {
-      if(/^\d+$/.test(key)) continue;
-      if(key === 'JSON') continue;
-      expect(swcTranspiler.targetMapping.has(ts.ScriptTarget[key as any] as any)).toBe(true);
+    for (const key of Object.keys(ts.ScriptTarget)) {
+      if (/^\d+$/.test(key)) continue;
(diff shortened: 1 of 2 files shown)


cspotcode:
Linking to #1510 since we may need to apply the same concept to `target` options.  We may also want to add some `>=` logic so that unrecognized target values translate to the newest swc target we know about.

safareli:
in my tsconfig.json I have:
```ts
{
  "compilerOptions": {
    "target": "ES2021",
    ...
 },
  "ts-node": {
    "require": ["tsconfig-paths/register"],
    "files": true,
    "transpileOnly": true,
    "transpiler": "ts-node/transpilers/swc-experimental"
  }
}
```

then I was requiring a module from local module (I have monorepos setup).
in `package.json` of the module that I required there is:
```
{
  ...
  "files": [
    "build"
  ],
  ....
}
```
so that when this module is required build js and d.ts files are used.

as target for my tsconfig was es2021 in compiled js the use of property `get` syntax was not de-suggered.
I had something like 
```
class Foo {
  get foo() {
    return "foo"
  }
}
```

and the error is here:

```
$ npm run foo

> foo
> ts-node --project ../../tsconfig.json ./src/foo.ts

error: jsc.target should be es5 or upper to use getter / setter
  --> /Users/safareli/dev/my-project-foo/packages/foo/src/code/index.ts:31:7
   |
31 |   get foo() {
   |       ^^^

/Users/safareli/dev/my-project-foo/node_modules/@swc/core/index.js:135
        return bindings.transformSync(isModule ? JSON.stringify(src) : src, isModule, toBuffer(newOptions));
                        ^
Error: failed to process js file

Caused by:
    failed to parse module: error was recoverable, but proceeding would result in wrong foo
    at Compiler.transformSync (/Users/safareli/dev/my-project-foo/node_modules/@swc/core/index.js:135:25)
    at Object.transformSync (/Users/safareli/dev/my-project-foo/node_modules/@swc/core/index.js:213:21)
    at Object.transpile (/Users/safareli/dev/my-project-foo/node_modules/ts-node/src/transpilers/swc.ts:114:32)
    at /Users/safareli/dev/my-project-foo/node_modules/ts-node/src/index.ts:1142:35
    at Object.compile (/Users/safareli/dev/my-project-foo/node_modules/ts-node/src/index.ts:1185:30)
    at Module.m._compile (/Users/safareli/dev/my-project-foo/node_modules/ts-node/src/index.ts:1309:30)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/safareli/dev/my-project-foo/node_modules/ts-node/src/index.ts:1313:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12) {
  code: 'GenericFailure'
}
npm ERR! Lifecycle script `foo` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: @my-project-foo/baz
npm ERR!   at location: /Users/safareli/dev/my-project-foo/packages/baz
```

cspotcode:
Does this look like a complete list?
https://rustdoc.swc.rs/swc/config/enum.JscTarget.html

PR #1522, opened 2021-10-19
update version checks for v16

shrujalshah28:
Ref: https://github.com/nodejs/node/pull/40504


cspotcode:
Should be 16.12.0, right?  Can we also add 16.11.0 to our test matrix, so that we are certain to test against 2x 16.x versions: one before the loader hooks changed, and one after?

cspotcode:
@shrujalshah28 thank you!

PR #1545, opened 2021-11-17
Passing arguments in shebang on Linux (env -S)

sheeit:
EDIT from @cspotcode: Closes #1448, also give credit to @chee in release notes

---

It is possible to pass arguments into commands in the shebang (at least on GNU/Linux) using the `-S` option for env [0][1].

The README currently says that it's not possible to do this on Linux, which is wrong.

[0] [env(1) manpage in Debian Bullseye](https://manpages.debian.org/bullseye/coreutils/env.1.en.html#S).
[1] [GNU coreutils env documentation](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html).

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,11 +162,17 @@ console.log("Hello, world!")
 ```
 
 Passing CLI arguments via shebang is allowed on Mac but not Linux.  For example, the following will fail on Linux:
+### Passing CLI arguments via shebang
 
-    #!/usr/bin/env ts-node --files
-    // This shebang is not portable.  It only works on Mac
+- On Mac:
 
-Instead, specify all ts-node options in your `tsconfig.json`.
+      #!/usr/bin/env ts-node --files
+
+- On Linux:
+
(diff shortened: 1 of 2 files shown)


cspotcode:
This has come up a few times before.  You can search for those discussions,
PRs, or issues.  We talked about how this is only possible on certain
versions of env, so a docs change will need to mention that.

I'm on mobile and haven't been able to read this PR, so it might already
mention these things.  But just saying right away that it will be important
to talk about the Linux version requirements.

cspotcode:
@sheeit which version of `env` adds the `-S` flag?  We should add mention of this to this PR before merging.

cspotcode:
Another thought: does Mac support `-S`?  I don't have a Mac to test on.  I'm wondering if it is possible to write a cross-platform shebang in this style.

cspotcode:
We may consider also mentioning `--debug` as well, to aid solving problems: `#!/usr/bin/env -S --debug ts-node --flags-go-here`

PR #1557, opened 2021-12-01
Update typescript to 4.5.2

Pokute:
This updates used typescript version to 4.5.2 and typedoc to 0.22.10.
These version bumps are required for [Import assertion support](https://github.com/microsoft/TypeScript/issues/40694).

typedoc@0.22.4 has a peer dependancy of `typescript@"4.0.x || 4.1.x || 4.2.x || 4.3.x || 4.4.x"` so it needs to be updated too with each typescript update since typedoc doesn't support future versions of TS.


cspotcode:
Thanks, why is it necessary to update typedoc?  We should have that information captured in this pull request for future reference.

Just a note that our tests actually run against an array of typescript versions, different than we use to compile and to render docs.  Our test matrix is declared here: https://github.com/TypeStrong/ts-node/blob/3a2848c9968f03f7b05e57e833ca256b52ab4e18/.github/workflows/continuous-integration.yml#L48-L128

Having said that, it still makes sense to frequently upgrade to the latest typescript version for our build, so this is a good change.

Pokute:
> Thanks, why is it necessary to update typedoc? We should have that information captured in this pull request for future reference.

typedoc@0.22.4 has a peer dependancy of typescript@"4.0.x || 4.1.x || 4.2.x || 4.3.x || 4.4.x" so it needs to be updated too.

> Just a note that our tests actually run against an array of typescript versions, different than we use to compile and to render docs.

Oh, never came to mind to check the `.github` directory. I'll check it out.

PR #1558, opened 2021-12-01
Update prettier to 2.5.0

Pokute:
This PR includes #1557 

This updates used prettier version to 2.5.0 and updates all source files from running prettier too.

The version bump of prettier is required for [Import assertion support](https://github.com/microsoft/TypeScript/issues/40694).


cspotcode:
Replaced with #1573.

PR #1562, opened 2021-12-06
fix: ERR_UNKNOWN_FILE_EXTENSION is undefined

bluelovers:
> TypeError: ERR_UNKNOWN_FILE_EXTENSION is not a constructor


cspotcode:
Thank you!  I suppose we should add a quick test for this.  Maybe just add a file that tries to `import './unrecognized.txt';`?

@bluelovers do you have time to add a quick test, or should I do that?

Issue #1414, opened 2021-07-22
TS's new node ESM support


cspotcode:
Will need work on our side to integrate.

https://github.com/microsoft/TypeScript/pull/44501
https://github.com/weswigham/TypeScript/pull/67

## Roughly

adds new `module: node12` and `module: nodeNext` configs
In them:
- existing extensions obey package.json "type" field
- new extensions `.mts`, `.mjs`, `.cts`, `.cjs` force one or the other; ignore package.json "type" field

## Questions I still have

Are any of the above assumed to be modules without needing to add `export {};`?

alvis:
With 4.5 beta announced 2 weeks ago, and the final on the horizon, time for us to do something.

PodaruDragos:
Will this mean that we can now potentially have `import { foo } from './foo'`, basically omitting the extensions just like in `commonJS` ?

cspotcode:
@PodaruDragos you're probably looking for node's `--experimental-specifier-resolution=node` which we already support.  Just as node requires you to pass that flag, we *also* require you to pass that flag, often via the `NODE_OPTIONS` environment variable.

https://nodejs.org/dist/latest-v17.x/docs/api/esm.html#customizing-esm-specifier-resolution-algorithm

PR #1564, opened 2021-12-09
feat: support .mts .cts

bluelovers:
https://github.com/TypeStrong/ts-node/issues/1007

https://github.com/TypeStrong/ts-node/issues/1414

### DIFF
--- a/dist-raw/node-cjs-loader-utils.js
+++ b/dist-raw/node-cjs-loader-utils.js
@@ -23,6 +23,8 @@ function assertScriptCanLoadAsCJSImpl(service, module, filename) {
   const tsNodeClassification = service.moduleTypeClassifier.classifyModule(normalizeSlashes(filename));
   if(tsNodeClassification.moduleType === 'cjs') return;
 
+  // TODO modify to ignore package.json when file extension is ESM-only
+
   // Function require shouldn't be used in ES modules.
   if (tsNodeClassification.moduleType === 'esm' || (pkg && pkg.data && pkg.data.type === 'module')) {
     const parentPath = module.parent && module.parent.filename;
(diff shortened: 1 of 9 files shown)


cspotcode:
Like all PRs, this will need tests before we can merge it.

bluelovers:
i don't know how make test for this

cspotcode:
A proper test needs to use the feature that you're adding.  So if you're adding support for new file extensions, then the test needs to use the file extensions to prove that ts-node behaves the way you want it to behave.

Most of ts-node's tests are very straightforward.  They run the ts-node CLI and tell it to run code in a sample project.  The code that spawns the CLI lives in `./src/test/**.spec.ts`.  The sample project lives in `./tests`

Here's an example you can use as a starting point.
https://github.com/TypeStrong/ts-node/blob/3a2848c9968f03f7b05e57e833ca256b52ab4e18/src/test/index.spec.ts#L580-L584
https://github.com/TypeStrong/ts-node/tree/main/tests/main-realpath

cspotcode:
This is gonna need more work:

* must register .js extension handler even if allowJs is off
  * necessary to compile .cjs, .cts, and to throw ESM error for .mts
* consider conditionally registering .mjs extension when [moduleType overrides](https://typestrong.org/ts-node/docs/module-type-overrides) are set
* modify `assertScriptCanLoadAsCJSImpl` to understand that .mts overrides package.json
* modify ESM loader to resolve from .mjs / .cjs to .mts / .cts, the same way it does for .js/.jsx/.ts/.tsx

And finally...
* test coverage of the above

cspotcode:
Might want to ship with #1361 so that `require("./foo.cjs")` resolves to `require("./foo.cts")`

cspotcode:
I've merged #1614 into this PR, meant to establish a common core for `_resolveFilename` hooks. In particular, I hope it avoids a painful merge with #1585.

cspotcode:
Docs updates to go with this:

Update moduleTypes jsdoc and markdown: they say ts cannot use mts and cts; this is wrong

cspotcode:
Rolling into #1694

PR #1572, opened 2021-12-22
Forward whole context to ESM hooks

geigerzaehler:
Ensure that all properties of the context objects passed to the `resolve` and `load` ESM hooks are forwarded to the default hooks.

This change is necessary to for using [import assertions in Node v17][1]. Node v17 requires these import assertions for JSON modules. If the assertions are not passed down to the default `load` function, Node refuses to import the JSON module.

To prevent similar issues we also forward the context for the `resolve` hook.

I’d love to add some tests for this but I’d need some help as to how to approach
this.

[1]: https://nodejs.org/dist/latest-v17.x/docs/api/esm.html#import-assertions

### DIFF
--- a/src/esm.ts
+++ b/src/esm.ts
@@ -62,17 +62,36 @@ export interface NodeLoaderHooksAPI2 {
 export namespace NodeLoaderHooksAPI2 {
   export type ResolveHook = (
     specifier: string,
-    context: { parentURL: string },
+    context: ResolveContext,
     defaultResolve: ResolveHook
   ) => Promise<{ url: string }>;
   export type LoadHook = (
     url: string,
-    context: { format: NodeLoaderHooksFormat | null | undefined },
+    context: LoadContext,
     defaultLoad: NodeLoaderHooksAPI2['load']
   ) => Promise<{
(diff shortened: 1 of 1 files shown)


cspotcode:
Thanks, is this similar to #1557?  If so, can we collaborate to create a single pull request with all the necessary tests?  There are no hard and fast rules here, as long as we end up with good tests and a good PR that we can merge, everyone will get attribution in the release notes.

#1557 has some tests, but I'm not sure if they are sufficient; I haven't had enough time to review.  If you think it is appropriate to combine these two pull requests, can you offer a code review?  Hopefully @pokute is available to discuss as well.

It's the holiday season for me so my time is limited, but hopefully we can merge and release a bunch of things near the beginning of the year.

geigerzaehler:
> Thanks, is this similar to #1559?

Yes it is and I’ve added my comments to the PR. (The PR somehow escaped my notice). 

My PR also forwards the context for the `resolve()` hook which `#1559` does not. I’ll gladly rebase my PR once `#1559` is merged

cspotcode:
Closing in favor of #1559

PR #1585, opened 2021-12-29
Add path mapping support to ESM and CJS loaders

### LINKED ISSUE #1586: Set correct `code` on Errors returned by ESM resolver
Fixed by #1585 

Creating this issue to ensure this enhancement is mentioned in the appropriate release notes.

geigerzaehler:
The ESM loader now resolves import paths using TypeScripts [path mapping][1] feature.

[1]: https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

EDIT:
Closes #1586


cspotcode:
Thanks for this.  Because of the American holidays, I have somewhat of a backlog of issues and PRs which need attention before I'll be able to properly review this.  There may be some delay.

I see some refactoring in node-errors; can you explain the motivation for some of those changes?  It's good to have an explanation for such changes when reviewing.

Your description says this enables path mapping for the ESM loader.  We'd like to also extend the same functionality to the CommonJS loader.  I see that your `createPathMapper` implementation looks clean and modular; do you anticipate any issues if we attempt to use it for CommonJS as well?  We'd hook into `Module._resolveFilename` for this.

Is this feature-flagged?  We should decide if it's enabled or disabled by default, and decide how users can opt-in or opt-out.

geigerzaehler:
> There may be some delay.

No problem.

> I see some refactoring in node-errors; can you explain the motivation for some of those changes?

The new resolving code needs to detect `ERR_MODULE_NOT_FOUND` errors so it can try more mapped candidates. The builtin `NodeError` from Node’s internal `error` module [has a `code` property](https://github.com/nodejs/node/blob/f81c62704fc4e3e0eb1b1c813854ad52fbb8fe75/lib/internal/errors.js#L9-L10) that allows you to distinguish errors. The change in `node-errors` also sets this property on the error instances. The new logic for defining errors is [similar to the code](https://github.com/nodejs/node/blob/f81c62704fc4e3e0eb1b1c813854ad52fbb8fe75/lib/internal/errors.js#L341-L375) in Node’s `error` module.


> do you anticipate any issues if we attempt to use it for CommonJS as well?

I don’t think there will be any issues. The path mapping code is does not include any logic that is specific to ESM resolution. The only inputs to it are the TypeScript compiler options.

> Is this feature-flagged? We should decide if it's enabled or disabled by default, and decide how users can opt-in or opt-out.

It’s not feature-flagged at the moment because I couldn’t think of a scenario where you would want the path mapping to be disabled. If you have configured the `paths` compiler option and you are using it in your code, your code will only run if the path mapping is enabled. If path mapping is disabled the code will always fail at the module resolution stage which makes the option unusable. So I’d lean towards not feature-flagging this for simplicity. But if there’s a good reason to put this behind a feature flag it should be easy to accomplish

cspotcode:
> If you have configured the paths compiler option and you are using it in your code, your code will only run if the path mapping is enabled.

Good point.  And I think someone could disable it like this:

```
{
  "compilerOptions": {
    "paths": { ... }
  },
  "ts-node": {
    "compilerOptions": {
      "paths": {} // empty object: disable path mappings in ts-node
    }
  }
}
```

src/esm.ts
```diff
@@ -147,12 +149,40 @@ export function createEsmHooks(tsNodeService: Service) {
       return defer();
     }
 
-    // pathname is the path to be resolved
+    let candidateSpecifiers: string[] = [specifier];
+
+    if (context.parentURL) {
+      const parentUrl = new URL(context.parentURL);
+      if (parentUrl.pathname && extname(parentUrl.pathname) === '.ts') {
```
cspotcode, line 156: We'll need to update this to match all parents which should get path mapping, not just `.ts` files.  This potentially includes `tsx`, `mts`, `jsx`, etc.

We can do `if(!tsNodeService.ignored(path)) {...`

Note that will only do mapping if the parent is transformed by us.  For example, given a mix of .ts and .js with allowJs turned off, it will map paths when the parent is .ts but not .js, even if in the same directory.  Should be fine since this PR already excludes .js.

Will also need to handle cases where parentUrl is a non-file URL, just to avoid errors.

src/esm.ts
```diff
@@ -93,6 +94,7 @@ export function registerAndCreateEsmHooks(opts?: RegisterOptions) {
 
 export function createEsmHooks(tsNodeService: Service) {
   tsNodeService.enableExperimentalEsmLoaderInterop();
+  const mapPath = createPathMapper(tsNodeService.config.options);
```
cspotcode, line 97: We can move creation of this into `create()` and store it on the service as `tsNodeService.pathMapper` since it'll be used for CommonJS, too.

src/esm.ts
```diff
@@ -147,12 +149,40 @@ export function createEsmHooks(tsNodeService: Service) {
...
+        if (mappedSpecifiers) {
+          candidateSpecifiers = mappedSpecifiers.map((path) =>
+            pathToFileURL(path).toString()
+          );
+        }
+      }
+    }
+
+    let candidateSpecifier: string | undefined;
+    while ((candidateSpecifier = candidateSpecifiers.shift())) {
```
cspotcode, line 167: Personal nit: prefer `for()` loop to mutating array.

src/esm.ts
```diff
@@ -147,12 +149,40 @@ export function createEsmHooks(tsNodeService: Service) {
...
+      try {
+        return await nodeResolveImplementation.defaultResolve(
+          candidateSpecifier,
+          context,
+          defaultResolve
+        );
+      } catch (err) {
+        const isNotFoundError = (<any>err).code === 'ERR_MODULE_NOT_FOUND';
```
cspotcode, line 175: Nit: prefer `as any`

src/esm.ts
```diff
@@ -147,12 +149,40 @@ export function createEsmHooks(tsNodeService: Service) {
...
+        if (isNotFoundError && candidateSpecifiers.length > 0) {
+          continue;
+        } else {
+          throw err;
```
cspotcode, line 179: I wonder, should we generate a new kind of error indicating that we tried many candidates?  `err` talks specifically about the *last* candidate which was tried.  Will that be confusing to the user?

In the following example, assume `config` maps to both `src/config` and `src/codegen/config`.
The user has a typo in their specifier, but the error makes them think it has skipped over the path mapping to `src/config`:

```
$ node
> import('config/defalts')
Uncaught:
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '/d/ts-node/path-mappings/src/codegen/config/defalts' imported from /d/ts-node/path-mapping/repl
```

An example of the style of message we might return instead:

```
Uncaught:
Error [ERR_MODULE_NOT_FOUND]: Cannot find module for specifier 'config/defalts' imported from /d/ts-node/path-mapping/repl using TypeScript path mappings
Candidate specifiers attempted:
- /d/ts-node/path-mappings/src/config/defalts
- /d/ts-node/path-mappings/src/codegen/config/defalts
```

This error makes it clear that `src/config` was attempted.

src/esm.ts
```diff
@@ -147,12 +149,40 @@ export function createEsmHooks(tsNodeService: Service) {
...
+        }
+      }
+    }
 
-    return nodeResolveImplementation.defaultResolve(
-      specifier,
-      context,
-      defaultResolve
+    throw new Error(
+      `Empty mapped paths for ${specifier} imported from ${context.parentURL}`
```
cspotcode, line 185: Nit: add a code comment explaining it is a bug to ever get here.

src/path-mapping.ts
```diff
@@ -0,0 +1,131 @@
...
+// Path mapper returns a list of mapped specifiers or `null` if the
+// given `specifier` was not mapped.
+type PathMapper = (specifier: string) => string[] | null;
+
+export function createPathMapper(
+  compilerOptions: ts.CompilerOptions
+): PathMapper {
+  if (compilerOptions.paths) {
+    if (!compilerOptions.baseUrl) {
+      throw new Error(`Compiler option 'baseUrl' required when 'paths' is set`);
```
cspotcode, line 13: Is `compilerOptions.baseUrl` guaranteed to be an absolute path, or do we need to resolve it relative to the tsconfig's location?

tests/esm-path-mapping/index.ts
```diff
@@ -0,0 +1,32 @@
+import * as assert from 'assert';
```
cspotcode, line 1: We can made a CJS version of `esm-path-mapping` at `cjs-path-mapping` to test that this also works for CJS.

cspotcode (changes requested):
I left comments from an initial code review.  I plan to implement the requested changes myself, but if you are feeling enthusiastic and would like to do them before me, I will certainly appreciate the assistance.  No pressure though.

Overall this looks great to me.  I definitely want to add CommonJS support which shouldn't be too difficult.  I'm already familiar with the hooking API.

src/path-mapping.ts
geigerzaehler: Good point. I assumed that we always get the configuration from [`ts.readConfigFile()` in `src/configuration.ts`](https://github.com/TypeStrong/ts-node/blob/8949654f2b7050b2b692eefe14b225d3f6e7901c/src/configuration.ts#L113) and the function resolves the base URL relative to the config file. But there are probably other ways to set the base URL like through the command line, API, or environment.

To address this I think it make sense to resolve any base URL that is not a URL but a relative path to the current directory.

src/esm.ts
geigerzaehler: https://github.com/TypeStrong/ts-node/pull/1585/commits/2fcb9821593fd9235f45585ed0c95b1943a9c3de

src/esm.ts
geigerzaehler: https://github.com/TypeStrong/ts-node/pull/1585/commits/2fcb9821593fd9235f45585ed0c95b1943a9c3de

src/esm.ts
geigerzaehler: https://github.com/TypeStrong/ts-node/pull/1585/commits/2fcb9821593fd9235f45585ed0c95b1943a9c3de

src/esm.ts
geigerzaehler: https://github.com/TypeStrong/ts-node/pull/1585/commits/2fcb9821593fd9235f45585ed0c95b1943a9c3de

src/esm.ts
geigerzaehler: I’ve updated the code accordingly in https://github.com/TypeStrong/ts-node/pull/1585/commits/48f5262e64adaa270b16a309011857a9722bf2fd

src/esm.ts
geigerzaehler: > I wonder, should we generate a new kind of error indicating that we tried many candidates? err talks specifically about the last candidate which was tried. Will that be confusing to the user?

That’s a good point and I like your suggestion of providing more information. I implemented this in https://github.com/TypeStrong/ts-node/pull/1585/commits/a19454b107bf824a4324824251545595ee7d7d9b

geigerzaehler:
> I definitely want to add CommonJS support which shouldn't be too difficult. I'm already familiar with the hooking API.

I’d prefer to keep scope of this PR small and I’m not familiar enough with the CommonJS side of things so I’ll leave this to you.

One thing to consider is that we’d probably want the users to explicitly opt in to path mapping for CommonJS. They’ll be using `tsconfig-paths` at the moment and it is unclear if the `ts-node` implementation can be used side-by-side or compatible.

cspotcode:
Fair point.  Would it be a great hardship for those users to disable
tsconfig-paths?  I hope they would appreciate the simplicity of removing a
component.  I would like our default behaviour to be the best
forward-thinking behaviour, but perhaps we add an optional flag which can
be used to selectively disable it for commonjs?  Or we somehow detect
tsconfig-paths and throw a warning when both mappers are competing?

src/esm.ts
```diff
@@ -175,26 +176,33 @@ export function createEsmHooks(tsNodeService: Service) {
...
         );
       } catch (err) {
-        const isNotFoundError = (<any>err).code === 'ERR_MODULE_NOT_FOUND';
-        if (isNotFoundError && candidateSpecifiers.length > 0) {
-          continue;
-        } else {
+        const isNotFoundError = (err as any).code === 'ERR_MODULE_NOT_FOUND';
+        if (!isNotFoundError) {
           throw err;
+        } else if (i == candidateSpecifiers.length - 1) {
```
cspotcode, line 190: Nit: use `===`

src/esm.ts
```diff
@@ -175,26 +176,33 @@ export function createEsmHooks(tsNodeService: Service) {
...
+          throw new MappedModuleNotFound(
+            specifier,
+            context.parentURL,
+            candidateSpecifiers
+          );
```
cspotcode, line 195: This looks great.  Should we special-case it to throw `err` when `candidateSpecifiers.length === 1`?

src/esm.ts
```diff
@@ -175,26 +176,33 @@ export function createEsmHooks(tsNodeService: Service) {
...
+          continue;
         }
       }
     }
 
+    // This code should be unreachable: The for-loop always returns or
+    // throws.
     throw new Error(
-      `Empty mapped paths for ${specifier} imported from ${context.parentURL}`
+      `Unreachable code mapping ${specifier} in ${context.parentURL}`
```
cspotcode, line 205: Thanks, future me appreciates the clarity here!

src/esm.ts
```diff
@@ -332,3 +340,19 @@ export function createEsmHooks(tsNodeService: Service) {
...
+    super(
+      [
+        `Cannot find '${specifier}' imported from ${base} using TypeScript path mapping`,
+        'Candidates attempted:',
+        ...candidates.map((candidate) => `- ${candidate}`),
+      ].join('\n')
+    );
+    this.name = `Error [${MappedModuleNotFound.code}]`;
+  }
+}
```
cspotcode, line 358: Is there any benefit to setting a ts-node specific `code` on these errors?  `"ERR_TS_NODE_MAPPED_MODULE_NOT_FOUND"`?  Or maybe make this a subclass of `AggregateError`?

We don't need to do this now.  I'll keep it in mind for the future.

src/index.ts
```diff
@@ -482,6 +483,14 @@ export interface Service {
...
   /** @internal */
   transpileOnly: boolean;
+  /**
+   * @internal
+   *
+   * Map import paths to candidates according to the `paths` compiler
+   * option. Returns `null` if the specifier did not match and was not
+   * mapped.
+   */
+  mapPath(specifier: string): string[] | null;
```
cspotcode, line 493: Looks great, thanks!

src/esm.ts
cspotcode: I realized that node itself throws multiple flavors of `ERR_MODULE_NOT_FOUND` with different phrasing for the error message.  So I think it would be reasonable for us to use the same error code for the errors we throw.  We can also set `.errors` to an array of the individual errors to match the shape of `AggregateError` instances.

I'm happy to implement this change.

src/path-mapping.ts
cspotcode: I see, we almost always get the configuration from `ts.readConfigFile()`, so we should be good.  I think the only way to set it otherwise is via the API?  Since we don't support all of `tsc`'s command-line flags.

Maybe the correct approach is to resolve `baseUrl` once within `create()` so that we have a reliable, absolute `baseUrl` to be used elsewhere.

src/esm.ts
cspotcode: ...and I just realized you already have the `code` there, nice.  I tweaked it to set `code` on the error instances, too.

cspotcode:
I made an attempt at adding the CommonJS resolver hook.  Still some outstanding TODOs that I need to address, but this is looking good.

cspotcode:
Looks like we have test failures on Windows due to module specifiers using forward slashes and windows paths using backslashes.  I haven't checked the code yet, but we probably have some logic assuming a module specifier and a path will always look the same.  Should be an easy fix.

geigerzaehler:
> Would it be a great hardship for those users to disable tsconfig-paths?

I don’t think it would. I just wanted to point out, that this is a breaking change. I do believe most users will appreciate this replacement.

Considering the release strategy it probably makes sense to have the commonjs mapping as opt-in and ask users to enable it to collect feedback. Once the feature is stable it can be enabled by default (or always) in a new major release.

> Or we somehow detect tsconfig-paths and throw a warning when both mappers are competing?

This sounds complex and error-prone. I think I’d prefer a solution where it is explicitly configured

src/path-mapping.ts
geigerzaehler: > Maybe the correct approach is to resolve baseUrl once within create() so that we have a reliable, absolute baseUrl to be used elsewhere.

Alternatively, we could require `baseUrl` to be absolute and throw an error otherwise. This would allow us to avoid implicit, environment-dependent behavior for users of the API. This might prevent some confusion when path mapping doesn’t work as expected but only fails at the resolution stage instead of the construction stage. (For example a user may think the base URL they set on the API is considered relative to `tsconfig.json` and their scripts work whenever they are run from the project root. But then they run the script from a subdirectory and it fails.)

cspotcode:
I like that plan: opt-in via a flag today, enable by default in the next major version, get feedback in the meantime.

How about `experimentalPathMapping: 'both' | 'cjs' | 'esm' | 'none'`?  Where `'esm'` is the default today, and 'both' is the default in the next major version, with the flag renamed `pathMapping`?

cspotcode:
I see that tsconfig-paths does some special-casing for node's built-in modules.  I'm not sure if we need to do the same.
https://github.com/dividab/tsconfig-paths/blob/0b259d4cf6cffbc03ad362cfc6bb129d040375b7/src/register.ts#L80-L88

Given the following admittedly contrived tsconfig;

```
"paths": {
  "*": ["src/*", "*"]
```

How should we resolve:

```
import fs from 'fs';
import lodash from 'lodash';
```

I'll do some testing with `tsc --traceResolutions`

cspotcode:
What if someone is using paths to declare types via `.d.ts` files?  For example they want `import "colors"` to resolve to `src/types/colors.d.ts` but they want node to execute `colors` from node_modules?

cspotcode:
I think we'll need to add a fallback so that when mappings fail, the bare specifier is resolved, too.  This matches tsc's behavior.

cspotcode:
We also need to handle when `baseUrl` is set and `paths` is not.  It appears to be equivalent to `paths: {"*": ["*"]}`

src/path-mapping.ts
cspotcode: Sounds good, let's do that.  In `create()` we throw an error if baseUrl is not absolute.

cspotcode: Actually, scratch that, seems we always pass options through TypeScript's API to be normalized, even when provided via our API:

```
require('ts-node').create({compilerOptions: {baseUrl: "foobar"}}).config.options.baseUrl
```

So I guess we're all good.

cspotcode (commented):
Summarizing the work still to be done:

* Add test when `baseUrl` is set and `paths` is omitted
* Add test when a specifier matches paths but does not get resolved by them
  - for example, `paths: {"*": ["src/*"]}` and `import "lodash"`
* Add test importing native modules when they match a path mapping
  - for example, same path config as above and `import "fs"`
* Add test for when mapper resolves to a .d.ts but should skip that and find a .js instead.
  - for example, `{paths: {"*": ["src/bundled-types/*"}}` with `src/bundled-types/ckeditor.d.ts`
* consider adding license note about the code being inspired by tsconfig-paths

geigerzaehler:
> How about `experimentalPathMapping: 'both' | 'cjs' | 'esm' | 'none'`? Where `'esm'` is the default today, and 'both' is the default in the next major version, with the flag renamed to `pathMapping`?

I like it!

> I see that tsconfig-paths does some special-casing for node's built-in modules. I'm not sure if we need to do the same.

In my opinion it’s preferable to keep the mapping algorithm as simple as possible. This means less surprises for the user and makes fewer assumptions on how ts-node is used. Considering the ability to compose multiple ESM loaders means that any special cases may interfere with other loaders and have unintended effects.

In the particular case of preserving builtin node modules I’m not sure that even makes sense. If we remap `fs`, for example, the type checker will use the mapped file for type checking. So `ts-node` should also run the mapped file.



> What if someone is using paths to declare types via `.d.ts` files? For example they want `import "colors"` to resolve to `src/types/colors.d.ts` but they want node to execute `colors` from node_modules?

That’s an interesting case. There are other ways to achieve this, for example the [`typeRoots` setting](https://www.typescriptlang.org/tsconfig/#typeRoots) and [type references](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-). Using the same rationale (simplicity and predictability) as above I’d argue that users use one of the alternatives instead of baking this into the mapping algorithm.


> We also need to handle when `baseUrl` is set and `paths` is not. It appears to be equivalent to `paths: {"*": ["*"]}`
> 
> EDIT: is this `"*": ["*"]` always implied, even when the user specifies a `paths` object?

Wow. Great find. I tested the last question and the mapping `* -> *` is not implied if the paths object is set.

cspotcode:
I found why tsconfig-paths prefers node's builtins: https://github.com/dividab/tsconfig-paths/pull/60

I also did some tests.

* the `"*": ["*"]` mapping is implied when the paths object is omitted *or* when the paths object has no `"*"` key
* `@types/node` builtin declarations are preferred over mapped modules.  This matches tsconfig-paths' preference for node builtins

Here are the tests: https://github.com/TypeStrong/ts-node-repros/tree/ts-path-mapping-tests
Let me know if you get different results

cspotcode:
I've merged #1614 into this PR, meant to establish a common core for `_resolveFilename` hooks.

The findings from https://github.com/TypeStrong/ts-node/pull/1585#issuecomment-1022241341 still need to be addressed.

geigerzaehler:
> * `@types/node` builtin declarations are preferred over mapped modules.  This matches tsconfig-paths' preference for node builtins

This behavior extends to all modules with [ambient declarations](https://www.typescriptlang.org/docs/handbook/modules.html#ambient-modules). For example

```typescript
// ambient.d.ts
declare module "foo" {
  export {}
}
```

```typescript
// index.d.ts
/// <reference path="./ambient.d.ts" />
import foo from "foo"
foo()
```

```typescript
// src/foo.ts
export default function () {}
```

Even if `foo` is mapped to `src/foo.ts` the type checker uses the declaration of module `foo` in `ambient.d.ts`.

If we don’t remap node modules as a special, case we create inconsistent behavior.

I think ideally we want a behavior where the modules used at runtime are the same as the modules resolved at type check-time. This is impossible because we don’t know where the module code for a ambient declaration is by the nature of the declaration. I’d also consider the case where path is resolved to a module that has a ambient declaration otherwise as uncommon.

With this in mind I’d suggest not treat remapping uniformly and not do any special casing.

cspotcode:
I think we still need to avoid this bug?
https://github.com/dividab/tsconfig-paths/issues/56

geigerzaehler:
> I think we still need to avoid this bug?
> [dividab/tsconfig-paths#56](https://github.com/dividab/tsconfig-paths/issues/56)

I‘m unsure if I’d consider the issue a bug. The issue boils down to: `querystring` references the core module when not transpiled and references the installed module when transpiled with the path mapping `* -> node_modules/*`. This seems like expected behavior to me.

We should also keep in mind that with ESM imports we have more workarounds for these kind of edge cases. We can always explicitly refrence a node core module with `node:querystring`, for example.

cspotcode:
> I‘m unsure if I’d consider the issue a bug. The issue boils down to: querystring references the core module when not transpiled and references the installed module when transpiled with the path mapping * -> node_modules/*. This seems like expected behavior to me.

I see what you're saying.  And there should be no need for the `node_modules/*` mapping.  If path mapping fails, then it should fallback to node's resolver which should look at node built-ins first, then `node_modules`.

Users hit performance issues with this:
https://github.com/dividab/tsconfig-paths/issues/56#issuecomment-416596613
https://github.com/dividab/tsconfig-paths/pull/60#issue-355280581
> Additionally this increases performance by avoiding unnecessary disk access. For example webpack calls require("crypto") in a loop.

However, if we scope the path mapping so that it only applies to the project's own code, not to `node_modules` dependencies, then this should be mitigated.  If `node_modules/foo/index.js` does `import "express"` then we don't want that redirecting to `./src/express.ts`  Or if webpack calls `require('crypto')` in a loop, our path mappings should not apply.

I think the logic's already there, but we'll need tests to cover these cases.
- test that path mapping does not affect require calls coming from `node_modules/*/*`
- test that we fallback to node's resolver when path mapping fails (if '*'->'src/*' and `import "lodash"`, should fallback to finding `node_modules/lodash`)

cspotcode (changes requested):
Summarizing the tests that need to be added:

- test when `baseUrl` is set and `paths` is omitted
  - should still map `import "foo"` to `${baseUrl}/foo`
- test when `baseUrl` is set and `paths` maps `"*": ["lib/*"]`
  - `import "foo"` should *not* attempt `${baseUrl}/foo`; should only attempt `${baseUrl}/lib/foo`
- test when specifier matches a paths pattern but must fallback to node_modules
  - for example, `paths: {"*": ["src/*"]}`, `import "lodash"` should fallback to `node_modules/lodash`
- test when specifier matches a paths pattern but must fallback to node built-in module
  - same as above but for `import "fs"`
- When mapping candidate resolves to a `.d.ts`, resolution should be skipped; try next one
- test that external modules (`node_modules/*`) are unaffected by mappings
  - if path mappings map "lodash" to "./src/lodash" and `node_modules/external-library/index.js` does `import "lodash"`, it should *not* get `./src/lodash`
- test that mapping does not apply to relative imports
  - `import "./foo"` should not be mapped even if `"*"` mapping exists
  - `import "/absolute/path/foo"` should not be mapped even if `"*"` mapping exists

cspotcode:
Any update on this?  Any thoughts on the proposed test coverage?  I'm wondering if this will make it in for v10.6.0

charles-allen:
> Any update on this? Any thoughts on the proposed test coverage? I'm wondering if this will make it in for v10.6.0

Are you still looking for someone to add some tests? I have some time today :)

cspotcode:
@charles-allen any luck?  Anything I can help with?  I'm always looking for excuses to improve [`CONTRIBUTING.md`](https://github.com/TypeStrong/ts-node/blob/main/CONTRIBUTING.md)

charles-allen:
> @charles-allen any luck? Anything I can help with? I'm always looking for excuses to improve [`CONTRIBUTING.md`](https://github.com/TypeStrong/ts-node/blob/main/CONTRIBUTING.md)

I forked the PR, cloned, ran the tests (3 are failing already?), and started looking at the code. That was all pretty straight-forward; no issues so far. I can see some new tests in the diff -- I assume the tests I need to write will be similar to those. Mostly I just need to squeeze out a little more time; I really want to get it done this week, with sooner === better!

I'll try to get one test written & pushed tomorrow so I can share my repo URL here for early feedback

cspotcode:
Sounds great, thanks.  I listed some of the tests we need here: https://github.com/TypeStrong/ts-node/pull/1585#pullrequestreview-869432652 (I forget if I already linked to this comment or not)

> ran the tests (3 are failing already?)

Yeah, I forget exactly why, but that is expected.  The new tests will also fail, but once we have them, it'll be that much easier to finish the implementation.

> I can see some new tests in the diff -- I assume the tests I need to write will be similar to those.

Yeah, this style of test seems good for simplicity, since each is basically its own, mini-project.

charles-allen:
Quick update -- I didn't do much Wednesday; but I have blocked the whole day tomorrow to work on this :)

charles-allen:
@cspotcode - I've pushed an attempt at the first 4 tests to my fork:
https://github.com/charles-allen-forks/ts-node/tree/path-mapping
https://github.com/charles-allen-forks/ts-node/commits/path-mapping

So far I have 2 questions:
- I think `exec` is not working correctly in my tests, but I can't work out what I've done differently to the 2 existing ESM tests. Any ideas?
- You listed ~7 tests; but I should replicate these for both ESM & CJS cases, right? So 14 tests total?

I plan to continue tomorrow or Sunday -- still hoping to wrap up the tests this week!

cspotcode:
> I think exec is not working correctly in my tests, but I can't work out what I've done differently to the 2 existing ESM tests. Any ideas?

Hmm, I'm not sure.  What's going wrong?

> You listed ~7 tests; but I should replicate these for both ESM & CJS cases, right? So 14 tests total?

Ideally yeah.  But if you want, and you think it makes sense, I'm open to use some clever ways to automate the ESM & CJS testing so that you can use one set of test files to test both permutations.

cspotcode:
@charles-allen I think I may have found the cause of those failures.

In [src/test/helpers.ts](https://github.com/charles-allen-forks/ts-node/compare/path-mapping~10...path-mapping#diff-bdb8eeb768e3e34c3fe09a39d68b1d729a9082a682f2d8b0e30d7731cc70557c) on line 38 a space has been removed.

cspotcode:
@charles-allen I think it might have to do with the `env`.  I see elsewhere, we pass env like this:

```
env: {
  ...process.env,
  TS_NODE_PROJECT: `./path/to/tsconfig.json`,
},
```

Perhaps if you omit `...process.env`, it messes up the $PATH env var and so exec is unable to find whatever commands it needs to run.

charles-allen:
> Ideally yeah. But if you want, and you think it makes sense, I'm open to use some clever ways to automate the ESM & CJS testing so that you can use one set of test files to test both permutations.

I think to get started it's better to have specific tests for each, given CJS has different import syntax support. It's likely more pain than benefit to factor out similarities & it's no problem for me to recreate a similar set of tests.

> @charles-allen I think it might have to do with the `env`. I see elsewhere, we pass env like this:
> 
> ```
> env: {
>   ...process.env,
>   TS_NODE_PROJECT: `./path/to/tsconfig.json`,
> },
> ```
> 
> Perhaps if you omit `...process.env`, it messes up the $PATH env var and so exec is unable to find whatever commands it needs to run.

That makes a lot of sense! I hadn't considered that I was overwriting the env. I assumed I was adding to it. I'll give that a go and get back to you.

cspotcode:
> I assumed I was adding to it.

I thought the same to be honest, but I guess not, and I had forgotten.

cspotcode:
@charles-allen I'm back at my computer after a weekend away.  Any luck trying `...process.env`?

charles-allen:
> @charles-allen I'm back at my computer after a weekend away. Any luck trying `...process.env`?

Yes, I think that fixed it!

I've added the remaining tests for ESM & pushed. I still need to add the equivalent for CJS. 

I think I will refactor a little first; I think it's cleaner / easier to read if tests with different tsconfigs are different folders. Having them all together couples the tests & makes the combined example imports more complex.

charles-allen:
Maybe we could start looking at fixing the tests, since that will give me more confidence that the tests are correct

cspotcode:
> Maybe we could start looking at fixing the tests, since that will give me more confidence that the tests are correct

Good idea, I'll take a look after my workday and either push fixes to your branch, or update here with any blockers I hit.  You can expect to see that update from me in about 10 hours.

charles-allen:
Sure. Tell me if I need to PR my updates. I've not dealt with forking a fork before!

cspotcode:
> Sure. Tell me if I need to PR my updates. I've not dealt with forking a fork before!

Sounds good.  I feel confident that I can get your changes into this PR without any intervention from you, and then you should get credit as a collaborator on this PR.

geigerzaehler:
Thanks for taking this over @charles-allen! I had a look at your changes and they look great. I’ve updated my branch to incorporate them. I also added you as a collaborator to my repo so you should be able to push to that branch.

If there’s anything I can support you with I’d be happy to since I now have more time again.

charles-allen:
> Thanks for taking this over @charles-allen! I had a look at your changes and they look great. I’ve updated my branch to incorporate them. I also added you as a collaborator to my repo so you should be able to push to that branch.
> 
> If there’s anything I can support you with I’d be happy to since I now have more time again.

I was planning to clean up a few loose ends with the tests today, then I think the focus should be getting everything green & merged :)

9-lives:
When can i use experimentalResolverFeatures: true since it doesn't work now?

charles-allen:
> I was planning to clean up a few loose ends with the tests today, then I think the focus should be getting everything green & merged :)

Sorry for the delay -- this has been on my mind but I've had a couple of busy weeks. I will make time early this week!

> I’ve updated my branch to incorporate them. I also added you as a collaborator to my repo so you should be able to push to that branch.

Ok, then I will only make changes here from now on!

charles-allen:
@cspotcode @geigerzaehler - Quick update on my side. I've been swamped recently. I haven't forgotten about this, but if you need to move without me, go for it! Otherwise I hope to make some time for this "soon".

cspotcode:
@charles-allen No worries, I've been pretty busy myself.  Finally finding time for some bigger ts-node features like this one.  Sadly, support for .cts/.mts/node12/nodenext has priority since I want to get that out ahead of the next stable release of TS.  So I'm hammering away on that and will get back to this path mapping stuff after.

cspotcode:
I merged the latest main branch changes into this PR.  I also added file extensions to a bunch of the ESM imports, since node's ESM stuff requires file extensions.  Still more to be done but getting there.

A great next step, for me or anyone who has a quick minute:
> import relative should not succeed using star-path

I think this means we're apply a `"*"` mapping to a relative import: `import "./should-not-resolve.js";`
But path mappings should not apply to relative import specifiers.

Probably a relatively simple change to our logic in `src/path-mapping.ts`

``````
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import at baseUrl without file extensions Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import at baseUrl without file extensions Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › ignore type definitions (5.7s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import at baseUrl Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › ignore type definitions (5.7s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import invalid path should error & list candidates Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import from js, jsx, tsx Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › fallback to node built-in (5.7s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import below baseUrl Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import node built-in (5.7s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import node_modules (5.7s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › fallback to node built-in (5.7s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import within node_modules ignores paths (5.7s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import relative Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import below baseUrl Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-no-paths.json › import at baseUrl Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import from js, jsx, tsx Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import at baseUrl without file extensions Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import node built-in (6.1s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import node_modules (6.2s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import below baseUrl Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import at baseUrl Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import node built-in (6.3s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import relative Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import invalid path should error & list candidates Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import relative Rejected promise returned by test
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › ignore type definitions (6.4s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import within node_modules ignores paths (6.4s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import within node_modules ignores paths (6.4s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › fallback to node built-in (6.4s)
  ✔️ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json › import node_modules (6.4s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import from js, jsx, tsx Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json › import invalid path should error & list candidates Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map using a prefix Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-star-path.json only › import relative should not succeed using star-path Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map to static (no wildcard) Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map to first available candidate Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map to js, jsx, tsx Rejected promise returned by test
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map using more specific candidate Rejected promise returned by test
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import at baseUrl without file extensions Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import invalid path should error & list candidates (6.9s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › fallback to node built-in (7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import at baseUrl (7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import node built-in (7s)
  ❌ path-mapping › path mapping cjs › tsconfig-baseurl-some-paths.json only › map from js, jsx, tsx Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import below baseUrl (7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › ignore type definitions (7.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › ignore type definitions (7.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import from js, jsx, tsx (7.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import within node_modules ignores paths (7.2s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import at baseUrl without file extensions Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import node_modules (7.2s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-no-paths.json › import relative Rejected promise returned by test
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import at baseUrl Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › fallback to node built-in (7.4s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import below baseUrl Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import within node_modules ignores paths (7.5s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import node_modules (7.6s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › ignore type definitions (7.6s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import at baseUrl without file extensions Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import node built-in (7.7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import below baseUrl (7.7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import node built-in (7.7s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import from js, jsx, tsx (7.7s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import relative Rejected promise returned by test
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import from js, jsx, tsx Rejected promise returned by test
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json › import invalid path should error & list candidates Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › fallback to node built-in (7.8s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import node_modules (7.8s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import at baseUrl (7.8s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import within node_modules ignores paths (7.9s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import invalid path should error & list candidates (8s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map to static (no wildcard) (8s)
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json › import relative Rejected promise returned by test
  ❌ path-mapping › path mapping esm › tsconfig-baseurl-star-path.json only › import relative should not succeed using star-path Rejected promise returned by test
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map using more specific candidate (8.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map using a prefix (8.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map to first available candidate (8.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map to js, jsx, tsx (8.1s)
  ✔️ path-mapping › path mapping esm › tsconfig-baseurl-some-paths.json only › map from js, jsx, tsx (8.2s)
  ─
```

sleep-written:
I'm making a [custom loader](https://github.com/sleep-written/bleed-believer/tree/path-alias/packages/path-alias) to aliases paths using only the tsconfig.json file. It's a new version of the obsolete [ts-path-mapping](https://github.com/sleep-written/ts-path-mapping).

An example of how to use the loader is:
```bash
node --loader ts-node/esm --loader @bleed-believer/path-alias ./src/index.ts
```

I have one problem: In my previous library, the mechanism to detect is ts-node is running, is check the symbol property `'ts-node.register.instance'`. But when i check that prroperty symbol, it doesn't exists:
```ts
export const esm = new class {
    #isTsNode: boolean;
    #aliases: TsconfigAlias;
    #symbol: symbol;


    constructor() {
        // Mark this process that this library is injected
        this.#symbol = Symbol('@bleed-believer/path-alias');
        (process as any)[this.#symbol] = true;

        // Check if we are using ts-node
        // THIS FAIL BECAUSE THE SYMBOL DOESN'T EXISTS
        this.#isTsNode = Object
            .getOwnPropertySymbols(process)
            .some(x => x.description === 'ts-node.register.instance');

        // Get alias
        const tsconfig = new Tsconfig();
        this.#aliases = tsconfig.getAliases(this.#isTsNode);
    }

    // ...
}
```

My question is: __Exist another method of get if the app is running with ts-node (in loaders)?__

sleep-written:
I made a package to handle the path alias in development (with ts-node) or directly with node in production:
[Link to @bleed-believer/path-alias](https://www.npmjs.com/package/@bleed-believer/path-alias)

Works with CJS and ESM projects, and works with [ava](https://www.npmjs.com/package/ava) (unit testing). I'll apreciate your feedbacks.

charles-allen:
> Any update on this pull?

FYI @cspotcode @geigerzaehler - I have time to work on this beginning of July

muneeb26 (commented):
Heya, any update on this PR? some estimate of when it'll be resolved?

GabrielMSharp:
Heya, any update here? Cheers

AlanMorel:
Would love to get an update on this, if at all possible. Cheers!

rhummelmose:
:(

menosprezzi:
@cspotcode How can I help here?

hpx7:
Is this work still planned?

RedStars071:
some news of this pr?

testgitdl:
Hello! Do you know when this will be available? Thanks!

kuksik:
some news about this PR?

mprinc:
Sad that the `ts-node` tool is too complex (including its dependencies) that it is not possible to handle it promptly enough but it rather takes year and half ... honestly the whole ESM model with ts-node + ts paths is broken because of that, and a lot of libs are now delivering ONLY ESM packages ... Thanks God, I am using that workaround (`--experimental-specifier-resolution=node --loader `) and more less I am ok, but it is insane that it cannot be integrated properly and that I have to have that ugly scripts: 

`node  --experimental-specifier-resolution=node --loader $COLABO/ts-esm-loader-with-tsconfig-paths.js index.ts`  rather than just `ts-node index.ts` ...

0x80:
@mprinc The workaround is not yet clear to me from skimming through this discussion. Could you share the content of your `ts-esm-loader-with-tsconfig-paths.js` please?

trim21:
suggesting [tsx](https://github.com/esbuild-kit/tsx) for this feature

Naddiseo:
@0x80 
> @mprinc The workaround is not yet clear to me from skimming through this discussion. Could you share the content of your `ts-esm-loader-with-tsconfig-paths.js` please?

This is what I'm using with node 16

```typescript
/*
loader.js

Usage (I'm also using dotenv, but you can omit the dotenv parts if needed):
DOTENV_CONFIG_PATH=.env node -r dotenv/config --loader=./loader.js /bin/path/to/cli.ts
*/
import { pathToFileURL } from "node:url";
import { getFormat, load, resolve as resolveTs, transformSource } from "ts-node/esm";
import * as tsConfigPaths from "tsconfig-paths";

export { getFormat, transformSource, load };

const { absoluteBaseUrl, paths } = tsConfigPaths.loadConfig();
const matchPath = tsConfigPaths.createMatchPath(absoluteBaseUrl, paths);

export function resolve(specifier, context, defaultResolver) {
  const mappedSpecifier = matchPath(specifier);
  if (mappedSpecifier) {
    specifier = `${mappedSpecifier}.js`;

    /*
    the resolve functionality can only work with file URLs, so we need to convert, this is especially
    the case on windows, where the path might not be a valid URL.
    */
    const url = specifier.startsWith("file:") ? specifier : pathToFileURL(specifier.toString());

    return resolveTs(url.toString(), context, defaultResolver);
  } else {
    // If we can't find a mapping, just pass it on to the default resolver
    return resolveTs(specifier, context, defaultResolver);
  }
}
```

gudh:
@trim21 thanks, `tsx` works like a charm, no configuration needed!

troywweber7:
@trim21 I might be missing something... Why `tsx` and not `mts` (similar to `mjs`)? Or maybe I'm confusing who this was directed at...

trim21:
> @trim21 I might be missing something... Why `tsx` and not `mts` (similar to `mjs`)? Or maybe I'm confusing who this was directed at...

that's a link, click it

mprinc:
**NOTE**: I am investigating [`tsx`](https://github.com/esbuild-kit/tsx) and considering already using esBuild actively in my other projects like codemod tools, which I have a great experience with both efficiency and community, I might soon switch to it. It seems pretty straightforward to do it.

loynoir:
> suggesting [tsx](https://github.com/esbuild-kit/tsx) （and also esm/cjs loader in its description） for this feature

@trim21 Coming from https://github.com/esbuild-kit/tsx/issues/113 😞

Baterka:
Any movement on this?

sthasam2:
Is this still in the works?

Dudeprogram:
no dear

billomore:
`tsx` doesn't play well with monorepos, namely [with decorators in monorepos](https://github.com/privatenumber/tsx/issues/393). `swc` [doesn't play nice with ESM either](https://github.com/swc-project/register/issues/12), not sure if planned to fix. `ts-node` was the last citadel that worked out of the box until recently, but without any change or update it just stopped... sad :(

genesiscz:
we really need this :(

cameralis:
Tried to run ts. After 4 hours of debugging I ended up here. Where even am I???

ladal1:
We're two and a half year into this PR, loader hooks for esm are on release condidate stage, can we push this the final mile?

ethyaan:
Really? 3 years in the making?? is this hard to push few lines of code and test it nowadays?

phips28:
🍿  4 years

geigerzaehler:
Hi everybody. A lot has changed since I opened this PR. I’ve since switched to [tsx](https://tsx.is/) and [subpath imports](https://nodejs.org/docs/latest/api/packages.html#subpath-imports) and I have no motivation to continue this effort, so I’m closing the PR. The branch will remain available if anyone wants to pick it up.

To those frustrated by the lack of progress: open source runs on volunteer effort. If something matters to you, contribute. Drive-by complaints don’t help — they actively discourage maintainers and contributors from engaging.

That doesn’t mean there is nothing to criticize about how this PR was handled. But if you're just pointing out that a problem still exists, without offering help or insight, you're not moving things forward.

PR #1591, opened 2022-01-07
docs: add option to display the last update time

micalevisk:
Prior this PR, there was no way to tell if we are reading a new version of the docs besides reading it entirely.

Now we'll get something like this:

image

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -124,6 +124,7 @@ module.exports = {
       {
         docs: {
           sidebarPath: require.resolve('./sidebars.js'),
+          showLastUpdateTime: true,
           editUrl:
             'https://github.com/TypeStrong/ts-node/edit/docs/website/',
         },


cspotcode:
Git history has all this information, right?  And it'll show what paragraphs have been updated so you won't have to read the entire page again.

Do you know how docusaurus gets this information?  Will it be affected by our merge workflow? https://github.com/TypeStrong/ts-node/blob/main/CONTRIBUTING.md#documentation
I'm imagining a page gets updated on the `docs` branch, which bumps the last updated time, then merged to `main` which bumps the time again even though nothing changed.

micalevisk:
> Git history has all this information, right? And it'll show what paragraphs have been updated so you won't have to read the entire page again.

yeah, I meant by reading only the docs site.

> Do you know how docusaurus gets this information? Will it be affected by our merge workflow? 

This isn't clear on the their docs https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#showLastUpdateTime That could be an issue. 

Also, I believe I would have to change from `docs` instead. My bad. Thanks.

PR #1664, opened 2022-03-01
Path mapping part 2

### LINKED ISSUE #704: Add `tsconfig-paths` to documentation
I learned about [tsconfig\-paths](https://github.com/dividab/tsconfig-paths) in https://github.com/TypeStrong/ts-node/issues/138.

Path mappings are very useful, so using `tsconfig-paths` is often a must have for me when using `ts-node`.

I think it would be helpful to list of useful modules like [tsconfig\-paths](https://github.com/dividab/tsconfig-paths) in the [README\.md](https://github.com/TypeStrong/ts-node/blob/master/README.md). (Similar to recommendations made for [Watching and Restarting](https://github.com/TypeStrong/ts-node/blob/master/README.md#watching-and-restarting).)

EDIT from @cspotcode: marked this ticket as "help wanted" with a list of tasks here: https://github.com/TypeStrong/ts-node/issues/704#issuecomment-864211793

cspotcode:
@charles-allen I'm creating this pull request so we can discuss and review the tests you are adding for path-mapping.

Closes #704 
Closes #1585
Closes #1515

This is a continuation of the work started in #1585

Below is the list of tests, copied from https://github.com/TypeStrong/ts-node/pull/1585#pullrequestreview-869432652

- test when `baseUrl` is set and `paths` is omitted
  - should still map `import "foo"` to `${baseUrl}/foo`
- test when `baseUrl` is set and `paths` maps `"*": ["lib/*"]`
  - `import "foo"` should *not* attempt `${baseUrl}/foo`; should only attempt `${baseUrl}/lib/foo`
- test when specifier matches a paths pattern but must fallback to node_modules
  - for example, `paths: {"*": ["src/*"]}`, `import "lodash"` should fallback to `node_modules/lodash`
- test when specifier matches a paths pattern but must fallback to node built-in module
  - same as above but for `import "fs"`
- When mapping candidate resolves to a `.d.ts`, resolution should be skipped; try next one
- test that external modules (`node_modules/*`) are unaffected by mappings
  - if path mappings map "lodash" to "./src/lodash" and `node_modules/external-library/index.js` does `import "lodash"`, it should *not* get `./src/lodash`
- test that mapping does not apply to relative imports
  - `import "./foo"` should not be mapped even if `"*"` mapping exists
  - `import "/absolute/path/foo"` should not be mapped even if `"*"` mapping exists


cspotcode:
@charles-allen 

I have moved the path mapping tests into their own spec file: `src/test/path-mapping.spec.ts`

I am running the tests with this command:

```
npm run -- test-local --match 'path mapping*'
```

There are 2 suites in that file: `path mapping cjs` and `path mapping esm`.  The `cjs` suite always runs, but the `esm` suite runs conditionally on node versions that support esm hooks.

If you decide to organize this file differently, that's fine with me.

I have not yet fixed the failing tests.

tests/esm-path-mapping/tsconfig-base.json
```diff
@@ -0,0 +1,12 @@
+{
+  "include": ["**/*.ts", "**/*.mts", "**/*.tsx", "**/*.js", "**/*.jsx"],
+  "compilerOptions": {
+    "module": "ESNext",
+    "allowJs": true,
+    "jsx": "react",
+    "moduleResolution": "node"
+  },
+  "ts-node": {
+    "experimentalPathMapping": "both"
```
cspotcode, line 10: Path mapping is opt-in, so I added this to your tests to enable the path-mapper.

tests/esm-path-mapping/node_modules/ambient/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
```
cspotcode, line 2: Node was complaining about ESM `export` syntax in these fake `node_modules`; seems they need their own package.json set to type: module.

tests/esm-path-mapping/node_modules/ambient/index.js
```diff
@@ -0,0 +1 @@
+export default 'ambient';
```
cspotcode, line 1: Node only accepts ESM export syntax if module is in ESM mode; see next comment.

tests/esm-path-mapping/built-in-star-path.ts
```diff
@@ -0,0 +1,6 @@
+import * as assert from 'assert';
+
+// Fallback to built-in
+import { stat } from 'fs';
```
cspotcode, line 4: Attempting to import `assert` is also testing this fallback, since `assert` is a node builtin.

tests/esm-path-mapping/path-relative.ts
```diff
@@ -0,0 +1,2 @@
+// Should fail because path mapping should be ignored
+import foo from './1-foo';
```
cspotcode, line 2: This test is failing because this import is causing a TypeScript type-checking error, so the path-mapping error never happens.  You can tackle this a few different ways:

You can avoid the typechecking error with `@ts-expect-error` or enable `transpileOnly`.

Alternatively, can we test when `./1-foo.ts` *does* exist?  I'm not sure, but that might work.  The test's goal is to prove that path mapping is not applied.

What if we have `paths: {"*": ["subdirectory/*"]}`
It'll try to map everything into that subdirectory.

If we have two files `subdirectory/1-foo.ts` and `1-foo.ts` and we do `import from './1-foo'`, this test needs to prove that we end up loading `1-foo.ts`, *not* subdirectory/1-foo.ts`

tests/esm-path-mapping/tsconfig-base.json
cspotcode: Nevermind, I was wrong.  I forgot that path mapping is enabled by default for ESM, but not for CJS.

tests/esm-path-mapping/built-in-star-path.ts
charles-allen: Several of the tests have been failing due to this! I wonder if we should add a pre-condition on the other tests so the error is clearer, like:
> failed precondition: assert couldn't be resolved

Then when the tests failed you'll see:
- n x failed precondition: `assert` couldn't be resolved
- 1 x node built-in resolution failed for `assert`

tests/esm-path-mapping/path-relative.ts
charles-allen: I think we need to test both false positives & false negatives. If `@ts-expect-error` works that keeps the fix precise/localized.

tests/esm-path-mapping/node_modules/ambient/index.js
charles-allen: Good catch!

tests/esm-path-mapping/tsconfig-base.json
charles-allen: Relevant when I create the CJS tests though, right?

charles-allen:
> I am running the tests with this command:
> 
> ```
> npm run -- test-local --match 'path mapping*'
> ```

I was thinking about adding an example like this to the collab docs, since it took me a while to find the right syntax for partial test runs

> If you decide to organize this file differently, that's fine with me.

I was more keen to rearrange the example projects in the other folder

> I have not yet fixed the failing tests.

No problem. I aim to add some more failing tests tomorrow! :)

cspotcode:
> I was thinking about adding an example like this to the collab docs, since it took me a while to find the right syntax for partial test runs

Good idea, I went ahead and added a blurb here: https://github.com/TypeStrong/ts-node/pull/1678

tests/esm-path-mapping/path-relative.ts
cspotcode: This path mapping pull request only affects node's resolution behavior; it does not affect the typechecker at all.  This means the code changes in this pull request can be fully tested with typechecking disabled: `--transpileOnly` mode.  How about we do that for simplicity?  For all of these test cases, we use `transpileOnly` mode.

charles-allen: Happy with either! :)

charles-allen:
FYI I can get back on this tomorrow!

magyargergo:
Hello guys,

I just would like to know what are we missing from here? I really look forward to this PR! :) Amazing job guys!

cspotcode:
@magyargergo to see what we're missing, read through all the comments on this pull request and on #1585.  That will get you up to speed.

tests/cjs-path-mapping/tsconfig-base.json
```diff
@@ -0,0 +1,11 @@
+{
```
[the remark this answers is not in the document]
cspotcode: Do you want to write those tests?

cspotcode:
@charles-allen do you want me to resolve the merge conflicts with our `main` branch?  I'm happy to do that since they are likely caused by my changes, so I'm probably best equipped to do that quickly.

tests/cjs-path-mapping/tsconfig-base.json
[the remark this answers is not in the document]
magyargergo: I have to get myself familiar with the `test` setup. Currently, I'm not sure I could make a test setup which `tsconfig` contains 

**tsconfig.node12.json**
```json
{
  "ts-node": {
    "esm": true <-- can this be done by the current test infra?
  }
  "compilerOptions": {
    "baseUrl": "./",
    "paths": [ ... ],
    "moduleResolution": "node12"
  }
}
```
**package.json**
```json
{
  "devDependencies": {
    "typescript" "^10.6.0"
}

**Comment by charles-allen** [comment]
> @charles-allen do you want me to resolve the merge conflicts with our `main` branch? I'm happy to do that since they are likely caused by my changes, so I'm probably best equipped to do that quickly.

Sure; especially if you fixed the `assert` import under ESM! :)

**Comment by charles-allen** [comment]
@cspotcode - Another quick update; I've been making slow progress, but progress none-the-less!

My main focus this weekend was **expanding the tests to cover cjs**. I think we want to run the same tests across all environments, so I've tried to refactor to decouple the test cases from the environments. I've gone through a few iterations trying to work out how to do this without making it less readable / less approachable (not sure how successful I've been).

In my current attempt I've reduced `path-mapping.spec.ts` to a single list of shared test cases, which I loop over setting different values for `MODULE_TYPE` (cjs vs esm) and `PROJECT_CONFIG` (which is just varying `paths`: none vs partial vs star). I suspect the `execBuilder` code especially can be simplified/improved, especially the hack I added to conditionally throw the errors.

Here's an example showing the current test-run output (this is essentially just 4 tests):
```
npm run test-local -- --match '*path mapping*' -v
```
```
  ✔ path-mapping › path mapping cjs > project: tsconfig-baseurl-only.json > fallback to node built-in
  ✔ path-mapping › path mapping cjs > project: tsconfig-baseurl-only.json > fallback to node_modules
  ✔ path-mapping › path mapping cjs > project: tsconfig-baseurl-only.json > imports within node_modules ignore paths
  ✔ path-mapping › path mapping cjs > project: tsconfig-baseurl-only.json > ignore type definitions
  ✔ path-mapping › path mapping cjs > project: tsconfig.json > fallback to node built-in
  ✔ path-mapping › path mapping cjs > project: tsconfig.json > fallback to node_modules
  ✔ path-mapping › path mapping cjs > project: tsconfig.json > imports within node_modules ignore paths
  ✔ path-mapping › path mapping cjs > project: tsconfig.json > ignore type definitions
  ✔ path-mapping › path mapping cjs > project: tsconfig-star-path.json > fallback to node built-in
  ✔ path-mapping › path mapping cjs > project: tsconfig-star-path.json > fallback to node_modules
  ✔ path-mapping › path mapping cjs > project: tsconfig-star-path.json > imports within node_modules ignore paths
  ✔ path-mapping › path mapping cjs > project: tsconfig-star-path.json > ignore type definitions
  ✔ path-mapping › path mapping esm > project: tsconfig-baseurl-only.json > fallback to node built-in
  ✔ path-mapping › path mapping esm > project: tsconfig-baseurl-only.json > fallback to node_modules
  ✔ path-mapping › path mapping esm > project: tsconfig-baseurl-only.json > imports within node_modules ignore paths 
  ✔ path-mapping › path mapping esm > project: tsconfig-baseurl-only.json > ignore type definitions
  ✖ path-mapping › path mapping esm > project: tsconfig.json > fallback to node built-in 
  ✖ path-mapping › path mapping esm > project: tsconfig.json > fallback to node_modules 
  ✖ path-mapping › path mapping esm > project: tsconfig.json > imports within node_modules ignore paths 
  ✖ path-mapping › path mapping esm > project: tsconfig.json > ignore type definitions 
  ✖ path-mapping › path mapping esm > project: tsconfig-star-path.json > fallback to node built-in 
  ✖ path-mapping › path mapping esm > project: tsconfig-star-path.json > fallback to node_modules 
  ✖ path-mapping › path mapping esm > project: tsconfig-star-path.json > imports within node_modules ignore paths 
  ✖ path-mapping › path mapping esm > project: tsconfig-star-path.json > ignore type definitions 

  ...error messages...
```

I should be able to restore the remaining test cases fairly easily; I just ran out of time.

A couple of things you might have input on / know how to fix:
- I'm having some trouble importing an ESM dep in the CJS project. This doesn't seem to be a ts-node issue; tsc is also compiling away `await import('...')` to `require('...')`, and then node complains that I can't require an ESM dep (example commented out in `tests/cjs-path-mapping/import-node-modules.ts`)
- I've made quite a few subjective decisions; it might be worth having a look
- ...actually I think that's it for now :)

**Comment by cspotcode** [comment]
Looks great, I can give this a proper review tonight, but in the meantime, here are some quick answers:

---

> I suspect the execBuilder code especially can be simplified/improved, especially the hack I added to conditionally throw the errors.

I can take a look at this.  I am also fine with some hacky test code as long as I understand it, because I can improve it in the future.  "The perfect is the enemy of the done" as they say.

---

> I'm having some trouble importing an ESM dep in the CJS project

I know some tricks for this.  The easiest is to bypass TSC's compilation with `new Function`:

```
const importHelper = new Function('specifier', 'return import(specifier)')
const mod = await importHelper('./some-esm-module');
```

This might feel like a hack, but it's perfectly safe to use it in our tests, it's short and simple, and it's easy to explain in a comment.

---

> quite a few subjective decisions

Sounds good, I'll look tonight.

**Comment by charles-allen** [comment]
> I know some tricks for this. The easiest is to bypass TSC's compilation with `new Function`:
> 
> ```
> const importHelper = new Function('specifier', 'return import(specifier)')
> const mod = await importHelper('./some-esm-module');
> ```
> 
> This might feel like a hack, but it's perfectly safe to use it in our tests, it's short and simple, and it's easy to explain in a comment.

That's not so bad; and it looks like it can be extracted as a well-named helper :)

**Comment by cspotcode** [comment]
You just have to be careful because relative imports will happen relative to the file where the helper is created, not where it's invoked.

**Comment by magyargergo** [comment]
> > I know some tricks for this. The easiest is to bypass TSC's compilation with `new Function`:
> > ```
> > const importHelper = new Function('specifier', 'return import(specifier)')
> > const mod = await importHelper('./some-esm-module');
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This might feel like a hack, but it's perfectly safe to use it in our tests, it's short and simple, and it's easy to explain in a comment.
> 
> That's not so bad; and it looks like it can be extracted as a well-named helper :)

There is this option for `ts-node` https://typestrong.org/ts-node/docs/module-type-overrides/.

File: src/test/path-mapping.spec.ts [hunk 7]
```diff
@@ -0,0 +1,132 @@
...
+const EXEC_BUILDER_PARAMS: Record<ModuleType, ExecBuilderParams> = {
+  cjs: {
+    baseDir: 'cjs-path-mapping',
+    command: CMD_TS_NODE_WITHOUT_PROJECT_FLAG,
+  },
+  esm: {
+    baseDir: 'esm-path-mapping',
+    command: CMD_ESM_LOADER_WITHOUT_PROJECT,
+  },
+};
```
cspotcode, line 46: To simplify this boilerplate, could probably do

```
EXEC_BUILDER_PARAMS = {...} as const;
```
and then later on
```
for(const [moduleType, execBuilderParams] of Object.entries(EXEC_BUILDER_PARAMS)) {
```

The type system will figure it out.

src/test/path-mapping.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+} from './helpers';
+import { test } from './testlib';
+
+test.beforeAll(installTsNode);
+
+const execBuilder = (
+  command: string,
+  moduleDir: string,
+  tsConfig: string = 'tsconfig.json'
+) => {
```
cspotcode, line 19: Prefer function declarations in general.

```suggestion
function execBuilder (
  command: string,
  moduleDir: string,
  tsConfig: string = 'tsconfig.json'
) {
```

tests/cjs-path-mapping/tsconfig-base.json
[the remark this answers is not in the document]
cspotcode: > "experimentalPathMapping": "both"

This is sufficient; we do not need to test any other variations of this option.  Before merging this feature, I will probably merge `"experimentalPathMapping"` into [`"experimentalResolverFeatures"`](https://typestrong.org/ts-node/docs/options#experimentalresolverfeatures)  This will simplify things and set us up to enable by default in the future, once our various resolver behaviors stabilize.

> which version of TypeScript ts-node uses internally?

We use Typescript as a peer dependency.  We use the version in your project, supporting as far back as 2.7.  Our test matrix on Github Actions runs multiple versions; check the Github Actions configuration for details.  You shouldn't need to care for these tests, since as I mentioned previously, we can completely disable typechecking for these tests.  (discussion: https://github.com/TypeStrong/ts-node/pull/1664#discussion_r816426376)

cspotcode (commented):
I resolved the merge conflicts with our main branch and left some comments.  Looks good so far.

You mentioned that failing assert import.  If it's still the one caused by `TSError`, that's a typescript typechecking error, and we can disable typechecking for these tests.  Either TS_TRANSPILE_ONLY or adding transpileOnly to the tsconfigs.

src/test/path-mapping.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  return (file = 'index.ts', throwErrors = true) =>
+    partialExec(`${command} ${file}`).then((res) => {
+      if (throwErrors && res.err) throw res.err;
+      return res;
+    });
+};
+
+type ModuleType = 'cjs' | 'esm';
+const MODULE_TYPES: ModuleType[] = ['cjs', 'esm'];
```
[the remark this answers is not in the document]
charles-allen: I did have `const` in there at some point. Not sure why I removed it. I'll take another look.

src/test/path-mapping.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+      // Create ts-node runner for this config
+      const exec = execBuilder(
+        execBuilderParams.command,
+        execBuilderParams.baseDir,
+        project
+      );
+
+      test.suite(`project: ${project}`, (test) => {
+        test(`fallback to node built-in`, async (t) => {
+          await t.notThrowsAsync(exec('import-node-built-in.ts'));
```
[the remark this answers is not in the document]
charles-allen: The reason I changed this is because when the test fails, it prints the error 3 times!

src/test/path-mapping.spec.ts
charles-allen: Out of interest, because we often have this discussion on our projects, is this just preference or is one better?

src/test/path-mapping.spec.ts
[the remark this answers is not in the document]
cspotcode: Oh is this because using jest `expect()` has messier output than AVA's built-in assertions?  Yeah, that's annoying.  I've tried to make that better -- I think the improvements have merged in from the `main` branch -- but if you still don't like it, could use `t.is(null)`

cspotcode:
@charles-allen how's it going?  Any blockers?

charles-allen:
> @charles-allen how's it going? Any blockers?

No blockers; just time conflicts. This is still upfront in my mind; I am planning to wrap it up (the tests) this weekend! Watch this space :)

src/test/path-mapping.spec.ts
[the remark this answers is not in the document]
charles-allen: I've cleaned this up a bit

src/test/path-mapping.spec.ts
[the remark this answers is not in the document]
charles-allen: I've reverted to the suggested style, except that I've used `toBeNull()` instead of `toBe(null)` (because the expect JS docs said it gives better error messages). Currently I have no failing tests, so I guess we'll see!

src/test/path-mapping.spec.ts
charles-allen: I think `Object.keys` and `Object.entries` have pretty poor type signatures. But as it turns out, I only needed `Object.values`, so this is now looking better :)

I've always used the prefix `<const>` instead of the suffix `as const`; I think it reads more clearly. Any preference?

src/test/path-mapping.spec.ts
charles-allen: Done

charles-allen:
@cspotcode - I think I've finished the tests 🥳 🎉 

A lot of the new tests are currently failing. I think:
- a lot are imports that don't have an explicit file extension (ex: `import foo from 'foo'`) that fail, even when `from foo.js` would succeed -- I thought that with node resolution, dropping the extension is ok?
- a few are expected errors failing because the error message format is different to expected

I'm not 100% sure what `import xyz from '/xyz'` is supposed to do -- does it mean the root directory of the entire computer? Do you want to test that somehow? Maybe using `resolve` & `__dirname`?

Do you have any more ideas for tests? Or anything else that needs cleaning up?

cspotcode:
> I thought that with node resolution, dropping the extension is ok?

Node only allows dropping some extensions when using `require()` or when using `--experimental-specifier-resolution=node`.  For example, `require()` is allowed to omit a `.js` extension but not `.cjs` or `.mjs` extensions.  Our ESM loader maps from `.js` imports to `.ts` where appropriate, but we do not have a corresponding `require()` resolver hook to do the same mapping.  This means ESM can `import from './foo.js'` to load `foo.ts`, but `require('./foo.js')` will fail to map to `./foo.ts`.  I am working on adding this mapping behavior to `require()` in another pull request.

> a few are expected errors failing because the error message format is different to expected

Does this mean we need to update the tests to understand different variants of error messages?

> I'm not 100% sure what `import xyz from '/xyz'`

Yeah, that's probably testing absolute paths, to make sure they are not accidentally treated as being relative or dependencies.  Where do you see `/xyz`?  In the code or in a comment?

cspotcode:
Also, congrats on finishing the tests!  I will review them this week.

charles-allen:
> Also, congrats on finishing the tests! I will review them this week.

I spoke too soon. I think on review, I actually haven't finished these:

> * test when specifier matches a paths pattern but must fallback to node_modules
>   * for example, `paths: {"*": ["src/*"]}`, `import "lodash"` should fallback to `node_modules/lodash`

I need to add paths for `some-cjs-dependency` & `some-esm-dependency` that miss, so that it's actually falling-back

> * test when specifier matches a paths pattern but must fallback to node built-in module
>   * same as above but for `import "fs"`

I need to add a path for `fs` that misses local files

> * test that external modules (`node_modules/*`) are unaffected by mappings
>   * if path mappings map "lodash" to "./src/lodash" and `node_modules/external-library/index.js` does `import "lodash"`, it should _not_ get `./src/lodash`

I need to add a path for `lodash` that doesn't miss; I currently have it hitting based on the baseUrl only

> * test that mapping does not apply to relative imports
>   * `import "/absolute/path/foo"` should not be mapped even if `"*"` mapping exists

Still need to write this one!

charles-allen:
> > I thought that with node resolution, dropping the extension is ok?
> 
> Node only allows dropping some extensions when using `require()` or when using `--experimental-specifier-resolution=node`. For example, `require()` is allowed to omit a `.js` extension but not `.cjs` or `.mjs` extensions. Our ESM loader maps from `.js` imports to `.ts` where appropriate, but we do not have a corresponding `require()` resolver hook to do the same mapping. This means ESM can `import from './foo.js'` to load `foo.ts`, but `require('./foo.js')` will fail to map to `./foo.ts`. I am working on adding this mapping behavior to `require()` in another pull request.

Should I be explicitly adding `.js` everywhere then?

> > a few are expected errors failing because the error message format is different to expected
> 
> Does this mean we need to update the tests to understand different variants of error messages?

I copied the error messages from the example tests. I wasn't sure if you intended to update the tests to match the implementation or the implementation to match the tests. If it's the former, I could do that!

> > I'm not 100% sure what `import xyz from '/xyz'`
> 
> Yeah, that's probably testing absolute paths, to make sure they are not accidentally treated as being relative or dependencies. Where do you see `/xyz`? In the code or in a comment?

It's the last requested test case! (I haven't done it yet)

charles-allen:
That sounds like a lot left, but I think the above are small fixes, so I think I'm pretty close -- I'll try to do another push tomorrow night to wrap it up :)

cspotcode:
> Should I be explicitly adding .js everywhere then?

Hmm, I'm not sure, I'll try to answer that when I review later this week.

> I copied the error messages from the example tests. I wasn't sure if you intended to update the tests to match the implementation or the implementation to match the tests. If it's the former, I could do that!

Also not sure about this one, I'll check when I review.

> It's the last requested test case! (I haven't done it yet)

I see.  Yeah, for that one, we need to test that path mapping does *not* apply to absolute and relative paths.

* import "./foo" should not be mapped even if "*" mapping exists.
  * For example, even if `foo` maps to `projectroot/mapped/foo.js`, `./foo` should *not* be mapped because it is a relative path
* import "/absolute/path/foo" should not be mapped even if "*" mapping exists
  * For example, even if `foo` maps to `projectroot/mapped/foo.js`, `/foo` should *not* be mapped because it is an absolute path

src/test/path-mapping.spec.ts
cspotcode: I use the postfix ` as ` syntax for all type assertions since it's compatible with TSX files.  I realize this project doesn't have TSX, but that's just the convention I use everywhere.  So we should stick to that for consistency.

cspotcode (commented):
Coming back to this:

> Should I be explicitly adding .js everywhere then?

Is it feasible to add `.js` when `import`ing, and omit it for `require()`s?  Or is that what you've already done?

It would allow us to run the tests *before* we implement `.js`->`.ts` mapping in `require()`.  ...if that makes sense.

> I copied the error messages from the example tests.

What examples are you referring to?  Did they come from me?  It's possible that I forgot providing examples somewhere.

If the error messages already being raised look appropriate, then we should update the tests to match the errors being raised.

If you think the error messages being raised are incorrect, inadequate, or should be adjusted, then let's talk about that.

charles-allen:
> What examples are you referring to? Did they come from me?

They were in the original PR before I started

> If the error messages already being raised look appropriate, then we should update the tests to match the errors being raised.
> If you think the error messages being raised are incorrect, inadequate, or should be adjusted, then let's talk about that.

Ok, I will review the current errors & get back to you :)

cspotcode:
Closing now that we can all collaborate on #1585
But should be sure that we address all tasks from this thread.

PR #1665, opened 2022-03-01
Allow json modules to be resolved in Node >=17.5.0 without flag

### LINKED ISSUE #1661: [Node 17.5.0] JSON support in ESM no longer requires --experimental-json-modules
### Search Terms

json, json modules, esm, experimental-json-modules

### Expected Behavior

To be able to import json modules without the `--experimental-json-modules` flag in `ts-node/esm`.

As of Node 17.5.0 the flag is no longer required, it just prints a usual experimental warning.

### Actual Behavior

Currently `ts-node/esm` loader only allows JSON modules if `--experimental-json-modules` is passed to the Node executable. If this option is not passed it fails with an error despite Node no longer requiring the flag.

Import a json module and use `node --loader ts-node/esm ./jsWithJsonImport.js`.

The exact line that checks the condition is [here](https://github.com/TypeStrong/ts-node/blob/main/dist-raw/node-esm-default-get-format.js#L57). Simply removing the conditional would make the behaviour be consistent with node.

### Specifications

```txt
ts-node v10.5.0
node v17.5.0
compiler v4.5.5
```

```json
{
  "include": [
    "**/*.ts",
    "**/*.json"
  ],
  "compilerOptions": {
    "target": "esnext",
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "sourceMap": true,
    "downlevelIteration": true,
    "composite": true,

    "strict": true,
    "noImplicitAny": true,
    "noImplicitThis": true,
    "alwaysStrict": true,
    "noEmit": true,

    "allowSyntheticDefaultImports": true,
    "useDefineForClassFields": true,
    "importsNotUsedAsValues": "error",
    "noUncheckedIndexedAccess": true
  },
  "ts-node": {
    "transpileOnly": true
  }
}
```

Jamesernator:
Closes #1661.

### DIFF
--- a/dist-raw/node-esm-default-get-format.js
+++ b/dist-raw/node-esm-default-get-format.js
@@ -11,7 +11,11 @@ const {
 const { extname } = require('path');
 const { getOptionValue } = require('./node-options');
 
-const experimentalJsonModules = getOptionValue('--experimental-json-modules');
+const [nodeMajor, nodeMinor] = process.versions.node.split('.').map(s => parseInt(s, 10));
+const experimentalJsonModules =
+  nodeMajor > 17
+  || (nodeMajor === 17 && nodeMinor >= 5)
+  || getOptionValue('--experimental-json-modules');
 const experimentalSpeciferResolution =
   getOptionValue('--experimental-specifier-resolution');
 const experimentalWasmModules = getOptionValue('--experimental-wasm-modules');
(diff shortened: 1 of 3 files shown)


cspotcode (approved):
Looks great, thanks!

cspotcode:
I had to bump the release of 10.6.0 to today to unblock users of TS 4.6.0, so I've scheduled this to be included in 10.7.0 which I'll publish tomorrow.

cspotcode:
10.7.0 is published, including this change.  Thanks again.

Issue #1668, opened 2022-03-02
Upgrade from diff 4 to diff 5


cspotcode:
Diff 4 was flagged in some sort of audit at work.

For starters, this audit is wrong:
the audit tool got confused and assumed that a devDependency of diff would be used by us.  Transitive devDependencies are not installed; this is not how the ecosystem works.

...but all the same, if the upgrade is painless, we might as well do it.

cspotcode:
Breaking changes in v5
https://github.com/kpdecker/jsdiff/blob/master/release-notes.md#v500

PR #1672, opened 2022-03-03
upgrade from diff 4 to 5; start bundling a smaller version with only what we need

cspotcode:
Implements #1668 

If we decide to merge this, we must get the licensing right, since we're bundling.


blakeembrey:
Why are you inlining the dependency? Seems like a headache for maintenance and actual security issues since it won't know we're relying on diff anymore.

blakeembrey:
Gotcha. We should just lazy load it as a dependency in that case. Or use a simpler diff package even. I don't remember comparing them much, just needed something simple at the time.

cspotcode:
I've generally been pretty careful about lazy-loading everything, but in this case, looks like I am not.  I'll fix that.

Another perf thing I want to investigate is avoiding `require('typescript')` entirely.  It'll be tough, I'll have to see if the complexity is worth it.  But with `@swc/core` being so fast, and with `--esm` spawning a child process, I want to see if I can make startup as snappy as possible.  Using `typescript` to read the config file is great and all but it incurs about ~100ms startup time, and that's *with* using v8-compile-cache-lib.

PR #1677, opened 2022-03-04
Fix #1667 multiline function arguments support in REPL

### LINKED ISSUE #1667: ts-node REPL: error on multiline function arguments
Functions with arguments declared in one line work in `ts-node` REPL:

```
> ts-node 
> function myFn(a: string, b: string) {
...   console.log(a, b)
...   return a
... }
undefined
> myFn('test', '!')
test !
'test'
```

Trying to paste function with multline arguments declaration into `ts-node` REPL:

```
function myFn2(
  a: string,
  b: string
) {
  console.log(a, b)
  return a
}
```

result:

```
> ts-node
> function myFn2(
<repl>.ts:9:10 - error TS2391: Function implementation is missing or not immediately following the declaration.
             
9 function myFn2(
           ~~~~~
<repl>.ts:9:10 - error TS7010: 'myFn2', which lacks return-type annotation, implicitly has an 'any' return type.

9 function myFn2(
           ~~~~~
<repl>.ts:10:1 - error TS1005: ')' expected.
```

### Specifications

```
> ts-node -vv       
ts-node v10.6.0
node v14.16.1
compiler v4.2.3
```

```
> cat /etc/lsb-release
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=19
DISTRIB_CODENAME=tara
DISTRIB_DESCRIPTION="Linux Mint 19 Tara"
> cat /etc/upstream-release/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04 LTS"
```

d9k:
Fixes #1667


cspotcode:
Are there situations where `TS7010: Function, which lacks return-type annotation, implicitly has an 'any' return type.` is a good diagnostic that you *want* to see in the REPL?  Situations where you don't want our recovery code logic to kick in, because you've entered complete code, but that error should be raised?

cspotcode:
@d9k 

I found a case where that diagnostic *should* be raised:

[Playground](https://www.typescriptlang.org/play?strictNullChecks=false#code/MYewdgzgLgBAZmGBeGAKAhgLhusBPASmQD4dkYwBXAG2oG4AoIA)
With `strictNullChecks` turned off:

```
(a: any) => a = null;
```
This is a complete line of code, so the REPL user should *not* see the `...` continuation prompt.  They should see the diagnostic.

To address this, what if we introduce the concept of dependencies to RECOVERY_CODES?  Some codes are only considered recoverable if they occur alongside another code?  In this case, `TS7010` is only recoverable if there is also a code `TS2391: Function implementation is missing or not immediately following the declaration.` or `TS1005: "')' expected.", "'}' expected."`

cspotcode:
@d9k I've implemented the dependency mechanism.  Can you test and ensure this still covers your use-case?

cspotcode:
@d9k Friendly ping to please test this code change and verify that it still solves your use-case.

cspotcode:
@d9k have you had a chance to look at this?  If not, will you be able to look at it at any point in the future?

Issue #1705, opened 2022-04-01
Add error locations to TSError


paulbrimicombe:
### Desired Behavior

When a `TSError` is thrown it would be very helpful to be able to programatically extract the location of the compilation error(s) in the source code.

### Is this request related to a problem?

* I maintain a library that compiles TypeScript snippets in markdown files: https://github.com/bbc/typescript-docs-verifier
* I use `ts-node` to do the compilation because it provides a really simple and easy-to-use API (thanks! 🎉)
* If the compilation of a snippet fails, the library displays the code snippet with the lines with compilation errors highlighted.
* Currently I have to parse the `diagnosticText` to do this which is problematic:
  * The message format keeps changing which breaks the parser
  * I have to handle both coloured and non-coloured outputs which are formatted slightly differently

### Alternatives you've considered

* N/A

### Additional context

* N/A

PR #1706, opened 2022-04-01
Add full diagnostics to tserror

paulbrimicombe:
- This PR addresses https://github.com/TypeStrong/ts-node/issues/1705
- I've been through a few iterations and I _think_ the nicest API is to have a mapping of the error locations for each diagnostic code.
  - This allows clients to look up the location of each error code and also iterate through the object values depending on their requirements.
- I've added tests for the errors that are thrown when using `ts-node` as a library.
  - I opted to keep the tests separate for each of the assertions (error message, diagnostic codes, etc.) to make each different feature clear. I'm happy to change this.


cspotcode:
Thanks for this, always happy to support consumers of our API doing cool things.

I'm inclined to expose the diagnostics directly, in exactly the form that typescript provides.  That way, consumers will have everything they need no matter what they're trying to accomplish.  Is that compatible with your use-case?

src/index.ts
```diff
@@ -470,14 +470,24 @@ export const DEFAULTS: RegisterOptions = {
 export class TSError extends BaseError {
   name = 'TSError';
   diagnosticText!: string;
+  diagnostics!: ReadonlyArray<_ts.Diagnostic>;
 
-  constructor(diagnosticText: string, public diagnosticCodes: number[]) {
+  constructor(
+    diagnosticText: string,
+    public diagnosticCodes: number[],
+    diagnostics: ReadonlyArray<_ts.Diagnostic> = []
```
cspotcode, line 478: Is this argument optional to avoid a breaking change to our API surface, since `TSError` is part of our API?

src/test/index.spec.ts
```diff
@@ -974,6 +975,96 @@ test.suite('ts-node', (test) => {
...
+      expect(diagnostics).toHaveLength(1);
+      expect(diagnostics[0]).toMatchObject({
+        code: 2345,
+        start: 10,
+        length: 3,
+        messageText:
+          "Argument of type '123' " +
+          "is not assignable to parameter of type 'string | undefined'.",
+      });
+    });
```
cspotcode, line 1066: Note to self: consider moving these tests into a new spec file, perhaps `tserror.spec.ts` or `throws-diagnostics.spec.ts`  Consider also combining it with pre-existing tests for diagnostics.

cspotcode (approved):
Looks good thanks again. I have one question and one note for myself.

I can do that work when I merge; I am trying to re-organize the tests slowly over time.  Years ago, they were all in index.spec.ts.  I'm trying to break that file apart a little bit at a time.

src/index.ts
paulbrimicombe: > Is this argument optional to avoid a breaking change to our API surface, since `TSError` is part of our API?

Yes, that's exactly why it's optional — I didn't want to make this a breaking change.

paulbrimicombe:
Thanks very much for merging @cspotcode (sorry for the slow response — I've been on vacation).

PR #1729, opened 2022-04-21
Switch to new @cspotcode/source-map-support, using trace-mapping

cspotcode:
Getting test failures because the mapped stack filenames look like this in some tests: (note the `/file:/` mid-way through the path)

```
file:///d/ts-node/tests/esm/file:/d/ts-node/tests/esm/throw error.ts
```

Full example:
```
Command failed: node  --loader ts-node/esm \\"throw error.ts\\"
    (node:2859) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time␊
    Error: this is a demo
        at Foo.bar (file:///d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/file:/d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/throw error.ts:100:17)
        at new Foo (file:///d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/file:/d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/throw error.ts:98:10)
        at file:///d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/file:/d/Personal-dev/@TypeStrong/ts-node/trace-mapping/tests/esm/throw error.ts:102:1
        at ModuleJob.run (node:internal/modules/esm/module_job:197:25)
        at async Promise.all (index 0)
        at async ESMLoader.import (node:internal/modules/esm/loader:337:24)
        at async loadESM (node:internal/process/esm_loader:88:5)
        at async handleMainPromise (node:internal/modules/run_main:61:12)
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,11 +13,11 @@
       ],
     },
     {
-      "name": "Debug 2",
+      "name": "Debug Example: running a test fixture against local ts-node/esm loader",
       "type": "pwa-node",
       "request": "launch",
       "cwd": "${workspaceFolder}/tests/esm",
-      "runtimeArgs": ["--loader", "ts-node/esm"],
+      "runtimeArgs": ["--loader", "../../ts-node/esm"],
       "program": "throw error.ts",
       "outputCapture": "std",
       "skipFiles": [
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,6 +11,18 @@
       "skipFiles": [
         "<node_internals>/**/*.js"
       ],
+    },
+    {
+      "name": "Debug 2",
+      "type": "pwa-node",
+      "request": "launch",
+      "cwd": "${workspaceFolder}/tests/esm",
+      "runtimeArgs": ["--loader", "ts-node/esm"],
+      "program": "throw error.ts",
+      "outputCapture": "std",
+      "skipFiles": [
+        "<node_internals>/**/*.js"
+      ],
     }
   ],
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -317,7 +317,7 @@
       "integrity": "sha512-41qniHzTU8yAGbCp04ohlmSrZf8bkf/iJsl3V0dRGsQN/5GFfx+LbCSsCpp2gqrqjTVg/K6O8ycoV35JIwAzAg=="
     },
     "@cspotcode/source-map-support": {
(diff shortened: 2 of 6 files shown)


jridgewell:
Hi @cspotcode, I just published https://github.com/jridgewell/resolve-uri/releases/tag/v3.0.6 which should resolve your `file:` issues.

cspotcode:
Interesting, my testing tracked the issue down to some faulty logic in source-map-support, so I'm curious what was changed on your end.  Did you make it detect whether the input paths are posix paths, not file URI, and if so, also return posix paths instead of file URI?

jridgewell:
I'm assuming you have a have two `file:/d/` URLs in a sourcemap (either a `sourceRoot` and a mapUrl when constructing the `TraceMap`, or an absolute `sources` array and a `sourceRoot`, etc). `TraceMaps` `resolvedSources` uses `resolve-uri` to join those together, and I didn't properly handle semi-valid `file:` URLs (they're supposed to be `file:///foo/bar/…`, not `file:/foo/bar/…`) when joining.

cspotcode:
For posterity, here is the fix:
https://github.com/cspotcode/node-source-map-support/pull/38/files#diff-6c09ea90fab3390be21a951054452be687b7de7b791f14e10874f785dc095864R189-R207

In this particular case, the fixes in trace-mapping weren't causing this problem.  The repaired logic linked above needed to accept a file URI as the `url` argument.  It was, funny enough, expecting a path instead of a url.  When running native ESM modules, node's stack traces use file: URLs instead of absolute paths.

---

Some additional info, for curious readers:

At the time of writing, ts-node provides sourcemaps with `file` and `sources` that are absolute paths but _not_ file URIs.  (e.g. `"file": "/root/project/generated.js"`)  The sourcemaps do *not* have a `sourceRoot`.  I don't remember the reasons it was implemented this way.

https://github.com/TypeStrong/ts-node/blob/9e255575aa44a8af3cb01f4d810f8480541fe943/src/index.ts#L1546-L1555

jridgewell:
> not entirely sure if this logic is still required, if trace-mapping guarantees that all `sources` are coming to us as absolute file URI

`trace-mapping` should handle absolute file paths without issue, just make sure the POSIX and not Windows format.

Might I recommend you switch to using `@jridgewell/resolve-uri` to perform relative URL joining? It behaves similar to `new URL(url, baseUrl)`, but supports absolute URLs, protocol-relative URLs, absolute POSIX paths, and relative POSIX paths.

Issue #1505, opened 2021-10-11
[META] v11 breaking changes


cspotcode:
Breaking changes we should include in the next major, v11

## Implemented in #1760

Minimum node version: v14
Minimum TS version: 4.0 (from https://github.com/DefinitelyTyped/DefinitelyTyped#support-window)

* [x] Remove `ts.createIncrementalProgram` fallback to `ts.createEmitAndSemanticDiagnosticsBuilderProgram`
  * https://github.com/TypeStrong/ts-node/blob/1cc91d9e2cdc9ad6676296f6d3e2d190e592652b/src/index.ts#L988-L989
  * https://github.com/TypeStrong/ts-node/blob/1cc91d9e2cdc9ad6676296f6d3e2d190e592652b/src/index.ts#L1023-L1025
  * TS 3.7 already has both of these
* [x] stop installing source-map-support in `create()`; only in `register()`
* [x] Re-align TS version support with DefinitelyTyped
  * Find [DT's minimum supported version](https://github.com/DefinitelyTyped/DefinitelyTyped#support-window)
  * Find all semver checks for versions older than that; remove them
* [x] Make `enableExperimentalEsmLoaderInterop()` on by default
  * and remove the conditional code, remove that enabler method
* [x] Re-align node version support w/EOLs
  * Did we officially announce the dropping of node 10?  Or do we merely support non-EOLed versions as a general policy
* [x] remove `ts-script` entrypoint; replaced by `ts-node-script`
* [x] Make `ts-node/dist/bin` `main()` export `@internal`
* [x] Modify `nodeEval` function signature to accept `context: Context` (See TODO in the code)
* [x] change signature of REPL `nodeEval(code: string, context: any` to `nodeEval(code: string, context: Context`
* [x] remove @tsconfig/bases node10 and node12
* [x] add @tsconfig/bases node18
* [x] switch all @tsconfig/bases deps to wildcard (like is typically done for @types) so they will respect whatever version you want to install

## Still to do

* [ ] Mention minimum supported node and TS versions in docs: node 14, TS 4.0
* [ ] Clean up `createEvalAwarePartialHost` API surface
  * Augment `Service` API to allow setting overridden file contents
  * Use this new override API surface in the REPL
  * Eliminate the need for `evalAwarePartialHost` but leave a pass-through shim for backwards compatibility
  * Update code samples to use simpler API
[...]

PR #1731, opened 2022-04-26
fix(docs): Clarify `--project` shorthand option description

lobsterkatie:
To hopefully save anyone else the 45 minutes I wasted because I am not a good noticer.

Because of `tsc`'s `-p` shorthand for the `--project` option, when I saw that `ts-node` had a similar option, I didn't read closely enough to see that in `ts-node`'s case, it's a capital P.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,7 +309,7 @@ All command-line flags support both `--camelCase` and `--hyphen-case`.
 
 ## TSConfig
 
-*   `-P, --project [path]`   Path to TypeScript JSON project file <br/>*Environment:* `TS_NODE_PROJECT`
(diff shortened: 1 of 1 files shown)


cspotcode:
Thanks, the README is actually generated from the website sources.  There's a blurb about it here: https://github.com/TypeStrong/ts-node/blob/main/CONTRIBUTING.md#documentation

The "edit this page" buttons on the website will point to the correct source file, and will make the edits against the docs branch, which can update the website immediately when merged, instead of waiting for a new npm release.
https://typestrong.org/ts-node/docs/options#project

cspotcode:
I've also added a note to the list of breaking changes for v11: when `-p` is specified without `-e`, consider it an error, and exit with a helpful error suggesting that you may have wanted `-P` instead

https://github.com/TypeStrong/ts-node/issues/1505

cspotcode:
I've made this change to the docs in #1734.  I had to do it as a separate PR to follow our workflow for docs changes, but I copy-pasted the prose from here.

PR #1770, opened 2022-05-22
Use basename in source map sources field

### LINKED ISSUE #1769: Malformed source URLs in source map cache on Windows
### Search Terms

v8 coverage, coverage, source map windows

### Expected Behavior

The `sources` field of the generated [source map cache](https://nodejs.org/api/cli.html#source-map-cache) consists of file URLs (e.g., `file:///D:/Example/index.ts`).

### Actual Behavior

Some generated `sources` field contain malformed file paths (e.g., `d:/Example/index.ts`).
I think it "malformed" because the disk letter, along with `:`, is treated like a protocol. At least, it should be in upper case. It will be better if it follows the `file:///` prefix like the others.

1. Enable Node.js coverage output by setting [`NODE_V8_COVERAGE`](https://nodejs.org/api/cli.html#node_v8_coveragedir)
2. Run any `.ts` file with `ts-node`
3. Compare the `sources` fields in each produced coverage output file with each other.

### Minimal reproduction

https://github.com/PaperStrike/ts-node-bug-repro
The workflow does basically the steps above, and prints the malformed URLs.

### Specifications

* ts-node version: 10.8.0
* node version: 16.15.0
* TypeScript version: 4.6.4
* Operating system and version: Windows 11

PaperStrike:
Resolves #1769 by using the base name of the file instead of the full path in the source map `sources` field.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1668,7 +1668,7 @@ function updateOutput(
 function updateSourceMap(sourceMapText: string, fileName: string) {
   const sourceMap = JSON.parse(sourceMapText);
   sourceMap.file = fileName;
-  sourceMap.sources = [fileName];
+  sourceMap.sources = [basename(fileName)];
   delete sourceMap.sourceRoot;
   return JSON.stringify(sourceMap);
 }


cspotcode:
Closing in favor of #1771

PR #1792, opened 2022-06-09
Allow JSON imports in NodeJS 16.15

queengooborg:
This PR is a quick fix that enables JSON module support by default in NodeJS v16.15 < v17 as well as v17.5 and up.  The version number was selected based upon [prior manual testing](https://github.com/mdn/browser-compat-data/pull/16413) to determine when JSON module support was enabled by default.

This change was tested by attempting to run a script that uses a JSON import and performing the same tweak directly in `node_modules`.

### DIFF
--- a/dist-raw/node-internal-modules-esm-get_format.js
+++ b/dist-raw/node-internal-modules-esm-get_format.js
@@ -12,6 +12,7 @@ const [nodeMajor, nodeMinor] = process.versions.node.split('.').map(s => parseIn
 const experimentalJsonModules =
   nodeMajor > 17
   || (nodeMajor === 17 && nodeMinor >= 5)
+  || (nodeMajor === 16 && nodeMinor >= 15)
   || getOptionValue('--experimental-json-modules');
 const experimentalWasmModules = getOptionValue('--experimental-wasm-modules');
 const { URL, fileURLToPath } = require('url');
--- a/src/test/esm-loader.spec.ts
+++ b/src/test/esm-loader.spec.ts
@@ -15,7 +15,7 @@ import {
   EXPERIMENTAL_MODULES_FLAG,
   nodeSupportsEsmHooks,
   nodeSupportsImportAssertions,
-  nodeSupportsImportAssertionsTypeJson,
+  nodeSupportsUnflaggedJsonImports,
   nodeSupportsSpawningChildProcess,
   nodeUsesNewHooksApi,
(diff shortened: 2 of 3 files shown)


cspotcode:
Thanks, seems simple enough.  Was this mentioned in the nodejs 16.15 changelog?  Would be nice to keep a link to that changelog somewhere, in case questions arise in the future.

queengooborg:
Looks like it is, yes! https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#other-notable-changes

cspotcode:
Awesome, thanks!

We run our tests against a bunch of node versions to test stuff like this, and we conditionally check the node version in the tests to figure out what we should test.

Here is the test for JSON imports, where we check the node version.  Depending on the check, we either pass `--experimental-json-modules` or we don't.
https://github.com/TypeStrong/ts-node/blob/17aad13f845a794e6e4a559dd31897d47ab6abec/src/test/esm-loader.spec.ts#L270-L306

Can you also update that version check in the tests?  Most of the version checks are declared as boolean exports from helpers.ts, like this:
https://github.com/TypeStrong/ts-node/blob/17aad13f845a794e6e4a559dd31897d47ab6abec/src/test/helpers.ts#L60-L78

Can we move this version check into that file?  Can be named something like `export const nodeSupportUnflaggedJsonImports =`

cspotcode:
Looks like the `nodeSupportsImportAssertions` boolean should also be updated to include node >= 16.14.0

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V16.md#2022-02-08-version-16140-gallium-lts-danielleadams

https://github.com/TypeStrong/ts-node/blob/17aad13f845a794e6e4a559dd31897d47ab6abec/src/test/helpers.ts#L67-L70

cspotcode:
Tests on node nightly are failing due to a node bug; you can safely ignore those test failures.

queengooborg:
Test updates completed -- I opted for the variable name `nodeSupportsImportAssertionsTypeJson` to match the syntax (`assert {type: 'json'}`) and open and preemptively opening the door for any other import assertion properties and values!

queengooborg:
Hey there!  Just checking in to see if this needs anything more from my end, or if we're just waiting on review?

cspotcode:
@queengooborg Merged and will go out in the next release.  Thanks!

PR #1793, opened 2022-06-09
swc: Support paths from tsconfig

spmiller:
The swc transpiler can support paths, but they have to be converted to absolute paths using the baseUrl first.

### DIFF
--- a/src/test/transpilers.spec.ts
+++ b/src/test/transpilers.spec.ts
@@ -3,13 +3,23 @@
 // Should consolidate them here.
 
 import { context } from './testlib';
-import { ctxTsNode, testsDirRequire } from './helpers';
+import {
+  CMD_TS_NODE_WITHOUT_PROJECT_FLAG,
+  ctxTsNode,
+  TEST_DIR,
+  testsDirRequire,
+} from './helpers';
 import * as expect from 'expect';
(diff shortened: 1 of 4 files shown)


spmiller:
Hi maintainers, some more context in case it's not clear: `tsconfig.json` supports [paths](https://www.typescriptlang.org/tsconfig#paths), but this wasn't wired into the ts-node swc transpiler. I wired this in using absolute paths (as recommended [here](https://github.com/swc-project/swc/issues/2725#issuecomment-1051655984))

spmiller:
Hi @cspotcode, I see you're also working on swc support in #1802. Assuming you see the value in this PR, how do think it should progress? I'm happy to help out if you would like it to go in as part of #1802.

erfanium:
@spmiller 
I felt into this today. but here's a thing. we should not add a feature that is only available on swc. 
I think this feature is not supported by TypeScript official transpiler.

erfanium:
@spmiller 
I have made a tiny repo for you:
https://github.com/erfanium/esm-ts-paths

in tsconfig file, either using `swc: true` to `swc: false` doesn't make it work. that means this feature is not supported by tsc even.

or maybe i am wrong

PR #1814, opened 2022-06-24
Fix ESM node processes being unable to fork into other scripts

### LINKED ISSUE #1812: ESM flag breaks NodeJS child process `fork`
### Search Terms

child process, fork, esm, loader, brotli, config, exec argv, exec path

### Expected Behavior

Similar to when running with `node --loader ts-node/esm` or plain `node`, the main process should be able to fork and execute other NodeJS scripts (not necessarily another TS file).

When vanilla Node executes, the `execArgv` never contains the script entry-point, so that the `fork` can set the new script that is intended to be run.

### Actual Behavior

Using the builtin NodeJS `fork` method does not work when running with `ts-node --esm` because the `process.execPath` and `process.execArgv` contain the brotli-base64 encoded config and the custom loader script. This is resulting in the same script being executed all the time, going into an infinite loop based on the script.

1. Create a TS ESM file e.g. `.mts`
2. Use NodeJS `fork` and try to load another file, e.g. a plain `worker.mjs` file
3. Observe that the worker script is never executing, but instead the orignal `.mts` file is executed again, and so on.

### Minimal reproduction

1. Clone https://github.com/devversion/ts-node-esm-fork/tree/main
2. Run `yarn test`

### Specifications

```
ts-node v10.8.1
node v16.14.2
compiler v4.7.4
Done in 0.35s.
```

* Linux 64bit

devversion:
Currently, Node processes instantiated through the `--esm` flag result
in a child process being created so that the ESM loader can be
registered. This works fine and is reasonable.

The child process approach to register ESM hooks currently prevents
the NodeJS `fork` method from being used because the `execArgv`
propagated into forked processes causes `ts-node` (which is also
propagated as child exec script -- this is good because it allows nested
type resolution to work) to always execute the original entry-point,
causing potential infinite loops because the designated fork module
script is not executed as expected.

This commit fixes this by not encoding the entry-point information into
the state that is captured as part of the `execArgv`. Instead the
entry-point information is always retrieved from the parsed rest command
line arguments in the final stage (`phase4`).

Fixes #1812.


devversion:
Okay, I pushed again. Should fix the previous CI failures due the use of `import.meta.url` in TS versions which don't know about it. Other than that, I think the tests should be passing now.

https://github.com/TypeStrong/ts-node/runs/7043357305?check_suite_focus=true#step:15:609 is a little confusing to me, but that might have been just flakiness. I tried with the same Node version and couldn't reproduce.

devversion:
Hey, just FYI that this PR is ready from my side now. No rush, please let me know if I can do anything. Thanks!

src/bin.ts
```diff
@@ -48,7 +48,7 @@ export function main(
   const state: BootstrapState = {
     shouldUseChildProcess: false,
     isInChildProcess: false,
-    entrypoint: __filename,
+    tsNodeScript: __filename,
```
cspotcode, line 51: Was this renamed to indicate that this is the path to ts-node's bootstrapper, but *not* the user's desired entrypoint?

src/bin.ts
```diff
@@ -403,23 +407,76 @@ function phase3(payload: BootstrapState) {
...
+ *   2. In phase 4 to determine the actual entry-point script.
+ *
+ * Note that we need to explicitly re-resolve the entry-point information in the final
+ * stage because the previous stage information could be modified when the bootstrap
+ * invocation transitioned into a child process for ESM.
+ *
+ * Stages before (phase 4) can and will be cached by the child process through the Brotli
+ * configuration and entry-point information is only reliable in the final phase. More
+ * details can be found in here: https://github.com/TypeStrong/ts-node/issues/1812.
+ */
```
cspotcode, line 433: Thanks for adding all these comments.

It makes sense to me that we must repeat resolution in phase 4.  Are there any situations where we know, without any risk, that we can skip this duplicate resolution?  It would be a performance optimization.

If we cannot safely do this, then we shouldn't.  But if there's a safe way to do this, then I might try my hand at adding it.

src/bin.ts
```diff
@@ -566,12 +623,13 @@ function phase4(payload: BootstrapState) {
...
-    entrypoint,
+    tsNodeScript,
     ...argv.slice(2, argv.length - restArgs.length)
   );
+
   // TODO this comes from BoostrapState
   process.argv = [process.argv[1]]
-    .concat(executeEntrypoint ? ([scriptPath] as string[]) : [])
+    .concat(executeEntrypoint ? ([entryPointPath] as string[]) : [])
     .concat(restArgs.slice(executeEntrypoint ? 1 : 0));
```
cspotcode, line 633: `executeEntrypoint` is computed from our `parseArgvResult` and `phase2Result`.  Is it possible that `executeEntrypoint` might be false in a child process, when actually, it should be true?  Because the host was a REPL, but the child should not be?

I'm not sure, but I'm wondering if there is a situation where someone uses our REPL to `require('./this-script-will-fork-a-child-process')`

In that case, the host process has `executeEntrypoint === false` because it's a REPL.  But the forked child should have it set to true.

I see in `getEntryPointInfo`, we are using `argvResult.code`, `.interactive`, and `.restArgs`  In a forked process, should we avoid doing that?

cspotcode: Nevermind, I see now that `restArgs` is overwritten in `child-entrypoint.ts`.

cspotcode: Nevermind, I see now that `restArgs` is overwritten in `child-entrypoint.ts`.

cspotcode:
Thanks @devversion, I have a couple questions but otherwise this is excellent work.

src/bin.ts
```diff
@@ -403,23 +407,76 @@ function phase3(payload: BootstrapState) {
...
+  const executeEval = code != null && !(interactive && restArgs.length);
+  const executeEntrypoint = !executeEval && restArgs.length > 0;
+  const executeRepl =
+    !executeEntrypoint &&
+    (interactive || (process.stdin.isTTY && !executeEval));
+  const executeStdin = !executeEval && !executeRepl && !executeEntrypoint;
+
+  /** Unresolved.  May point to a symlink, not realpath.  May be missing file extension */
+  const entryPointPath = executeEntrypoint
+    ? resolve(cwd, restArgs[0])
```
cspotcode, line 453: cwd might be the wrong cwd, if the child process was forked with a different one. The cwd option has always been confusing, let me see if I can figure out why we're using it in the first place.

src/bin.ts
devversion: Yes, I renamed that mostly to avoid the ambiguity here (at least ambiguity to me). Happy to revert if you would want it the old way.

src/bin.ts
devversion: I think there is an issue with the `cwd` option. It's quite ambiguous: It doesn't actually control the working directory of the invoked process (kind of surprising). It's mainly used for the script to be resolved and e.g. the tsconfig

src/bin.ts
devversion: I think we could re-use this when the `argvResult` has not changed. We only need the rerun this function due to the patches in the child loader entry-point. We could indicate this change through a flag and request re-computation.

This feels like a micro optimization through, not worth it personally as this is just some simple logical expressions.

src/bin.ts
devversion: @cspotcode I added a new commit (which might be considered a breaking change? to me it sounds like a fix):

* The `--cwd` option now also sets the working directory properly
* The working directory is properly preserved when we transition into the child process for ESM.
* Added test scenarios for non-ESM `--cwd` and also for the scenario you outlined in this comment (fork cwd)

src/bin.ts
```diff
@@ -73,13 +73,17 @@ export function bootstrap(state: BootstrapState) {
   if (!state.phase2Result) {
     state.phase2Result = phase2(state);
     if (state.shouldUseChildProcess && !state.isInChildProcess) {
-      return callInChild(state);
+      // Note: When transitioning into the child-process after `phase2`,
+      // the updated working directory needs to be preserved.
+      return callInChild(state, state.phase2Result.targetCwd);
```
devversion, line 78: This explicit cwd is strictly not needed since spawn would just use the working directory we switched to, but I think being explicit is helping with readability and also makes this safer

src/bin.ts
cspotcode: Renaming is good, makes sense to me.

src/bin.ts
cspotcode: I'll have to double-check.  The cwd option has never been great, IMO, and as far as I remember, is only there for backwards compatibility to support some legacy use-cases.

It makes ts-node's logic pretend to be in that cwd but without actually changing the process's cwd.  So it affects how ts-node's `--cwdMode` locates the tsconfig, and affects paths in typechecking errors.  `--cwdMode` isn't even the default anymore because it was confusing and the newer default is better.  Just goes to show, I don't love the cwd option.

As confusing as it may be, pretty sure we should not be changing it to affect `process.cwd()`.  But I'll have to double-check the old PRs where I cleaned up this logic; it was a couple years ago.  I'm not at my computer; will check later today.

cspotcode: It is possible for projects to intentionally launch child processes in a different cwd, right?  Either by passing the cwd directly to `fork` options, or by changing the parent's working directory before forking.  We'll need to support that.

devversion: Correct. Forking into different working directories should still work. I added a test for this in ESM, but should also make sure with CJS.

 We are using the cwd option in a couple of places, and naturally I also expected it to change the initial working directory, similar to how e.g. `yarn --cwd` works. This allows me to directly call TS node scripts e.g. in the `package.json` without needing more code or an extra script.

devversion: I fixed this, and added a test. Regardless of what you decide, the additional tests can be helpful. Happy to revert the cwd changes. I believe the semantics should be that way, or the option should be removed altogether (even though it would actually be useful that way -- like `yarn --cwd`). Proposed semantics:

* `--cwd` changes the working directory before the user entry-point is invoked
* The working directory flag is omitted when forking into child processes. It's up to the user then.

cspotcode: Unfortunately, I think removal makes most sense to reduce complexity.  The option predates my involvement with the project, it's confusing, and `node` nor `tsc` has an equivalent `--cwd` option.  There's also programmatic registration scenarios, where ts-node is being registered into an existing process and probably shouldn't be changing the cwd.  Could make the cwd option a CLI-only thing, but it's really just messy feature creep.

I make frequent use of `git -C` so I understand the appeal of `--cwd` in general.  But in this case, it feels like a maintenance liability to keep it around.

Can't remove till the next major, though, cuz it'd be a breaking change.

devversion: Yes, it definitely makes some things more difficult. Not too bad IMO though.

Anyway though, I agree that the removal is a good consideration. What are the next steps for this PR?

cspotcode: I think the only thing left is: entrypoint resolution in the first child uses `--cwd`, vs subsequent forked children use `process.cwd()`.

The first child process spawned by our `--esm` flag needs to obey the CLI's `--cwd` option for backwards compatibility.  (ugh annoying)  So if you are in `/project` and you run `ts-node --cwd foo bar.ts` then the entrypoint needs to be resolved to `/project/foo/bar.ts`

Subsequently forked children should behave like vanilla node and resolve their entrypoint relative to `process.cwd()` 
 If you `process.chdir('/home/user')` and do `fork('script.ts')`, the child's entrypoint needs to resolve to `/home/user/script.ts`.

---

I think this means we need to pass a boolean to child processes telling them if they are the first child spawned by `--esm`, or they are a subsequent child spawned by `fork()`.  `child-entrypoint.ts` will need to modify the `execArgv` array to pass this flag to subsequent children.  One method is a new `--is-first-child=true|false` flag, another method is to modify and re-compress the brotli payload.

---

Maybe there are other benefits to re-compressing the brotli payload: #1831 
I'm not suggesting we implement #1831 in this PR, more thinking that re-compressing the brotli payload in this PR will have benefits for future work.

devversion:
@cspotcode Ah thanks for helping out here. I was actually working quite some time today and yesterday on always using the brotli payload (even without esm). I have something for this pretty soon, but I'm not sure if we want to land it as part of this PR.

It's a breaking change as you realized as well because it will always use the same preloaded project. I think that is worth doing though.

devversion:
do you want me to push that here or how should we proceed? The cwd option will behave like before (like you suggested)

cspotcode:
Oops, I forgot to update the tests since they use `--cwd`, I assume that's why CI failed.  I also haven't code-reviewed them yet, I might have time tonight for that.  I'll post here before I attempt further code changes to avoid conflicting with your work.

src/test/esm-loader.spec.ts
```diff
@@ -358,6 +359,88 @@ test.suite('esm', (test) => {
...
+            `${BIN_PATH} --esm --cwd ./working-dir/esm/ index.ts`
+          );
+
+          expect(err).toBe(null);
+          expect(stdout.trim()).toBe('Passing');
+          expect(stderr).toBe('');
+        });
+
+        test.suite(
+          'with NodeNext TypeScript resolution and `.mts` extension',
```
cspotcode, line 374: Is this `.mts` variant still required?  I see that the previous test case used regexp matching to avoid `import.meta.url`.  But if we can avoid that issue via `transpileOnly`, then I think this test case is identical to the previous one?

src/test/esm-loader.spec.ts
```diff
@@ -358,6 +359,88 @@ test.suite('esm', (test) => {
...
+            );
+
+            expect(err).toBe(null);
+            expect(stdout.trim()).toBe('Passing: from main');
+            expect(stderr).toBe('');
+          }
+        );
+
+        test.suite(
+          'with NodeNext TypeScript resolution and `.mts` extension',
```
cspotcode, line 427: Same question for this one: is the `NodeNext` and mts variant necessary?

tests/esm-child-process/process-forking-nested-esm-node-next/index.mts
```diff
@@ -0,0 +1,23 @@
+import { fork } from 'child_process';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
+// Initially set the exit code to non-zero. We only set it to `0` when the
+// worker process finishes properly with the expected stdout message.
+process.exitCode = 1;
+
+const projectDir = dirname(fileURLToPath(import.meta.url));
+const workerProcess = fork(join(projectDir, 'worker.mts'), [], {
```
cspotcode, line 10: Why does this test case use an absolute path for the worker's entrypoint?  Is it testing something that hasn't occurred to me?  I figure it should be same as the other tests.

cspotcode:
@devversion I think this is ready to merge and publish.  I removed the `.mts` tests because, as far as I can tell, it is sufficient to test `.ts` ESM.  I'd appreciate your review on that change, but otherwise I think we're ready to go.

As soon as this is merged, I can publish 10.9.0 (https://github.com/TypeStrong/ts-node/milestone/16), and then I think I can start merging breaking changes for 11.0.0 (will combine the https://github.com/TypeStrong/ts-node/milestone/13 and https://github.com/TypeStrong/ts-node/milestone/17 milestones)

src/test/esm-loader.spec.ts
devversion: Likely yes. Although these tests are still a little different:

* One loads the worker through an absolute path
* One loads the worker relatively

tests/esm-child-process/process-forking-nested-esm-node-next/index.mts
devversion: See response above. One test, with `.mts` I used `import.meta.url` to construct an absolute path to the worker. The other tests were not using `import.meta` (and couldn't --without transpile-only) and use a relative path.

cspotcode: Ok cool, sounds like all the things we need to test are:

1. entrypoint script runs with correct `process.cwd()`
2. `fork()` entrypoint resolves with absolute path
3. `fork()` entrypoint resolves with relative path relative to `process.cwd()`, not parent's `ts-node --cwd` flag

...for both CJS and ESM, with 3. disabled in the non-ESM case since that bug will be fixed by other tickets.

I'll update the PR now.

devversion:
@cspotcode Thanks for fixing up the rest. I was working on it already but you came first. Much appreciated.

PR #1838, opened 2022-07-14
Attempt workaround for node regression

cspotcode:


faradaytrs:
There is problem with --loader ts-node/esm as well

cspotcode:
@faradaytrs please share details on #1839, we need enough information to reproduce the issue.

PR #1849, opened 2022-07-26
when calling default* functions, you should not recursively pass default* function

giltayar:
According to the documentation (https://nodejs.org/api/esm.html#loaders), when calling `nextResolve`
and `nextLoad`, one should not pass the `next*` function again to itself.

This causes a problem when using multiple loaders, as evidenced by this issue: https://github.com/testdouble/testdouble.js/issues/496.

Two notes:

1. Not sure exactly what the scenario is, so I couldn't figure out what test to write. All I know is that
   in the reproduction, when I removed ts-node's recursive call, everything worked.
2. Now that the parameter `defaultResolve/load` signature is not the same as `resolve/load`-s,
   I made it `any`. Any other try created errors in other places. Sorry, I'm not a TS expert.

### DIFF
--- a/dist-raw/node-internal-modules-esm-resolve.js
+++ b/dist-raw/node-internal-modules-esm-resolve.js
@@ -846,7 +846,7 @@ function resolveAsCommonJS(specifier, parentURL) {
   }
 }
 
-function defaultResolve(specifier, context = {}, defaultResolveUnused) {
+function defaultResolve(specifier, context = {}) {
   let { parentURL, conditions } = context;
   if (parentURL && policy != null && policy.manifest) {
     const redirects = policy.manifest.getDependencyMapper(parentURL);
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,6660 @@
 {
   "name": "ts-node",
   "version": "10.9.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "ts-node",
+      "version": "10.9.1",
+      "license": "MIT",
+      "dependencies": {
+        "@cspotcode/source-map-support": "^0.8.0",
+        "@tsconfig/node14": "*",
+        "@tsconfig/node16": "*",
+        "@tsconfig/node18": "*",
+        "acorn": "^8.4.1",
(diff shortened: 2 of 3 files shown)


giltayar:
@cspotcode: from what I've seen, this changed in 18.6.0, when they added multiple loader support. Before you could call with the third argument and it worked.

It's not fishy: it's not _really_ calling the next resolution in the chain, but rather going through some code of their own. But you're right in that if the function should be called with two arguments, then they should be totally ignoring the third argument!

But what can I say 🤷‍♂️? In the end, the documentation says that you should not pass the third argument, so the fix is still a _correct_ one, and it fixes problems when this loader is in a chain.

cspotcode:
I've raised this with the loaders team on Slack.  Proper bugfix should happen in node, so we'll pursue that route.  In the meantime, downgrading node is always an option.

It's an unwritten rule[^1] of JS, that functions implicitly agree not to change behavior when passed additional arguments, unless they're documented to take rest args.  The node code in question was actually flagged in code review, so seems likely that my report will cause them to fix this.  The code is also checking `arguments.length` instead of `=== undefined`, which is another code smell.

[^1]: "Rule" meaning we should follow it because it makes everyone happier, not because of any tangible enforcement mechanism with the runtime or anything like that.

giltayar:
@cspotcode your call, of course! Feel free to close this PR without merging it, because in the end you are right and this is a Node.js bug/smell. But from what I've seen, removing that third argument does not harm in _any_ version of Node, so it's a benign change.

cspotcode:
Got confirmation on Slack that this will be fixed in node.  I've added a note to fix the types here: #1505 (that issue is titled breaking changes, but even if this types change isn't breaking, our next version is going to be a major, so that was an easy place to write it down)

I meant to dig up another example of why functions shouldn't rely on `arguments.length` nor change behavior when passed undocumented args: `Array.prototype.map`

E.g. it's common to see `function mapStuff(input) {return input.foo ? true : false}` and then do `items.map(mapStuff)`.
Additional args are being passed, but that's ok, because `mapStuff`'s documentation says that it accepts a single arg, so the unwritten expectation is that it'll accept a second and third arg without breaking.

cspotcode:
Fixed by https://github.com/nodejs/node/pull/44109

PR #1852, opened 2022-07-28
Pass jsxImportSource to swc

alecmev:
I'm using it for [`hastscript`](https://github.com/syntax-tree/hastscript/blob/2f8c133920fa57405815de3f3cdc21d1bc892444/readme.md#jsx). Tested locally.


cspotcode:
As you can probably guess, like so many open-source projects we do not rely on local testing, so that is insufficient.  Will need proper automated tests if it is to be merged.

cspotcode:
@jraoult thanks for taking a look.  I think you're correct, that is the file to put tests.  It looks like we have two kinds of tests:

a) tests that `createSwcConfig()` is generating the correct SWC configuration based on a TypeScript tsconfig
b) tests that `compile()` is transforming a snippet of input TS into the correct output JS

Adding one of each should be good.

RomainLanz:
Hey there! 👋🏻 

I'd like to take ownership of this PR/issue.

Currently, `ts-node` doesn't play nicely with `swc` when altering the `jsxImportSource` setting in a `tsconfig.json` file. I've managed to get it working by monkey-patching the module with the provided changes from @alecmev.

Regarding tests, adding them seems straightforward. However, on two of my machines, the test suites fail even without any changes. This makes it difficult to guarantee the code's functionality cleanly.

Note that I don't believe it's necessary to include a test for this PR merge. Many existing features, such as the current `jsxFactory` and `jsxFragmentFactory` options, are also untested.

PR #1854, opened 2022-07-31
Dev experience

cspotcode:
- switch to yarn
- [x] test that node and TS versions are correct, in case CI scripts have a flaw
  - e.g would be possible for typescript install to go wrong, and then we would not be testing against old version of TS


cspotcode:
Removed the following from OP; will track with a follow-up PR

- [ ] testability improvements?
  - allow configuration tests to skip executing code
  - `service.getInternalConfiguration()`

PR #1901, opened 2022-10-19
Update options.md

sosoba:
See https://nodejs.org/en/blog/announcements/v19-release-announce/


cspotcode:
This option probably still works in ts-node even when running on node 19.  You can test it to be sure.

PR #1946, opened 2023-01-29
Update GitHub workflow badge route

dscalzi:
see https://github.com/badges/shields/issues/8671


cspotcode:
Good catch, thank you.

PR #1958, opened 2023-02-13
TS 5.0 Supporting Multiple Configuration Files in extends

### LINKED ISSUE #1954: Add support for TS 5.0 "extends" array
TS 5.0 adds support for "extends" to be an array of one or more extended configs, not just a single config.

https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#supporting-multiple-configuration-files-in-extends

We can do the same, also merging `ts-node` options from multiple tsconfigs.

cspotcode:
Fixes #1954


cspotcode:
No set time.  I recommend installing from git in the meantime.

https://github.com/TypeStrong/ts-node/discussions/1977

cspotcode:
To be sure we're on the same page, you're in an environment where the `prepare` script will not run for a git dep?  But yeah, you can go with any workflow where you `npm pack` and then put that tarball somewhere and use it.

WoodyWoodsta:
In my environment, `ts-node` is a transitive dependency and we use `yarn` as a package manager. I believe `yarn` > 1 doesn't support `prepare`: https://github.com/yarnpkg/berry/issues/2305

cspotcode:
Thanks for bringing that to my attention.  I've switched us over to using `prepack` instead #1979

WoodyWoodsta:
No problem - thank you for making that change!

DreamOfIce:
Is there anything preventing us from releasing a patch version? 
Installing from git is slow and often reports errors

slavafomin:
Hello! Thank you for your hard work, however, what's blocking this patch from being released? It's been half a year already and a lot of people are still having trouble with this issue.

roydukkey:
If your still waiting for this release, you might look into [tsx](https://github.com/esbuild-kit/tsx).

slavafomin:
> If your still waiting for this release, you might look into [tsx](https://github.com/esbuild-kit/tsx).

Wow, it looks pretty cool. Thank you for mentioning it. I will definitely give it a try.

Elyahou:
Any news about the release ?

FreePhoenix888:
I am going to use typescript and node instead of using ts-node because this takes so long

mrkjdy:
@blakeembrey Can a patch version be released to fix this problem as well? I was hoping that this was available in `10.9.2`, but it's still not.

cAttte:
will this very simple feature ever be released?

PR #1968, opened 2023-03-13
fix: update swc.ts to disable useDefineForClassFields option for swc

vagusX:
@swc/core@1.3.40 enable useDefineForClassFields

https://github.com/swc-project/swc/pull/7065

### DIFF
--- a/src/transpilers/swc.ts
+++ b/src/transpilers/swc.ts
@@ -240,6 +240,7 @@ export function createSwcOptions(
         transform: {
           decoratorMetadata: emitDecoratorMetadata,
           legacyDecorator: true,
+          useDefineForClassFields: false,
           react: {
             throwIfNamespace: false,
             development: jsxDevelopment,


Graphmaxer:
Any update on this ? I'm stuck on SWC 1.3.39 without this.

ldiqual:
Same here, we're stuck on 1.3.39. I think it would be good to either read `.swcrc`, allow customizing the swc config, or set this default so that users can upgrade.

CitadelCore:
Would really appreciate this merged in

grundb:
In my opinion, it seems like a better idea to respect the TS config and get the value from there. Many options are already extracted here: https://github.com/TypeStrong/ts-node/blob/47d4f45f35e824a2515e17383a563be7dba7d8ff/src/transpilers/swc.ts#L131-L145. One could add `useDefineForClassFields` to this destructuring, then just pass it along to the line which this PR intends to change. I experimented a bit locally and it seemed to work.

courage173:
Why can't this be merged in been stuck on 1.3.26 a long time

cspotcode:
Fixed via #2062

PR #1991, opened 2023-04-13
Add syntax highlight for the input

bimusiek:
Added syntax highlight for the input.
CleanShot 2023-04-14 at 01 01 40@2x


bimusiek:
I have also included `--preprocessor` option as this helps us a lot with interactive CLI where we can have some initial setup done.
CleanShot 2023-04-14 at 02 53 25@2x

package.json
```diff
@@ -161,7 +161,6 @@
     "acorn-walk": "^8.1.1",
     "arg": "^4.1.0",
     "diff": "^4.0.1",
-    "glob": "^10.0.0",
```
Hi-Angel: The changes of this commit are actually fixing the previous commit, so it shouldn't be a separate commit.

Issue #1997, opened 2023-04-20
ERR_UNKNOWN_FILE_EXTENSION on Node v20.0.0


septs:
### Search Terms

Node, ERR_UNKNOWN_FILE_EXTENSION

### Expected Behavior

Fix it

### Actual Behavior

see Minimal reproduction

### Steps to reproduce the problem

see Minimal reproduction

### Minimal reproduction

```console
$ cat example.ts
console.log('example')
$ cat tsconfig.json
{ "ts-node": { "esm": true } }
$ npx ts-node example.ts
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /Users/septs/Projects/example/example.ts
    at new NodeError (node:internal/errors:399:5)
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:99:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:139:38)
    at defaultLoad (node:internal/modules/esm/load:83:20)
    at nextLoad (node:internal/modules/esm/hooks:781:28)
    at load (/Users/septs/.npm/_npx/1bf7c3c15bf47d04/node_modules/ts-node/dist/child/child-loader.js:19:122)
    at nextLoad (node:internal/modules/esm/hooks:781:28)
    at Hooks.load (node:internal/modules/esm/hooks:381:26)
    at handleMessage (node:internal/modules/esm/worker:153:24)
    at checkForMessages (node:internal/modules/esm/worker:102:28) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

### Specifications

* ts-node version: 10.9.1
* node version: 20.0.0
* TypeScript version: 5.0
* tsconfig.json:

  ```json
  { "ts-node": { "esm": true } }
  ```

* package.json:

  ```json
  {}
  ```

* Operating system and version: macOS 13.3.1

septs:
https://github.com/microsoft/TypeScript/issues/53922

csvn:
I have the same issue. I tried using `--esm`, `{ "esm": true }` and `--loader="ts-node/register"`, but got the `ERR_UNKNOWN_FILE_EXTENSION` no matter what I tried. Could not get `ts-node` to run on Node v20 with native ESM modules.

Jack-Works:
FYI Node 20 has made a breaking change:

> Custom ESM loader hooks run on dedicated thread
[...]

PR #2003, opened 2023-04-28
Add @tsconfig/node20, add node 20 to test matrix

chentsulin:
Add @tsconfig/node20, add node 20 to test matrix.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -95,8 +95,8 @@ jobs:
             typescriptFlag: latest
           # Node nightly
           - flavor: 6
-            node: nightly
-            nodeFlag: nightly
+            node: 21-nightly
+            nodeFlag: 21_nightly
             typescript: latest
             typescriptFlag: latest
     steps:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -63,7 +63,7 @@ jobs:
       matrix:
         os: [ubuntu, windows]
         # Don't forget to add all new flavors to this list!
-        flavor: [1, 2, 3, 4, 5]
+        flavor: [1, 2, 3, 4, 5, 6]
         include:
           # Node 16
           - flavor: 1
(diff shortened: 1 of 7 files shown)


RobinTail:
Does this fix https://github.com/TypeStrong/ts-node/issues/1997 (the ESM support) , @chentsulin ?

R-Rajaneesh:
+1 Bump

.github/workflows/continuous-integration.yml
```diff
@@ -87,10 +87,21 @@ jobs:
             nodeFlag: 18
             typescript: next
             typescriptFlag: next
-          # Node nightly
+          # Node 20
```
cspotcode, line 90: also need to add to `flavor` array above

.github/workflows/continuous-integration.yml
```diff
@@ -87,10 +87,21 @@ jobs:
...
-            nodeFlag: 20_nightly
+            node: 20
+            nodeFlag: 20
+            typescript: latest
+            typescriptFlag: latest
+          - flavor: 6
+            node: 20
+            nodeFlag: 20
+            typescript: next
+            typescriptFlag: next
```
cspotcode, line 100: We can remove this one; I'm trying to keep only a single entry for typescript@next.  I suspect that github actions throttle if you use it too much, so I try to keep the test matrix small where possible.

cspotcode:
I renamed this PR to clarify that it does not add support for node 20, it only adds the @tsconfig/bases for node 20.

R-Rajaneesh:
@cspotcode So when will ts-node support node v20, my projects are currently in a standstill due to this

.github/workflows/continuous-integration.yml
```diff
@@ -87,10 +87,16 @@ jobs:
...
-            node: 20-nightly
-            nodeFlag: 20_nightly
+            node: 20
+            nodeFlag: 20
+            typescript: latest
+            typescriptFlag: latest
+          # Node nightly
+          - flavor: 6
+            node: nightly
+            nodeFlag: nightly
```
cspotcode, line 99:
```suggestion
            node: 21-nightly
            nodeFlag: 21_nightly
```

OneCricketeer:
Should there be a release newer than a year ago that makes mention on Node20? 

https://github.com/TypeStrong/ts-node/releases

PR #2049, opened 2023-09-01
Pass useDefineForClassFields tsconfig compiler option to SWC

valentin-nemcev:
Defaults for this option may be different between tsc and SWC, and explicit override in ts-node is ignored.

This prevents drop-in replacement of `ts-node --swc` in place of `ts-node` for cases when the value of `useDefineForClassFields` is important

Here is an example of the problem this PR is attempting to fix:

tsconfig.json:
```json
{
  "compilerOptions": {
    "target": "es2021",
    "useDefineForClassFields": false
  }
}
```

```bash
$ ts-node -pe 'new class { f?:any }'
{}

$ ts-node --swc -pe 'new class { f?:any }'
{ f: undefined } # but should be {}
```

### DIFF
--- a/src/transpilers/swc.ts
+++ b/src/transpilers/swc.ts
@@ -134,6 +134,7 @@ export function createSwcOptions(
     importHelpers,
     experimentalDecorators,
     emitDecoratorMetadata,
+    useDefineForClassFields,
     target,
     module,
     jsx,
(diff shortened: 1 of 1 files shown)


cspotcode:
Fixed via #2062

Issue #2056, opened 2023-09-07
SWC: Reflect renamed keepImportAssertions configuration option


felixh10r:
### Desired Behavior

With version 1.3.83, swc renames the experimental configuration option `jsc.experimental.keepImportAssertions` to `jsc.experimental.keepImportAttributes` as per https://github.com/swc-project/swc/pull/7914 . 

### Is this request related to a problem?

All setups with swc > 1.3.82 will fail. This creates a dependency version limitation which should probably be reflected in `peerDependencies` or simply updated in a new `ts-node` release.

### Additional context

If the solution is to _simply_ change the name of the property, I'm happy to open a PR.

felixh10r:
since the swc config option is experimental, it was changed in a _patch_ update to `@swc/core`, which may be updated by developers without knowing about the breaking behavior.

meabed:
I tried commenting in SWC :) but no luck 😢

isaacs:
Actually, we should just remove that option from ts-node, I think.  swc only emits `import .. from .. with {}`, not `import .. from .. assert {}`, and no version of node supports that syntax.

1.3.83 adds `emitAssertForImportAttributes`, but [it doesn't do that](https://github.com/swc-project/swc/issues/7928), so typescript currently actually relying on `import/assert` is (a) currently broken when using swc with ts-node, and (b) not likely to work any time soon.

Better to just have it strip the attributes, imo.

isaacs:
`emitAssertForImportAttributes` would be nice, but I'd still recommend waiting until it's promoted out of experimental status.

RobinTail:
Perhaps It would be convenient to have an ability to configure swc options in tsconfig besides just having
```
"swc": true
```

meyfa:
I think a way of configuring SWC would definitely be nice long-term. For now, I would like to get SWC `>=1.3.83` to not fail our CI builds with ts-node. It seems like PR #2057 would achieve this. Is there any way to help moving this forward? @cspotcode

MuTsunTsai:
Experiencing the same issue as well. Have to stay at SWC 1.3.82 for now.

PR #2057, opened 2023-09-07
remove swc experimental options

isaacs:
This removes support for keeping import assertions, which were broken in swc at some point, and unconditionally transpiled into import attributes. (Ie, `import/with` instead of `import/assert`.)

No version of node supports import attributes with this syntax yet, so anyone using swc to import json in ESM is out of luck no matter what.

And swc 1.3.83 broke the option that ts-node was using. The position of the swc project is that experimental features are not supported, and may change in patch versions without warning, making them unsafe to rely on (as evidenced here, and the reason why this behavior changed unexpectedly in the first place).

Better to just not use experimental swc features, and let it remove import assertions rather than transpile them into something that node can't run.

Fix: https://github.com/TypeStrong/ts-node/issues/2056

### DIFF
--- a/src/test/transpilers.spec.ts
+++ b/src/test/transpilers.spec.ts
@@ -120,23 +120,4 @@ test.suite('swc', (test) => {
       );
     });
   });
-
-  test.suite('preserves import assertions for json imports', (test) => {
-    test.if(tsSupportsImportAssertions);
-    test(
-      'basic json import',
-      compileMacro,
-      { module: 'esnext' },
-      outdent`
-        import document from './document.json' assert {type: 'json'};
-        document;
-      `,
-      outdent`
-        import document from './document.json' assert {
-            type: 'json'
-        };
-        document;
-      `
-    );
-  });
 });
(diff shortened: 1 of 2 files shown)


isaacs:
#2009 is rebased onto this, so landing that would close this as well. Happy to separate them if that's easier.

meabed:
@isaacs would it be possible to move this forward soon?

wololodev (approved):

isaacs:
@meabed ready to land, as far as I'm concerned. I'm not the one who needs to say yes, though.

meabed:
@wololodev - hope you well, could this get merged soon? Thank you 😄

cspotcode:
This is resolved via #2062 

Node mandates import assertions for json imports.

jraoult:
@cspotcode The least official release of `ts-node` is more than a year old so I can imagine the main branch to be quite packed in changes and might take time to be ready for release. Are you planing to cherry pick and release this particular fix soon? Im trying to see if I should just patch `ts-node` in my repo or just wait.

meabed:
@cspotcode Thank you fixing and merging this fix - would it be released soon?

PR #2062, opened 2023-09-14
Fix swc issues

### LINKED ISSUE #1996: @swc/core@1.3.51 does not return sourcemap for type-only files, causes error in `updateSourceMap`
### Search Terms
SWC

### Expected Behavior

I am using ts-node with the `--swc` flag. I expect it to work. It works fine for me on SWC < 1.3.50

### Actual Behavior
I tried upgrading to SWC >= 1.3.50 today and it caused ts-node to not work. I unfortunately did not see much useful output. If there is a way to get more useful output to fix this, I am happy to try it.

Upgrade to SWC >= 1.3.50

### Minimal reproduction
Be on ts-node version `10.9.1` and SWC version `1.3.50` and try to use ts-node with the `--swc` flag.

### Specifications

* ts-node version: 10.9.1
* node version: 18.x
* TypeScript version: 4.8.4
```
"ts-node": {
    "swc": true
}
```
* package.json:
The relevant part would be:
```
`... node_modules/ts-node/dist/bin.js --swc --transpile-only index.ts`
```
* Operating system and version: Mac

### LINKED ISSUE #2051: Tracking: swc erroneously converts import assertions `assert` keyword to `with`
Tracking: https://github.com/swc-project/swc/issues/7908

Once swc fixes this, we may need to require the newer version or pass a flag.

Converting to `with` might be correct in some situations, I'm not an expert on the spec changes.  But I know that for our purposes on node, we need to keep the `assert`.

### LINKED ISSUE #2056: SWC: Reflect renamed keepImportAssertions configuration option

### LINKED ISSUE #2059: The latest version of @swc/core has made a break change, resulting in ts-node reporting an error.
"ts-node": "^10.9.1"

"@swc/core": "^1.3.78" is ok,but subsequent versions will report errors.

The peerDependencies declaration in the ts node/package.json file will cause the installation of the latest dependent package instead of the specified one.

  "peerDependencies": {
    "@swc/core": ">=1.2.50",
    "@swc/wasm": ">=1.2.50",
    "@types/node": "*",
    "typescript": ">=4.2"
  }

cspotcode:
### Issues fixed

The latest version of swc renamed some options.
Also, a new option is necessary to tell swc to emit import assertions with the `assert` keyword instead of `with`.
This effectively requires a newer version of SWC, so I bumped the minimum required version

ts-node was not passing the correct `useDefineForClassFields` value to swc to match the user's tsconfig / tsc's behavior.

ts-node was erroring when custom transpilers did not return a sourcemap.

ts-node did not pass jsxImportSource to swc

swc 1.3.85 now forbids irrelevant `module.` options.  For example, the `noInterop` option has no effect if `module.type == 'es6'`, and 1.3.85 now raises an error if you try to pass `module: {type: 'es6', noInterop: true}`.  ts-node now conditionally passes those options to avoid this error

### Tickets resolved

Fixes #2059
Fixes #2056
Fixes #2051
Replaces #2057
Replaces #2049
Replaces #1968
Fixes #1996
Replaces, fixes #1852


RobinTail:
@cspotcode , please release

Issue #2072, opened 2023-10-04
Export `ts-node/esm-register` to use with `--import`


jlenon7:
### Desired Behavior

Register the `ts-node/esm` loader using the `register` function. Example of use:

`node --import=ts-node/esm-register`

### Is this request related to a problem?

No.

### Alternatives you've considered

.

### Additional context

.

AaronNGray:
For further context and alternative example see :- https://github.com/TypeStrong/ts-node/issues/2100

PR #2073, opened 2023-10-04
Export `ts-node/esm-register`

jlenon7:
Register `ts-node/esm` loader programmatically using `register` fn.

See issue #2072


klippx:
This is great, we will get rid of 
```
(node:2649) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("ts-node/esm", pathToFileURL("./")); register("extensionless", pathToFileURL("./"));'
```

if we use the `--import ts-node/esm` instead of `--loader ts-node/esm`

TheUnlocked:
Make sure to add esm-register.mjs to the files list in package.json.

kkolinko:
Regarding the proposed patch:

1. The first argument to `register()`: What is the point of calling `pathToFileURL()` on it? Why one would want to convert a relative path to a `file://` URL? (I would understand if it were an absolute path, resulting in an absolute URL. And in that case the second argument to register() could be omitted. But why for a relative one? `register()` is able to resolve relative paths relative to its second argument - that is the purpose of the second argument of `register()`.)

2. The second argument to `register()`: What is the meaning of "./" here? It usually means the current working directory of a process. Why not use `import.meta.url` - the URL of this very module?

An example can be found here:
https://nodejs.org/dist/latest-v20.x/docs/api/module.html#enabling
Click "CJS <-> ESM" switch in the top-right corner to show it in ESM syntax.

esm-register.mjs
```diff
@@ -0,0 +1,4 @@
+import { register } from 'node:module'
+import { pathToFileURL } from 'node:url'
+
+register(pathToFileURL('./esm.mjs'), pathToFileURL('./'))
```
thetutlage: This could be simplified to 

```ts
import { register } from 'node:module'
register('./esm.mjs', import.meta.url)
```

ihmpavel:
Hi @jlenon7 did you have time to look at the proposed changes? Looking forward to this PR to be merged. If I can help, i would love to

nickserv:
This looks good and it's nearly identical to the code Node recommends when using legacy `--loader`, though I'd recommend @thetutlage's suggestion. Is this enough to support `--import ts-node/esm` or would it have to be exposed?

jlenon7:
Sorry the delay, I've uploaded the code with the suggestions.

AaronNGray:
For further context and alternative example see :- https://github.com/TypeStrong/ts-node/issues/2100

alexsch01:
Any update on this?

alexsch01:
With this PR I get the following
```
C:\Users\WDAGUtilityAccount\Downloads\proj>node --import ts-node/esm-register script.ts
C:\Users\WDAGUtilityAccount\Downloads\proj\script.ts:1
function add(x: number, y: number) {
              ^

SyntaxError: Unexpected token ':'
    at internalCompileFunction (node:internal/vm:77:18)
    at wrapSafe (node:internal/modules/cjs/loader:1288:20)
    at Module._compile (node:internal/modules/cjs/loader:1340:27)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at cjsLoader (node:internal/modules/esm/translators:356:17)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:305:7)
    at ModuleJob.run (node:internal/modules/esm/module_job:218:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:329:24)

Node.js v20.11.1
```

ElvenMonky:
So what's the resolution on this? Is it coming in?

jlenon7:
> So what's the resolution on this? Is it coming in?

Well, the implementation is done. We just need to wait to the core contributors to merge it or some feedback if they need more changes.

parimalyeole1:
@jlenon7 Thanks for this PR.
@thetutlage, Is it possible to move forward with this PR? This will help us a lot.

Thank You

parimalyeole1:
@jlenon7 Thanks for this PR.
@blakeembrey @thetutlage, Is it possible to move forward with this PR? This will help us a lot.

Thank You

eduwardo:
Any update on merging this pr?

albertodiazdorado:
It would be lovely if we could get this merged

jlenon7:
Hey @cspotcode could you take a look on this please?

programmingwithsalim:
npm i -D tsx
npx tsx src/index.ts

This worked for me

albertodiazdorado:
I agree with @programmingwithsalim . Either `tsx` (based on `esbuild`, written in `go`) or `swc-node` (based on `swc`, written in `rust`) are the standard way to directly execute TypeScript files in 2025.

Sadly, `ts-node` is a relic of the past.

(Additionally: we shall keep an eye on the built-in feature of Node.js to execute TypeScript files, which is also based on `swc` if I am not mistaken)

davidthornton:
@albertodiazdorado, not to mention the speed benefits - which we've all heard about but I can anecdotally confirm have indeed been picked up for us too when running `node --watch`, `node --test` and when starting up in production! Very sad for `ts-node`, but very exciting for `swc-node` and TypeScript and the world! Do you have a reference to the built-in TypeScript running feature planning from Node.js? I would like to take a look. Thanks!

inoyakaigor:
@davidthornton I think this link could help you to start https://nodejs.org/docs/latest-v22.x/api/typescript.html#type-stripping

PR #2136, opened 2024-09-04
Fix SWC support with stage 3 decorators

MuTsunTsai:
Since TypeScript 5.0, stage 3 decorators had become the default while `experimentalDecorators` had become the legacy standard. However, with SWC enabled, ts-node doesn't work correctly with stage 3 decorators. This PR fixes the config pass to SWC to make it things work as it should.

Note: This part of the SWC config has been there since v1.3.47, so there's no need the change the dependency specifications.

### DIFF
--- a/src/transpilers/swc.ts
+++ b/src/transpilers/swc.ts
@@ -225,14 +225,15 @@ export function createSwcOptions(
         parser: {
           syntax: 'typescript',
           tsx: isTsx,
-          decorators: experimentalDecorators,
+          decorators: true,
           dynamicImport: true,
           importAssertions: true,
         } as swcWasm.TsParserConfig,
         target: swcTarget as swcWasm.JscTarget,
         transform: {
           decoratorMetadata: emitDecoratorMetadata,
-          legacyDecorator: true,
+          legacyDecorator: experimentalDecorators,
+          decoratorVersion: experimentalDecorators ? '2021-12' : '2022-03',
           react: {
             throwIfNamespace: false,
             development: jsxDevelopment,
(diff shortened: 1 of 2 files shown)


src/transpilers/swc.ts
```diff
@@ -225,14 +225,15 @@ export function createSwcOptions(
...
+          decorators: true,
           dynamicImport: true,
           importAssertions: true,
         } as swcWasm.TsParserConfig,
         target: swcTarget as swcWasm.JscTarget,
         transform: {
           decoratorMetadata: emitDecoratorMetadata,
-          legacyDecorator: true,
+          legacyDecorator: experimentalDecorators,
+          decoratorVersion: experimentalDecorators ? '2021-12' : '2022-03',
```
TechQuery, line 236:
```suggestion
          decoratorVersion: experimentalDecorators ? '2021-12' : '2023-11',
```

TechQuery (changes requested):

src/transpilers/swc.ts
MuTsunTsai: Where did you get that 2023-11 from? SWC doesn't have that option: https://swc.rs/docs/configuration/compilation#jsctransformdecoratorversion

TechQuery: > Where did you get that 2023-11 from? SWC doesn't have that option: https://swc.rs/docs/configuration/compilation#jsctransformdecoratorversion

I saw this value in SWC error message & Babel document, maybe it hasn't been implemented.

jzhan-canva: no it hasn't

Issue #1856, opened 2022-08-05
SWC ignore .swcrc config


Kaizer69:
### Desired Behavior
When use ts-node, you should use default swcrc that ts-node create, but if is present a .swcrc file, you should use that file.

### Is this request related to a problem?
Yes, because if I have a JS file with private members (hash # char), SWC wont compile

cspotcode:
What's your `.swcrc` look like?

Kaizer69:
```
{
  "$schema": "https://json.schemastore.org/swcrc",
  **"exclude": [".*.js$", ".*.map$"],**
  "jsc": {
    "parser": {
      "syntax": "ecmascript",
      "jsx": false,
      "dynamicImport": false,
      **"privateMethod": true,**
      "functionBind": false,
      "exportDefaultFrom": false,
      "exportNamespaceFrom": false,
      "decorators": true,
      "decoratorsBeforeExport": false,
      "topLevelAwait": false,
      "importMeta": false
    },
  },
}
```

I try to pure an invalid .swcrc and ts-node start without problems. The configuration is skipped.
I want to exclude JS (that SWC can parse for private methods) or use privateMethod flag of SWC.

cspotcode:
Is there a minimal example of the .js file raising this error?

Kaizer69:
I cannot provide the example, because the JS lib that I've use is highly obfuscated, so it's difficult extract only the private part.

cspotcode:
It can be a new one, just needs to be an example that can be used to test with.  Requesting a minimal reproducible example is common practice in open-source: https://en.wikipedia.org/wiki/Minimal_reproducible_example

vinczebalazs:
I have the same problem also, any .swcrc config file placed in my root directory is ignored.

cajoy:
The same here

falkenhawk:
Is it really necessary for ts-node to tell swc to ignore `.swcrc` altogether by this setting?
https://github.com/TypeStrong/ts-node/blob/main/src/transpilers/swc.ts#L228

radriaanse:
Slightly different use case here, using the `.swcrc` file to specify path mappings like;

```json
  "jsc": {
    "paths": {
      "@/some/*": ["thing/*"],
    },
```

ts-node could optionally allow specifying these (or fetching them from the `tsconfig.json`) but if we could enable loading the `.swcrc` ts-node doesn't have to be updated every time new options become available in swc.

[...]

PR #2143, opened 2024-12-03
fix adding swc configuration file support as an option

hanseltime:
# Summary

This PR implements what looks like the correct changes for ts-node to support specifying an swcConfig.

Unfortunately, the version of yarn has a bug that won't download the github based packages due to checksum mismatch and the tests seem very reliant on that particular yarn.  If someone can point me to documentation for how to run the tests with out debugging tons of failures, I would be more than happy to implement additional tests.

This would unlock a lot of swc based flows as indicated in:  https://github.com/TypeStrong/ts-node/issues/1856


sdotson:
hi @hanseltime . excited to see you try to get this into `ts-node`. I'm not a maintainer or have any real power here, but happy to try to help. What errors are you seeing? I pulled down your pr and I only see one failing test `transpilers › swc › verify that TS->SWC target mappings suppport all possible values from both TS and SWC`. 

To be fair, I see the same failure on main branch. Is this what you are seeing as well or are you referring to a different sort of error?

hanseltime:
Hey @sdotson,

If you are capable of running the various tests, etc.  Then that might help.  The contributing.md doc seems to be out of date (or just doesn't handle me running this on a linux machine pop-os).

A list of the problems I have when trying to use the prescribed calls:

1. `yarn install` fails because all of the "github" sourced packages fail checksums
3. When I tell yarn to ignore checksums, then `yarn build` fails due to lots of strict failures: 
```
src/test/repl/repl.spec.ts:574:23 - error TS7006: Parameter 't' implicitly has an 'any' type.

574   test('test', async (t) => {
```

As an example^

Running tests fails due to the same typescript build failures and it seems testing types aren't recognized.  I could try getting something to work, but since I'm not sure if that would affect tests, I'm personally hamstrung to verify it.

For anyone who may be able to verify this, please feel free to do so!

sdotson:
It sounds like something is going on with your local `node` or `yarn` installation. If you post whatever is logged to the conosle, I might be able to help. Google is telling me it might be worth checking your `yarn`/`node` version and perhaps running `yarn cache clean`.

As I noted in my previous message, it looks like tests still pass on my local machine! If you'd like to add an additional test, I can run and verify for you if needed.

hanseltime:
@sdotson - Something is definitely going on.  

The yarn version is dictated by corepack from the project and I have verified that they match.  Likewise with node.  I have tried both 18 and 20 (18 is specified at places in the package.json).  I may try getting to this to run on a different machine. 

Since I am not finding any additional direction in the Contributing.md though, I'm not very willing to spend a lot of time debugging the build/test system without someone who understands the system providing any gotchas (i.e. maybe they've always used X setting to bypass this problem).

PR #2163, opened 2025-08-17
fix https://github.com/nodejs/node/issues/59364

alexsch01:
Let me know if this is the right approach

No longer repros with https://github.com/nodejs/node/issues/59364#issuecomment-3157690609


alexsch01:
For anyone wondering, I didn't touch `module-typescript` because ESM with ts-node is very broken from my experience
- pre `--experimental-strip-types` flag

Issue #2140, opened 2024-10-23
Seeking sponsorship to restart development


blakeembrey:
I'm looking for companies or individuals who benefit from `ts-node` and would like to see continued updates to sponsor work on this package. I'm more than happy to coordinate with featuring company logos on the README/website, or discuss other ways to show appreciation for sponsorship of work on this package.

It's been a while since this package has had a regular maintainer, and more than a few years since I was involved, but I know many people use this package and I would like to keep it maintained and up-to-date. However, the surface area has grown over the years and it's certainly not going to be a small effort to update it, keep it updated, and respond to all incoming issues and PRs. I would use the sponsorship funds to go full time on updating the package.

Some of the things I'd like to see done:

* Rewrite to support the latest ESM loading APIs, remove all the hacks related to older node version support
* Bump the minimum node.js supported version
* Improve overall runtime performance and memory usage
* Fix and update existing test suites
* Simplify the supported API space so new maintainers can quickly onboard
* Review and close the 200 open issues and PRs

If you are seeking an alternative to `ts-node`, I would recommend `tsx`.

Please keep comments on topic. Feel free to email me if you prefer.

Hi-Angel:
@blakeembrey hi, I'm a bit confused. This project resides under `TypeStrong` namespace, as opposed to being under an individual's. This means it currently has *at least* 15 maintainers:

![Image](https://github.com/user-attachments/assets/df747d80-6798-4370-ac07-b63d565aa27c)

These are only those who opted in to being "public members", the actual number I presume is much higher *(on Github for some reason group membership is hidden by default unless you turn on your visibility explicitly)*.

So why are you saying the project has no maintainers?

beeman:
> [@blakeembrey](https://github.com/blakeembrey) hi, I'm a bit confused. This project resides under `TypeStrong` namespace, as opposed to being under an individual's. This means it currently has _at least_ 15 maintainers:
> 
> ![Image](https://github.com/user-attachments/assets/df747d80-6798-4370-ac07-b63d565aa27c)
> 
> These are only those who opted in to being "public members", the actual number I presume is much higher _(on Github for some reason group membership is hidden by default unless you turn on your visibility explicitly)_.
> 
> So why are you saying the project has no maintainers?

The number of 'public members' of a github org doesn't say anything. These are people that have been added to the github org at some point in time.

Check the contributors of this repo to see who actually did the work here (to no surprise, it's mostly @blakeembrey and @cspotcode (edit)).

[...]

Issue #2119, opened 2024-05-25
Last release was 6 months ago... is the project dead?


devlato:
Last release was 6 months ago... is the project dead?

EvilCodeMorty:
I don't know. The previous problem can't be solved by switching versions. It's over

richardsimko:
We transitioned to [tsx](https://github.com/privatenumber/tsx) and haven't looked back

tkodev:
> We transitioned to [tsx](https://github.com/privatenumber/tsx) and haven't looked back

Jest still uses ts-node unfortunately, so if you're using jest you're out of luck.

richardsimko:
We use [@swc/jest](https://www.npmjs.com/package/@swc/jest) to do TS -> JS transformation, while Jest still installs ts-node it's not actually used for anything in this case (And as a bonus it's much faster)

devlato:
@koheiio @richardsimko thanks. Neither `tsx` nor `swc` do type-checking our of the box though. 
At the same time, `ts-node` does, which makes it a perfect choice for running stuff locally and for building CI automation and developer tools.

tkodev:
What I'm struggling with on my end is jest uses ts-node to read tsconfigs, which means latest typescript things like an array for the "extends" property isn't supported, and subpath imports/export resolution.

If you have come across this problem and solved it, it would greatly make my life less painless! I hope there news about ts-node, whether it be good or bad.

In the meantime, I'll check out swc/jest

vboginskey:
> If you have come across this problem and solved it, it would greatly make my life less painless!

Not a great option, but you can:

1. Check out the latest commit.
2. Package it yourself with `yarn pack`.
3. Put the resulting `package.tgz` into a package repo of your choice, or commit it to your project repo.
4. Use the `file:` protocol in your `package.json` to consume it (if committed to repo).

This fixes the bug that otherwise prevents you from using an array `extends` in `tsconfig.json`.

erayhanoglu:
> We transitioned to [tsx](https://github.com/privatenumber/tsx) and haven't looked back

If i am not wrong, tsx and swc do not support typescript decorators. We strongly uses decorators.

cjnoname:
> > We transitioned to [tsx](https://github.com/privatenumber/tsx) and haven't looked back
> 
> If i am not wrong, tsx and swc do not support typescript decorators. We strongly uses decorators.

swc supports decorators well, while tsx doesn't support

[...]

PR #2169, opened 2026-01-15
Upgrade diff dependency from v4 to v8.0.3 (GHSA-73rr-hh4g-fpgx)

drzippie:
## Summary
Upgrades the `diff` package from v4 to v8.0.3 to address security vulnerability [GHSA-73rr-hh4g-fpgx](https://github.com/advisories/GHSA-73rr-hh4g-fpgx).

## Security Advisory
- **ID**: GHSA-73rr-hh4g-fpgx
- **Severity**: Low
- **Issue**: DoS vulnerability in `parsePatch()` and `applyPatch()` - line break characters can cause infinite loops or ReDoS
- **Affected versions**: < 8.0.3
- **Patched version**: 8.0.3
- **Note**: ts-node only uses `diffLines()` which is NOT affected by this vulnerability

## Changes
- Updated `diff` from `^4.0.1` to `^8.0.3`
- Removed `@types/diff` (v8 includes built-in TypeScript types)

## Impact
- Only `diffLines()` is used in `src/repl.ts` for REPL code execution
- The API is backward compatible - no code changes needed
- All REPL tests pass locally


nsalto (approved):

scorgn:
Thank you for putting this together! Would be great to get this out.

I noticed the MR was pointing towards main, which lists is version as 11.0.0-beta.1. I see there is still a 10.x branch though which lists version 10.9.2.

Question for the maintainers - could this patch be backported to v10.x considering v11 never left beta?

khoserdene-mezorn:
This change is urgently needed. 2 low severity vulnerabilities exist in the current 10.9.2 version.

richardkazuomiller:
It seems that this project is no longer maintained so users may need to mitigate downstream. I don’t use ts-node directly but I am affected because it’s a transitive dependency of something else I use, so if there’s no activity here for a while I may request that the dependent of ts-node that I use make a change on their part.

<https://github.com/TypeStrong/ts-node/issues/2119>
<https://github.com/TypeStrong/ts-node/issues/2140>

50bbx (approved):

wilpat (approved):

alumni:
> This change is urgently needed.

`diff@4.0.4` was released and the vulnerabilities were patched there.

So there's a fix that doesn't need waiting for this PR to be merged.

richardkazuomiller:
~The fix in diff will not propagate downstream wothout this PR being merged because it is in diff version 8.x while the version ts-node includes is 4.x.~

richardkazuomiller:
Sorry! Apparently I cannot read! It looks like this is no longer necessary to remediate downstream 🤠

vahidshirvani:
@blakeembrey would you kindly merge this and publish a release so we can get rid of vulnerability?

jeffsays:
bump?

